PR #46, opened 2015-01-03
Fixes for POSIX time parsing.

mookerji:
0. Removes Scientific truncation, where 1.415335333E9 is truncated to
   1 when parsed as an integer.
1. Returned POSIX JSON values should be parsed as numbers and not
   strings.


brendanhay:
Looks good, thanks!

PR #47, opened 2015-01-04
gen/overrides: RootDeviceName is not required in Instance

proger:
Not sure if regenerated code belongs here though.


brendanhay:
Thanks! Only the override is needed, I commit the output definition periodically for audit purposes.

PR #48, opened 2015-01-04
core/gen: transform enums using toLower before pattern-matching them

proger:
this helps avoid errors like:

```haskell
*Main> getEnv Singapore Discover >>= flip runAWST (send describeInstances)
Left (SerializerError "EC2" "Failed reading: Failure parsing PlatformValues from \"windows\"")
```

and probably many more errors like that in the future.

### DIFF
--- a/core/src/Network/AWS/Data/Internal/Text.hs
+++ b/core/src/Network/AWS/Data/Internal/Text.hs
@@ -20,6 +20,8 @@ module Network.AWS.Data.Internal.Text
 
     , ToText   (..)
     , showText
+
+    , Text.toLower
     ) where
 
 import           Control.Applicative
@@ -35,7 +37,7 @@ import           Data.Scientific
 import           Data.Text                         (Text)
 import qualified Data.Text                         as Text
 import qualified Data.Text.Encoding                as Text
-import qualified Data.Text.Lazy                    as LText
+import qualified Data.Text.Lazy                    as LText hiding (toLower)
 import           Data.Text.Lazy.Builder            (Builder)
 import qualified Data.Text.Lazy.Builder            as Build
 import qualified Data.Text.Lazy.Builder.Int        as Build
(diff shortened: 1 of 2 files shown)


brendanhay:
The cased enum values are actually encoded in the service descriptions - if they're wrong, that's worrying :) and they should also be fixed upstream. I'll raise an issue for this.

I'd prefer the use of `Data.CaseInsensitive` when matching the constructor branches, rather than lowering the `Text` in the template. I might add an actual parser function that returns this instead of raw `Text`.

proger:
`Windows` was also capitalized in the docs — and I have no idea how botocore obtains their model specs (i've seen Amazon providing xml WSDLs though) — could you point out the place where you're going to raise this issue?

Did not know about `case-insensitive` — it sure makes the solution nicer, thank you!

brendanhay:
The service descriptions are generated internally at Amazon and then used by all their respective in-house SDKs - I suspect the code used to generate the documentation is related to the descriptions.

I'll either email a contact at Amazon directly when I have enough ammunition, or I'll just open an issue directly on [botocore](https://github.com/boto/botocore/issues).

brendanhay:
Would you like a Hackage release for this fix specifically?

proger:
Yes please! :)

brendanhay:
I prepared `release/0.1.4` containing these changes, but unfortunately I've encountered a fairly significant slowdown in the compilation speed of `amazonka-ec2`.

I've asked [haskell-cafe](https://groups.google.com/forum/#!topic/haskell-cafe/bunCLYXG2Zg) for some insight, but I may just switch back to using your original `Text.toLower` approach.

Apologies for the delay!

proger:
That's an interesting follow-up. Thanks for the heads-up.

brendanhay:
Fix contained in release [0.1.4](https://github.com/brendanhay/amazonka/releases/tag/0.1.4).

PR #50, opened 2015-01-06
makefile surgery

proger:
in a nutshell: simplifies bootstrap workflow into smth like
```console
% make -w deps install DEPS="core amazonka amazonka-elb amazonka-ec2" CABAL_INSTALL_DEFARGS="-j --disable-coverage"
```

makefiles: consistent deps; split library.mk install into {install,deps}
also:
- collapse subdir forwarding into a single macro
- introduce overridable CABAL_INSTALL_DEFARGS
(useful when using cabal-install 1.22 to use --disable-coverage)

### DIFF
--- a/share/library.mk
+++ b/share/library.mk
@@ -6,19 +6,21 @@ SHELL   := /usr/bin/env bash
 NAME    ?= $(notdir $(CURDIR:a/%=%))
 VERSION ?= $(shell sed -n 's/^version: *\(.*\)$$/\1/p' $(NAME).cabal)
 
-CABAL_SANDBOX_CONFIG := $(TOP)/cabal.sandbox.config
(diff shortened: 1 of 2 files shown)


share/library.mk
```diff
@@ -6,20 +6,22 @@ SHELL   := /usr/bin/env bash
 NAME    ?= $(notdir $(CURDIR:a/%=%))
 VERSION ?= $(shell sed -n 's/^version: *\(.*\)$$/\1/p' $(NAME).cabal)
 
-CABAL_SANDBOX_CONFIG := $(TOP)/cabal.sandbox.config
+CABAL_INSTALL_DEFARGS ?= -j --disable-documentation --disable-library-coverage
```
brendanhay: Since this is a development Makefile I think we can safely assume the latest `cabal-install` is used, and set `--disable-coverage` as the default to avoid the obsolete flag.

PR #52, opened 2015-01-09
Don't derive the Whole typeclass

shlevy:
It was not preserved when Numeric.Natural was brought into base, which
practically means it is difficult if not impossible to build this
against newer GHCs, and also suggests that the class itself is not all
that useful.


shlevy:
Alternative solutions:

* Make a compat package that defines the `Whole` class and import it from there instead of directly from `Numeric.Natural`
* Wrap the derivation in `#ifdef`ery

Let me know if one of those would be preferable.

brendanhay:
Thanks, I believe completely removing it should be fine. In future the `gen/output/` and modified `Types.hs` etc are not required in the PR as these get regenerated periodically as part of a larger continuous integration process.

Issue #40, opened 2014-12-18
Upgrade monad-control dependency


brendanhay:
With the recent release of `monad-control >= 1` the `MonadBaseControl` instance in `amazonka` for `AWST` breaks due to the change from associated data types to associated type synonyms.

Since this is backwards/forwards incompatible, the cabal constraint is currently `< 1`. At some point this should be changed to `> 1` presuming no issues with the usual dependency hell for downstream consumers of this library.

shlevy:
Fixed by #56: dependency hell addressed by following PVP

PR #56, opened 2015-01-09
amazonka: Build against monad-control >=1

shlevy:


brendanhay:
So this is a slightly contentious one, which relates to issue #40. I personally have problems regarding my companies internal dependencies not supporting `>= 1`.

I'd like to know if this is isolated - or if this risks breakage for others? In which case some conditionals/CPP might be the only option.

Opinion?

shlevy:
Not completely sure I understand the question. I'm pretty sure that anything that depends on `amazonka` after this is merged in will not be able to depend on `monad-control` `<1`, so if we need to support both of those then CPP seems like the only option.

Should I go ahead with a conditional version of this patch then?

brendanhay:
Yes, exactly. I'm not in a position to upgrade currently hence the open ticket.

If you'd like to go ahead and add a conditional version that would be much appreciated.

brendanhay:
On second thoughts just bumping the version according to the PVP should be sufficient.

shlevy:
OK, pushed a version bump commit.

brendanhay:
The `*.cabal` file versions outside of `amazonka/amazonka.cabal`, `amazonka/amazonka-gen.cabal` and `core/amazonka-core.cabal` are controlled via `gen/src/Types.hs`. You can see an example version bump commit here: https://github.com/brendanhay/amazonka/commit/b91e638a09ee656d041acac0932162c7711c63cb

That version is then rendered into all the example and library cabal files.

shlevy:
Force-pushed a new version bump

brendanhay:
Thanks

PR #58, opened 2015-01-15
ghc-7.8.4 comes with time-1.4.x.  Do we really need time 1.5.x?

dysinger:
Requiring time-1.5.x creates some cabal hell for some projects.


dysinger:
never mind.  the compilation just failed.  that answers my question.  back to cabal hell & --force-reinstalls :|

brendanhay:
The whole `time` + `old-locale` shenannigans is highly annoying - I've stuck with `1.5` due to not wanting to pass flags to the dependencies (`aeson` specifically) and to avoid using `CPP` etc.

PR #64, opened 2015-01-18
Not in the 2014-10-01 spec but I did receive this error today:

dysinger:
SerializerError "EC2" "Failed reading: Failure parsing
OfferingTypeValues from \"no upfront\""

Adding the no upfront type fixed the error.

Amazon did change their pricing model for reserved instances in Nov/Dec
time frame.  I can't remember the exact date.  They now offer "no
upfront" RIs where you pay a reduced rate if your machines are long
lived.

### DIFF
--- a/amazonka-ec2/gen/Network/AWS/EC2/Types.hs
+++ b/amazonka-ec2/gen/Network/AWS/EC2/Types.hs
@@ -9836,6 +9836,9 @@ data OfferingTypeValues
     = HeavyUtilization  -- ^ Heavy Utilization
     | LightUtilization  -- ^ Light Utilization
     | MediumUtilization -- ^ Medium Utilization
+    | NoUpfront         -- ^ No Upfront
+    | PartialUpfront    -- ^ Partial Upfront
+    | AllUpfront        -- ^ All Upfront
       deriving (Eq, Ord, Show, Generic, Enum)
 
 instance Hashable OfferingTypeValues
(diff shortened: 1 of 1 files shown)


dysinger:
You don't want to merge this but I just wanted to show a situation where we have live amazon APIs doing things not in the spec for EC2 2014-10-01

dysinger:
You can see the new types in their command line tools docs also

http://docs.aws.amazon.com/AWSEC2/latest/CommandLineReference/ApiReference-cmd-DescribeReservedInstancesOfferings.html

dysinger:
These 3 new flex RI types are not mentioned here in the latest api docs http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ReservedInstancesOffering.html

brendanhay:
Thanks for investigating that, very helpful! We'll have to get the corrections merged into the upstream service definitions and see if/why there is a reason for the live production API is not supported by even the [official SDK](https://github.com/aws/aws-sdk-java/blob/a97f96595e19327142eeaf4e744f6c6b1b766bb7/aws-java-sdk-ec2/src/main/java/com/amazonaws/services/ec2/model/OfferingTypeValues.java#L54).

brendanhay:
In the meantime I will add an additional override field allowing additional enumeration values to be specified in `gen/overrides/<service_name>.json`.

brendanhay:
Let me know if #65 is sufficient in the interim.

PR #65, opened 2015-01-19
Add sum_values field to type overrides

brendanhay:
Allows manual specification of additional nullary constructor branches.

Includes an initial change to the EC2 type `OfferingTypeValues` which was the motivating example for this PR.

Fixes #64.


brendanhay:
I'm going to merge this and create a release due to the latest ElasticTranscoder and CloudHSM additions. If it's insufficient for whatever reason it can be altered and re-released as a hotfix if necessary.

PR #69, opened 2015-01-20
Derive Read instances for all suitable types

### LINKED ISSUE #67: Read instances on most datatypes
It would be great to have Read instances on a lot of these datatypes, especially enumerations such as `Network.AWS.S3.Types.ObjectStorageClass`.

brendanhay:
Adds `Read` instance derivation to the generator and renders deriving clauses for the appropriate upstream types.

Fixes #67.


brendanhay:
There is an issue where `(Read a, Show a) => Sensitive a` does not satisfy `read (show x) = x`. For now this is left as something to fix in the future.

PR #71, opened 2015-01-24
These attributes need to be CaptialCase not mixedCase.

dysinger:
The JSON spec file is wrong here on the 'locationName' for these attributes.  The EC2 API spec page shows the proper
CapitalCase style.  http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ModifyReservedInstances.html I get errors modifying reserved instances with specific mention of the unknown mixedCase attributes.  If I change them to CapitalCase, then I can use the EC2 API without complaints.

You probably don't want to accept this P/R as is (because I just edited the Boto json files).  Tell me how to fix this outside of the 3rd party JSON so I can fix my P/R.

### DIFF
--- a/gen/input/ec2/2014-10-01.normal.json
+++ b/gen/input/ec2/2014-10-01.normal.json
@@ -8083,7 +8083,7 @@
         "ClientToken":{
           "shape":"String",
           "documentation":"<p>A unique, case-sensitive token you provide to ensure idempotency of your modification request.</p>",
-          "locationName":"clientToken"
+          "locationName":"ClientToken"
         },
         "ReservedInstancesIds":{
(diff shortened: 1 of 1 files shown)


brendanhay:
Off the top of my head I think the `camelCase` identifiers are exclusive to the `XML` (de)serialisation, while `PascalCase` is used for any `Query` serialisation.

I have the code for this already, and will submit a PR shortly.

brendanhay:
Apologies for 'shortly' meaning nearly a day.

If you have the bandwidth, can you visually once over the changes to the [EC2 output](https://github.com/brendanhay/amazonka/pull/72) to ensure your changes are accounted for? Otherwise, no problem.

I'll run some CI and then look at merging #72.

Issue #79, opened 2015-02-09
Core depends on unordered-containers >0.2.3


tathougies:
amazonka-core depends on unordered-containers >0.2.3 but undordered-containers before 0.2.5 does not have a Read instance for HashMap. This Read instance is needed to instantiate Read for Network.Internal.AWS.Data.Internal.Map.

brendanhay:
I've released your fix as part of [0.3.0](http://hackage.haskell.org/package/amazonka-core), hope that helps.

tathougies:
Thanks for the quick turnaround. It helps a lot!

Travis

PR #80, opened 2015-02-09
Upped undordered-containers dependency to 2.5 for Read HashMap instance

tathougies:
Solves #79


brendanhay:
Thanks for identifying and fixing that!

PR #92, opened 2015-04-11
Fix incorrect parsing of region

geraud:


core/src/Network/AWS/Types.hs
```diff
@@ -439,8 +439,8 @@ instance FromText Region where
         "ap-southeast-2"     -> pure Sydney
         "cn-north-1"         -> pure Beijing
         "us-east-1"          -> pure NorthVirginia
-        "us-west-2"          -> pure NorthCalifornia
-        "us-west-1"          -> pure Oregon
+        "us-west-2"          -> pure Oregon
+        "us-west-1"          -> pure NorthCalifornia
```
brendanhay: Nice catch :) Thanks!

brendanhay:
:+1:

brendanhay:
Your fix has been released as part of version [0.3.4](hackage.haskell.org/package/amazonka-0.3.4) on Hackage.

geraud:
Thanks a lot for the quick turn around!

PR #97, opened 2015-04-16
Make ErrorCode optional

kim:
Motivating example:

```haskell
> runAWST aws $ sendCatch (describeStream "foo")
Right (Left (SerializerError "Kinesis" "key \"code\" not present:\n{\"__type\":\"ResourceNotFoundException\",\"message\":\"Stream foo under account XXX not found.\"}"))
```

This change would require re-generating the bindings for _every_ `amazonka-*` library. Let me know if that should be included in the PR.


brendanhay:
Seems sane, thanks.

Rule of thumb is to not include any generated (or vendored) definitions, generator output, or version bumps in the PRs as I have CI/automagickation that does this and can be considered the 'source of truth' for keeping everything in sync.

PR #103, opened 2015-05-01
Fix amazonka-ec2 to accommodate the change in type of awsErrorCode from ...

ryantrinkle:
...34ef8129264da230211145112968ce9950a0dd87


brendanhay:
Thanks for catching that! :+1:

PR #104, opened 2015-05-01
Convert MonadAWS from a constraint alias to a full typeclass

ryantrinkle:
The changes I've made in this patch are designed to be as minimal and backwards-compatible as possible.  If you're willing to make a more breaking change, it might be preferable to change AWST to *not* derive its MonadReader instances and such, and instead pass them through from below.  This would break any outstanding code that relied on those instances, though.

### DIFF
--- a/amazonka/src/Control/Monad/Trans/AWS.hs
+++ b/amazonka/src/Control/Monad/Trans/AWS.hs
@@ -6,6 +6,7 @@
 {-# LANGUAGE RankNTypes                 #-}
 {-# LANGUAGE TupleSections              #-}
 {-# LANGUAGE TypeFamilies               #-}
+{-# LANGUAGE DefaultSignatures          #-}
 
 -- This is required due to the MonadBaseControl instance.
 {-# LANGUAGE UndecidableInstances       #-}
@@ -132,17 +133,6 @@ type Error = ServiceError String
 -- | A convenient alias for 'AWST' 'IO'.
 type AWS = AWST IO
 
--- | Provides an alias for shortening type signatures if preferred.
---
--- /Note:/ requires the @ConstraintKinds@ extension.
-type MonadAWS m =
-    ( MonadBaseControl IO m
-    , MonadCatch m
-    , MonadResource m
-    , MonadError Error m
-    , MonadReader Env m
-    )
-
 -- | The transformer. This satisfies all of the constraints that the functions
 -- in this module require, such as providing 'MonadResource' instances,
 -- and keeping track of the 'Env' environment.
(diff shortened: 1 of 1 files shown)


brendanhay:
Thanks @ryantrinkle.

I'm going to sit on this for a little bit, while some other issues are ironed out - is it a problem for you if it's not uploaded to Hackage?

ryantrinkle:
Not at all.  I use Nix, so sitting on a fork, even for a protracted period, is pretty easy.  However, it would be great to get your feedback on the basic approach as soon as you can offer it - I'd rather not depend on this change if you decide to go in a different direction.

brendanhay:
Hi @ryantrinkle. Apologies for taking so long. I've recently explored an alternative to the current transformer interface based on feedback from some other users. The result is a specialisable constraint based approach without actually providing a transformer proper. I think this would have solved your original problem - can you take a look and let me know if it looks usable for you? It's not final, I'll keep working on it for a bit, but please let me know any feedback when you get a chance.

As an example, given a custom application `MyApp` with it's own environment type, error type, and base Monad:

```haskell
-- The requirements for this to fulfill the constraints are that it can specialise to:
--
-- (MonadReader r m, AWSEnv r) -- For some reader with environment 'r', a lens exists
--   to obtain the AWS specific Env out of 'r'.
--
-- (MonadError  e m, AWSEnv e) -- For some error 'e', a prism exists to obtain the
--   AWS specific Error out of 'e'.
--
newtype MyApp a = MyApp
    { unApp :: ExceptT MyErr (ReaderT MyEnv (ResourceT IO)) a
    } deriving ( Functor
               , Applicative
               , Monad
               , MonadIO
               , MonadThrow
               , MonadBase IO
               , MonadCatch
               , MonadError  MyErr
               , MonadReader MyEnv
               , MonadResource
               )

runApp :: MyEnv -> MyApp a -> IO (Either MyErr a)
runApp e m = runResourceT $ runReaderT (runExceptT (unApp m)) e

-- A custom application environment for whatever Monad stack you're using.
data MyEnv = MyEnv
    { _config :: Int
      -- This is the actual Network.AWS.Env configuration.
    , _env    :: Env
    }

-- This class adds a lens pointing to the Network.AWS.Env in the user's
-- custom MyEnv type.
instance AWSEnv MyEnv where
    env f s = f (_env s) <&> \a -> s { _env = a }

-- A custom error for whatever application, containing a single constructor
-- that wraps the AWS errors.
data MyErr where
    EndpointDisabled :: MyErr
    EndpointNotFound :: Text  -> MyErr
    NoEndpointArn    :: MyErr
    NoToken          :: Text  -> MyErr
    ReadError        :: Text  -> String -> MyErr
    InvalidArn       :: Text  -> String -> MyErr
    -- The actual Network.AWS.Types.Error is embedded here:
    GeneralError     :: Error -> MyErr

-- This class adds a prism to point to the Network.AWS.Types.Error
-- in the user's custom MyErr type.
instance AWSError MyErr where
    _Error = prism
        GeneralError
        (\case GeneralError e -> Right e
               x              -> Left  x)

-- Control.Monad.Error.Lens can be used to catch AWS specific errors
catching _ServiceError $ MyApp (send Bar)
  :: (ServiceError -> MyApp Bar) -> MyApp Bar
```

The initial constraint based code is in the `feature/ast` branch, here: [Control.Monad.Trans.AWS](https://github.com/brendanhay/amazonka/blob/fa87a87eef0cbe865732158a9e5f9ef9c57801ae/amazonka/src/Control/Monad/Trans/AWS.hs#L141).

proger:
Guys I don't really understand the wins from converting to a proper typeclass, could you help me out please?

brendanhay:
You will have to elaborate on what you mean by a 'proper typeclass'. Are you referring to something like `MonadSnap` and a mechanism like `liftSnap`?

The original issue was people using `ExceptT` for their own error handling with a custom type, which conflicts with the `MonadError ServiceError` instance for `AWST` due to the functional dependency, and the subsequent `newtype` wrapping / lifting. (Similarly for `MonadReader Env`.) Or alternatively having to rebuild similar functionality from scratch using `Network.AWS` so it embeds nicely in your own stack.

The above code I shared is simply a variant which does away with a `newtype` for the `AWST` instance and instead denotes constraints on the various functions in the `Control.Monad.Trans.AWS` module.

These constraints specialise in the simplest case to [`ExceptT Error (ReaderT Env (ResourceT IO)`](https://github.com/brendanhay/amazonka/blob/fa87a87eef0cbe865732158a9e5f9ef9c57801ae/amazonka/src/Control/Monad/Trans/AWS.hs#L105), and in the most general case to:

```
    ( MonadResource   m
    , MonadReader   r m
    , MonadError    e m
    , AWSEnv        r
    , AWSError      e
    )
```

This means rather than having to provide all possible permutations of `mtl` styled classes (and potentially incur other dependencies), whatever instances exist for `MonadReader r m` and `MonadError e m` can be used, as long as you provide a `Lens r` and a `Prism e` to obtain/wrap `Env` and `Error` respectively.

It also ensures the functions in that module compose with whatever your application stack might be, given it fulfills those preceding constraints, and you don't have to use `lift` or another specialised variant.

Does that make sense, do you have an alternative suggestion?

brendanhay:
`develop` now contains a modified `AWS` + `AWST` similar to the approach I outlined above which should be sufficient to solve the above issues. You can find the pre-release documentation [here](http://www.brendanhay.nz/amazonka-doc/amazonka/index.html).

The changes will stew in `develop` for a while, so please check them out and let me know any feedback so I can incorporate it pre `1.0`.

PR #105, opened 2015-05-04
Use locationName for XML list over 'member'

charleso:
We discovered that SQS.ReceiveMessageResponse wasn't deserialising from XML. It turns out `List "member"` should have been `List "Message"`.

https://github.com/brendanhay/amazonka/blob/1adc27543f335341059b96e952f4f7c34abb500b/amazonka-sqs/gen/Network/AWS/SQS/ReceiveMessage.hs#L190

It looks like amazonka should be using the [locationName](https://github.com/brendanhay/amazonka/blob/develop/gen/input/sqs/2012-11-05.normal.json#L1115) from the `MessageList` shape.

I fear that this change will break some/all `ToQuery` instances because they (I believe) share the same information, and I suspect that `member` was correct. Will this require different/extra generation for one of both of `ToQuery` and `FromXML` in regards to `List`?

### DIFF
--- a/gen/src/Gen/AST.hs
+++ b/gen/src/Gen/AST.hs
@@ -680,11 +680,12 @@ shapes proto time m =
             flatten = x ^. lstFlattened
                   <|> x ^. lstMember . refFlattened
 
-            ann | proto == Query = "member"
-                | otherwise      =
+            ann =
                     fromMaybe fld $
                             x ^. lstMember . refLocationName
                         <|> r ^. refLocationName
+                        -- Fallback to 'member' only if there is no locationName
+                        <|> if proto == Query then Just "member" else Nothing
 
     flat :: Maybe Bool -> Type -> Type
     flat _ | proto == Ec2 = TFlatten


brendanhay:
Thanks for raising the issue! Can you please re-formulate the PR _without_ the generated code? 

Regarding breakage of `ToQuery` instances, I'll try and dig out some counter-examples so we identify the correct model to use, whether only `Protocol`, absence/presence of `locationName`, or another facet like `serviceAbbrev` is taken into account to calculate the actual list element name.

For this the quickest approach is probably going to be to look through the other SDKs and see how they are differentiating, which I will try and do today sometime.

charleso:
@brendanhay Have force-pushed away my second commit. Sorry about taht.

I find the go APIs interesting, as they seem to have `member` for the query, and the `locationName` for the XML. I couldn't say whether that's actually "correct" though:

https://gowalker.org/github.com/hashicorp/aws-sdk-go/gen/sqs#MessageAttributeValue

charleso:
Yeah maybe ignore the Go version. I should have been looking at the Java API.

At least as far as ReceiveMessage goes, it looks like my change would have been correct:

https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-sqs/src/main/java/com/amazonaws/services/sqs/model/transform/ReceiveMessageRequestMarshaller.java#L53

I'll try to find some cases where "member" is used in Java but wouldn't be with this change.

charleso:
I've spot checked a random selection of the java uses of ".member" and it seems like they all line up still with the generated code from this change. I'm sure you'll want to do your own investigation too. Let me know if there's anything I can do on this end to help with this PR.

brendanhay:
The AWS maintained Go SDK is [here](https://github.com/awslabs/aws-sdk-go/blob/876a0d58f03015fb59eb3667c6e846590029bacd/service/sqs/api.go#L1282), along with the [Go ReceiveMessageOutput](
https://github.com/awslabs/aws-sdk-go/blob/876a0d58f03015fb59eb3667c6e846590029bacd/service/sqs/api.go#L1412).

Also the [.NET Marshaller](https://github.com/aws/aws-sdk-net/blob/bfe325af6eab8e36df9b68313b4774d3eda87c0d/AWSSDK_DotNet35/Amazon.SQS/Model/Internal/MarshallTransformations/SendMessageRequestMarshaller.cs#L63) and [.NET Unmarshaller](https://github.com/aws/aws-sdk-net/blob/bfe325af6eab8e36df9b68313b4774d3eda87c0d/AWSSDK_DotNet35/Amazon.SQS/Model/Internal/MarshallTransformations/ReceiveMessageResponseUnmarshaller.cs#L65).

I'll keep digging. Thanks again!

charleso:
Regarding the Go SDK, the only interesting mention of `member` is in the [queryutil.go#parseList](parseList) function. It's only done for unflattened lists.

charleso:
Awesome. So the Java and .NET APIs are different in some of these edge cases.

https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-redshift/src/main/java/com/amazonaws/services/redshift/model/transform/ModifyClusterParameterGroupRequestMarshaller.java#L55

and 

https://github.com/aws/aws-sdk-net/blob/bfe325af6eab8e36df9b68313b4774d3eda87c0d/AWSSDK_DotNet35/Amazon.Redshift/Model/Internal/MarshallTransformations/ModifyClusterParameterGroupRequestMarshaller.cs#L62

Looks like the Java one aligns with this PR change (not that it makes it correct):

https://github.com/ambiata/amazonka/blob/gen/xml-list-member-name/amazonka-redshift/gen/Network/AWS/Redshift/ModifyClusterParameterGroup.hs#L56

My money is on the Java version being correct.

brendanhay:
The service definitions are generated from something akin to the documentation, and the documentation naturally aligns with the .NET version: http://docs.aws.amazon.com/redshift/latest/APIReference/API_ModifyClusterParameterGroup.html - I'll raise the issue upstream for clarification.

My past-self remembers something along the lines of the Go (both SDKs) behaviour, in that; if the protocol is `EC2` or the list/map `Shape` is annotated with `"flattened": true` the query serialisation uses `"member"` as the serialised item name. Otherwise it should use the `locationName`. For XML serialisation it always uses `locationName`.

brendanhay:
And for documentation purposes, it's again worth pointing out that `EC2` and `S3` both have edge-cases (relative to the other services) regarding serialisation.

charleso:
Yeah I saw something hard-coded about EC2 in the generator, which I assumed was unfortunate but necessary.

Just to clarify - is there anything I can/should be doing at this point? I'm assuming that only one representation is "correct"? In that - if you use the wrong member/localName it will throw an error. In which case, I guess we should pick one that's out of sync and test it manually?

brendanhay:
Yes in that there should be one 'correct' representation which we can encode in either the generator, types, or both.

Manual testing can help, but generally there are too many affected services and it's usually easier to obtain clarification from Amazon and determine whether the documentation is correct and which SDKs are correct (or not).

I plan to reach out to Amazon/AWS to clarify what the correct behaviour is regarding the `locationName` and `flattened` facets pertaining to both `Query` and `XML` serialisation.

Hopefully we'll have an answer by end of day GMT, I'll keep this PR/issue updated.

charleso:
Thanks @brendanhay! I appreciate the help.

I'm a huge fan of what you've done with this project - we are/were using the handcraft `aws` and it was a relief to find your library. :)

brendanhay:
No problem mate. I'm glad you raised the issue - it's pretty hard to personally get usage/coverage across all the services on my lonesome, and I need users to be able to identify and iron out problems!

brendanhay:
Just a quick update - I still haven't heard back from Amazon, so I've reached out to another contact.

brendanhay:
I received the following information from Amazon:

Effectively, non-flattened lists use the “member” by default unless there is a locationName trait present on the member. If it’s flattened, the actual list member name itself is used (or locationName on that item if present). There is fairly good test coverage for this in the specs.

As for EC2, the query and ec2 protocols are slightly different. Here’s some currently-internal documentation we use to differentiate the EC2 protocol from Query that may help answer your questions:

* EC2 serialization remains very similar to Query serialization. The differences are:
Query supports flattened and non-flattened lists. EC2 only has lists that are effectively always flattened.
* EC2 does not have input maps
* The "locationName" must always be up-cased (first letter only) when used.
* When choosing the query name for a list, the list member "locationName" is never used.

I haven't had a chance to incorporate this feedback yet.

charleso:
Hi Brendan. Lazy question - how close was I in my change? "When choosing the query name for a list, the list member "locationName" is never used." sounds like the Java versions are wrong, but let me know if I'm misunderstanding that comment.

brendanhay:
You were semi-correct, at least within the bounds of the poor abstractions I had provided :smile:

I've written up a sort of truth-table to outline my understanding from going over the specs and the comments above, which is in a newer version of the generator not available yet:

```haskell

listInputName :: Protocol  -- ^ Service protocol
              -> Info      -- ^ Shape info
              -> (Id, Ref) -- ^ The struct member Id and Ref
              -> Ref       -- ^ The list item the struct member ref points to's member ref
              -> (Text, Maybe Text)
listInputName p i (k, v) e = (input p flat, output p flat)
  where
    flat = i ^. infoFlattened

    -- Use the locationName on the struct member if present,
    -- otherwise the struct member id.
    parent = fromMaybe (k ^. memberId) (v ^. refLocationName)

    -- Use the locationName on list element if present,
    -- otherwise defualt to 'member'.
    element = fromMaybe "member" (e ^. refLocationName)

    input Query    False  = (parent, Just element)
    input Query    True   = (parent, Nothing)
    input EC2      _      = (upperHead $ fromMaybe parent (v ^. queryName), Nothing)
    input JSON     _      = (parent, Nothing)
    input RestJSON _      = (parent, Nothing)
    input RestXML  True   = (parent, Nothing)
    input RestXML  False  = (parent, Just element)

    output Query    False = (parent, Just element)
    output Query    True  = (parent, Nothing)
    output EC2      False = (parent, Just element)
    output EC2      True  = (parent, Nothing)
    output JSON     _     = (parent, Nothing)
    output RestJSON _     = (parent, Nothing)
    output RestXML  True  = (parent, Nothing)
    output RestXML  False = (parent, Just element)
```

Is that understandable?

brendanhay:
So more or less:

* Take the `locationName` from the struct field for the 'parent', if it exists, otherwise using the actual struct field's name.
* Take the `locationName` from the list item itself, for the 'element', if it exists, otherwise using "member".

And the edge-cases:

* Rest/JSON is simple!
* EC2 is always flattened + possible `queryName` element + `upperHead`

charleso:
@brendanhay Wow that's a little more complicated that I imagined! Your truth table helps, and obviously looks like you'll need to retain/compile more information. I would be curious to see the intersection of what is (still) broken by my change, just because I haven't got a clear mental model of what would be affected.

brendanhay:
The above commit/fix is actually dependent on an answer from https://github.com/boto/botocore/issues/602.

PR #110, opened 2015-05-11
Increase EC2 metadata timeout to 2 seconds

charleso:
I'm assuming it wasn't intentional to make it 2 _microseconds_? What was the expected value?

### DIFF
--- a/amazonka/src/Network/AWS/EC2/Metadata.hs
+++ b/amazonka/src/Network/AWS/EC2/Metadata.hs
@@ -325,5 +325,5 @@ get m url = ExceptT . liftIO $ req `catch` err
 request :: Manager -> Text -> IO ByteString
 request m url = do
     rq <- parseUrl (Text.unpack url)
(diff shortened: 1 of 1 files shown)


brendanhay:
Hah, no it should not be 2 microseconds! :smile: The joy of `threadDelay` typos.

brendanhay:
Looking at the code, I think in hindsight it's a poor idea to update the timeout value when the function takes a `Manager`, since it prevents anyone from setting their own timeout. While in my experience `30s` is too high to retrieve instance metadata, I'd probably rather remove it and let people specify whatever they want.

charleso:
@brendanhay Happy to remove the timeout completely if you're OK with that?

brendanhay:
Go for it!

brendanhay:
:+1:

charleso:
@brendanhay Have updated, let me know if that looks OK. Also happy to add a comment somewhere you think is appropriate.

PR #123, opened 2015-06-25
gen: create AWSPager instance for all request types

proger:
This allows code to uniformly use conduits with `paginate`.

### DIFF
--- a/gen/templates/_include/pager.ede
+++ b/gen/templates/_include/pager.ede
@@ -36,4 +36,8 @@ instance AWSPager {{ request.name }} where
         {{ token.key }} = rs ^. {{ token.value.output }}
    {% endfor %}
   {% endcase %}
+{% else %}
+
+instance AWSPager {{ request.name }} where
+  page _ _ = Nothing
 {% endif %}


brendanhay:
The intent with `paginate/AWSPager` vs `send/AWSRequest`  was to make it clear at the type/constraint level whether an operation returns multiple results or not. If that difference doesn't exist then a user would naively have to assume _every_ operation potentially has multiple results and be prepared to handle this case - effectively rendering `paginate` as the only safe way to exhaustively handle the remote API in the naive case.

I'd rather think about introducing a separate way to abstract over the one-or-many case to keep the above distinction apparent, if that makes sense?

proger:
The idea of having a separate type explicitly specify that the request is subject to pagination is quite sound, however I'm afraid it's going to drag boilerplate code to every project that uses lots of `Describe*` actions (looks like not every `Describe` action supports pagination — which is a bit weird). 

So far the easiest way for me was to simply derive 4-5 orphans.

I'm going to close this and probably reopen if I come up with something smart that's upstreamable.

brendanhay:
I recently explored making more of the logic [configurable](https://github.com/brendanhay/amazonka/blob/fa87a87eef0cbe865732158a9e5f9ef9c57801ae/amazonka/src/Control/Monad/Trans/AWS.hs#L148), such as endpoints, signers, and so forth. Perhaps a similar thing could be achieved with the pagination? That way you could either use the implicit pagination logic via the `AWSPager` class if it's available, or provide some general mechanism for paging.

A completely incorrectly typed example:

```haskell
send :: AWSRequest a => a -> Rs a
send rq = ...

paginate :: AWSPager a => a -> Source (Rs a) ()
paginate rq = paginateWith (pager rq) rq

paginateWith :: AWSRequest a => (a -> Rs a -> Maybe a) -> a -> Source (Rs a) ()
paginateWith f rq = ...
```

That way I guess you could define your own class + instances like:

```haskell
class AWSRequest a => MyPager a where
    myPager :: a -> Rs a -> Maybe a

instance MyPager RunInstance where
    myPager x _ = Just x

myPaginate :: MyPager a => a -> Source (Rs a) ()
myPaginate rq = paginateWith (myPager rq) rq
```

Although after typing all that out, it does seem somewhat boilerplate-y anyway.

PR #136, opened 2015-07-24
Re-add Functor/Applicative constraints for GHC 7.8 compatibility

charleso:
@brendanhay Just starting to looking to upgrade our version of amazonka and this is the first thing we hit.

We've hit some other snags around the transformers, but that's a problem for another day...

### DIFF
--- a/amazonka/src/Control/Monad/Error/AWS.hs
+++ b/amazonka/src/Control/Monad/Error/AWS.hs
@@ -69,11 +69,12 @@ runAWST :: (MonadCatch m, MonadResource m, AWSEnv r)
 runAWST e = runExceptT . AWST.runAWST e
 {-# DEPRECATED runAWST "Exists for backwards compatibility with pre-@1.0@." #-}
 
-hoistError :: (MonadError e m, AWSError e) => Either Error a -> m a
+hoistError :: (Applicative m, MonadError e m, AWSError e) => Either Error a -> m a
 hoistError = either (throwing _Error) pure
 
 -- | /See:/ 'AWST.metadata'
 metadata :: ( MonadFree Command m
+            , Applicative m
             , MonadError e m
             , AWSError e
             )
@@ -83,6 +84,7 @@ metadata = AWST.metadata >=> hoistError . first HTTPError
 
 -- | /See:/ 'AWST.dynamic'
 dynamic :: ( MonadFree Command m
+           , Applicative m
            , MonadError e m
            , AWSError e
            )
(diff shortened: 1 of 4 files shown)


brendanhay:
Wow! I didn't even know it built on `6.8` :smile:

brendanhay:
If this is a supported use case, I'll some `6.8` buildsto the `travis.yml`.

charleso:
@brendanhay DOH! I mean 7.8. I'm an idiot...

charleso:
Now I'm wondering how the travis builds are passing?

https://travis-ci.org/brendanhay/amazonka/jobs/72418403

brendanhay:
Aha, good catch. Could it be that since `amazonka` proper doesn't have a `test-suite` declaration in the cabal file, it's not actually being built on CI due to `stack setup/stack test`? If so, that's an explosively wet brain fart on my part.

charleso:
Hmm. I couldn't say. You can see 'build' in the build output above, which suggests to me it is doing _something_

brendanhay:
I'll spend some time today verifying it fails _without_ your PR, before merging.

charleso:
No worries, and thanks. I'm just trying to build in isolation with stack.

brendanhay:
So it indeed turned out to be the lack of a `test-suite` in the `amazonka.cabal` configuration. By adding a basic `test-suite`, the build has started erroring: https://travis-ci.org/brendanhay/amazonka/jobs/72423722

Thanks for bringing the falsch CI pancea to my attention.

amazonka/src/Control/Monad/Error/AWS.hs
```diff
@@ -69,11 +69,12 @@ runAWST :: (MonadCatch m, MonadResource m, AWSEnv r)
 runAWST e = runExceptT . AWST.runAWST e
 {-# DEPRECATED runAWST "Exists for backwards compatibility with pre-@1.0@." #-}
 
-hoistError :: (MonadError e m, AWSError e) => Either Error a -> m a
+hoistError :: (Applicative m, MonadError e m, AWSError e) => Either Error a -> m a
 hoistError = either (throwing _Error) pure
```
brendanhay: Would changing the use of `pure` to `return` remove the need for the `Applicative` constraint entirely?

charleso: > Would changing the use of pure to return remove the need for the Applicative constraint entirely?

By the looks of it - yes. Sorry, I wasn't paying attention. That's a good idea.

brendanhay:
I've merged and made the `pure` -> `return` switch in `develop`.

charleso:
Thanks @brendanhay

PR #137, opened 2015-07-24
Add mechanism to conditionally massage requests before sending, per service configuration

### LINKED ISSUE #115: S3 DeleteObjects is missing required Content-MD5 header
AFAICT this is a problem with the upstream json definition

brendanhay:
Allows the conditional insertion of `Content-MD5` headers via a general preflight hook, similar to how some of the other SDKs works. 

See: [Ruby SDK MD5](https://github.com/aws/aws-sdk-ruby/blob/17508581a6887fb170e608242165b1ee18ccd9a9/aws-sdk-core/lib/aws-sdk-core/plugins/s3_md5s.rb#L15)

Fixes #115.


brendanhay:
Important bits are [here](https://github.com/brendanhay/amazonka/blob/f6273af27a30b7cd0f259f91a065878e7c64cef0/amazonka-s3/src/Network/AWS/S3/Internal.hs#L101) (S3 Internal), [here](https://github.com/brendanhay/amazonka/blob/f6273af27a30b7cd0f259f91a065878e7c64cef0/amazonka-s3/gen/Network/AWS/S3/Types.hs#L458) (S3 Types) and [here](https://github.com/brendanhay/amazonka/blob/f6273af27a30b7cd0f259f91a065878e7c64cef0/amazonka/src/Network/AWS/Internal/HTTP.hs#L42) (HTTP send).

brendanhay:
Closed in favour of #138.

PR #139, opened 2015-07-24
Prefer url-form-encoded POST requests where supported

### LINKED ISSUE #134: SendRawEmail fails with large data
When I try to send a large payload (e.g. a 50k attachment) using `SendRawEmail`, I end up getting an HTTP 505 Version Not Supported error from AWS.

It looks like `SendRawEmail` is putting the the entire request into the query string rather than the POST body, which I suspect is the problem.

brendanhay:
A majority of the `Query` protocol APIs support both `GET` and `POST` methods for their operations. `GET` requests have limitations according to the maximum URL length.

This change pushes the result of `ToQuery/toQuery` into the form body and sets the `Content-Type` header appropriately for all `POST` requests for `Query` protocol services.

The impact of this change is not quantified, in the sense the generated `ToQuery` instances are not granular and can potentially contain valid query components that should not be pushed into the form body (the service definitions don't differentiate) - although it appears at first glance this only applies to `json` + `xml` protocols.

More verification needs to be done before merging.

Fixes #134.


brendanhay:
I've verified this is in fact correct, testing still needs to be done.

Issue #143, opened 2015-07-26
ObjectKey opacity and combinators


brendanhay:
In `develop` I introduced the `ObjectKey` `newtype` instead of exposing raw `Text` in `amazonka-s3`. This was intended to make it easier to transition to + work on a fix for #127.

There are a couple of annoyances:

* Lack of a `Monoid` instance makes path manipulation difficult.
* The specific delimiter used needs to be known at encode/decode time.

Since the delimiter is only specified as part of operations such as `ListObject`, and there is no way to obtain that information during the encoding of the `ObjectKey` it looks likely it will need to also be contained in the `ObjectKey` data type.

Perhaps some combinators can be provided which assume a `/` delimiter, to make the default case as simple as possible and for those that use a custom delimiter, probably some less convenient but more exposed API can be offered.

brendanhay:
Since any forms of encoding/decoding rely on the user specifying the exact delimiter to be used, this is one approach off the top of my head:

```haskell
type Delimiter = Char

data ObjectKey
    = Decoded Char [ByteString]
    | Encoded ByteString

data ObjectKey = ObjectKey Delimiter [ByteString]
    deriving (Eq, Ord, Read, Show, Data, Typeable, Generic)

instance FromText ObjectKey where
    parser = Encoded . Text.encodeUtf8 <$> takeText

instance FromXML ObjectKey where
    parseXML = parseXMLText "ObjectKey"

instance ToByteString ObjectKey where
    toBS = encodeKey

instance ToText    ObjectKey where toText   = toText  . toBS
instance ToQuery   ObjectKey where toQuery  = toQuery . toBS
instance ToBuilder ObjectKey where build    = build   . toBS
instance ToXML     ObjectKey where toXML    = toXMLText

-- Would this cause more problems than it's worth?
instance IsString ObjectKey where
    fromString = objectKey '/'

-- no sensible Monoid instance, path combinators instead?

objectKey :: Char -> ByteString -> ObjectKey
objectKey c = Decoded c . BS8.split c

encodeKey :: ObjectKey -> ByteString
encodeKey = \case
    Decoded c xs -> BS8.intersperse c $ map (urlEncode False) xs
    Encoded bs   -> bs

decodeKey :: ObjectKey -> Either (Char -> ByteString) ByteString
decodeKey = \case
    Decoded c xs -> Right (BS8.intersperse c xs)
    Encoded bs   -> Left $ \c ->
        BS.intercalate (BS.singleton c) $
[...]

Issue #127, opened 2015-06-29
signature calculation failure for URLs with "=" in them


michaelneale:
Attempting to upload to something like: 

`s3://bks/a=1/file.txt`

will result in: 
"The request signature we calculated does not match the signature you provided. Check your key and signing method."

It appears that the `=` needs to be url encoded before being put in the header

brendanhay:
Hi @michaelneale,

In order to construct a contrived test-case - When you mention uploading, you are referring to `PutObject` specifically? And in regards to 'being put in the header', you're referring to the `Authorization` header specifically?

michaelneale:
@brendanhay yes, definitely was put - and yes I think it was in the auth header where it hurt (I was using various auth schemes, one of which was using STS tokens, so I am not sure if that affects what ends up in the header - I will check with @nhibberd today as he provided a work around.

michaelneale:
we had to do something like: 
```
encodeKey :: Key -> Text
encodeKey (Key k) =
  let splitEncoded = urlEncode True . encodeUtf8 <$> split (== '/') k
      bsEncoded = BS.intercalate "/" splitEncoded
  in
  decodeUtf8 bsEncoded
```

(assuming that '/' was the delimeter in our case)

michaelneale:
Actually we url encode the whole key before calling _send - so it may be more than just the Authorization header (basically anywhere the s3 key could appear in the header, would need encoding of non token parts).

michaelneale:
Hi @brendanhay - did some more investigation, this is for all verbs - put, get, head etc. all show the same behaviour. 

You can try it by having any key with `=` in the path (like s3://.../yeah=2015/month=07/ for example) and you will see it. I just happened to see it first with put. Strangely by URL encoding the whole key, it won't work, you have to leave the delimiters (/) unencoded (making me think something else encodes them). 

It will generally show up as a 404 or signature failure error of some kind. The key can appear in a header or the query string (which makes me think it is nothing to do with http conventions/restrictions but is at the amazon end). 

Sorry to not have more info - but easiest thing is to try it.

brendanhay:
Thanks, that's perfect.

michaelneale:
Alarmingly, not clear why the delimiters shouldn't be encoded (but other bits should be), but if you find out the theory behind it, would be good to know.

brendanhay:
[...]

PR #144, opened 2015-07-26
ObjectKey auto encoding and decoding

brendanhay:
In reference to #127 and #143.


brendanhay:
Closing due to soggy brain farts.

PR #148, opened 2015-07-28
route53 config: ResourceRecord can be empty in ResourceRecordSet

proger:
Example use: creating aliases, see http://docs.aws.amazon.com/Route53/latest/APIReference/CreateAliasRRSAPI.html


brendanhay:
Thanks!

PR #149, opened 2015-07-28
route53 config: ResourceRecord can be empty in ResourceRecordSet

proger:
Same as #148 for 0.3. Is it possible to have this released by chance?


brendanhay:
Sure

Issue #98, opened 2015-04-20
DynamoDB: omit empty fields from requests


kim:
DynamoDB seems to consider fields which are present, but empty, an error:

```haskell
> runAWST env . sendCatch $ createTable "test" (keySchemaElement "id" Hash :| []) (provisionedThroughput 10 10) & ctAttributeDefinitions .~ [attributeDefinition "id" S]

Right (Left (ServiceError "DynamoDB" (Status {statusCode = 400, statusMessage = "Bad Request"}) (JSONError {_jsonType = Just "com.amazon.coral.validate#ValidationException", _jsonCode = Nothing, _jsonMessage = "One or more parameter values were invalid: List of LocalSecondaryIndexes is empty"})))
```

```haskell
> runAWST env . sendCatch $ putItem "test" & piItem .~ (fromList [("id", attributeValue & avS .~ Just "someid")]) & piConditionExpression .~ Just ("id <> :id") & piExpressionAttributeValues .~ (fromList [(":id", attributeValue & avS .~ Just "someid")])

Left (ServiceError "DynamoDB" (Status {statusCode = 400, statusMessage = "Bad Request"}) "JSONError {_jsonType = Just \"com.amazon.coral.validate#ValidationException\", _jsonCode = Nothing, _jsonMessage = \"Can not use both expression and non-expression parameters in the same request: Non-expression parameters: {Expected} Expression parameters: {ConditionExpression}\"}")
```

Since I don't know about other AWS services, I'm not sure what the best solution for this would be. Perhaps wrap all optional fields in `Maybe` (even if they're lists) and omit `Nothing`-valued fields from the JSON?

kim:
I guess it might be necessary to introduce some other type (eg. `Optional`), as `Nothing == null` atm, which might be desired in the general case.

brendanhay:
Right, this seems (at present) only affect the JSON styled services due to how Aeson does the serialisaiton. Neither the XML nor Query serialisers are representing an empty list value in the request over the wire.

Potentially the `Optional` type can be internal in the following fashion:

```haskell
data Optional a
    = Present a
    | Absent

-- | Invalid Iso, should be a Prism but exists for ease of composition
-- with the current 'Lens . Iso' chaining to hide internal types from the user.
_Present :: Monoid a => Iso' (Optional a) a
_Present = iso f Present
  where
    f (Present x) = x
    f Absent      = mempty
```

1. The smart constructor for a data type sets any `Optional` fields to `Absent`:

[...]

PR #150, opened 2015-07-29
route53: explicitly mark ResourceRecords in RRS optional

proger:
Sorry I haven't tested a gen for #149. Looks like this is a constraint coming from the [input spec](https://github.com/brendanhay/amazonka/blob/0.3-series/gen/input/route53/2013-04-01.normal.json#L2942), so i added an appropriate override this time.

That should do it.

### DIFF
--- a/gen/overrides/route53.json
+++ b/gen/overrides/route53.json
@@ -15,6 +15,11 @@
                 "Marker"
             ]
         },
+        "ResourceRecordSet": {
+            "optional": [
+                "ResourceRecords"
+            ]
+        },
         "ResourceRecordSetRegion": {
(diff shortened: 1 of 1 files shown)


brendanhay:
So this is actually related to https://github.com/brendanhay/amazonka/issues/98#issuecomment-111815332 and has already been fixed in `develop` without the need for overrides. 

[amazonka-0.3.6.1](http://hackage.haskell.org/package/amazonka-route53-0.3.6.1) is now available on Hackage with the regenerated `ResourceRecordSet` type.

proger:
Thank you!

PR #164, opened 2015-08-09
Use standard AWS environment variables

adbrowne:
All SDKs now seems to use these variables: http://blogs.aws.amazon.com/security/post/Tx3D6U6WSFGOK2H/A-New-and-Standardized-Way-to-Manage-Credentials-in-the-AWS-SDKs

There might be a good reason to use something different. If so would you accept a PR that tries both sets of environment variables?

### DIFF
--- a/amazonka/src/Network/AWS/Auth.hs
+++ b/amazonka/src/Network/AWS/Auth.hs
@@ -67,12 +67,12 @@ import           System.Mem.Weak
 import           Prelude
 
 -- | Default access key environment variable.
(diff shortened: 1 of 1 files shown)


brendanhay:
No reason to use something different other than inertia. Quite happy to make the change, but I'm going to sit on the PR for a few days so others can be made aware.

Thanks for the link, and the PR.

adbrowne:
Thanks!

brendanhay:
OK, nevermind the wait. @1.0@ is imminent with quite a few breaking changes so this is as good a time as any to get it in there. Thanks again.

PR #167, opened 2015-08-10
Ensure nested ToXML instance fields receive a namespace if none specified

### LINKED ISSUE #166: route53: XML serializer does not retain namespaces properly
I've only spotted this in route53 for now,

Tested with a406e2a8ba and 0.3.6.1,

The error is:

```haskell
*Network.AWS.Route53 Control.Monad.Trans.AWS Network.AWS.Prelude Control.Monad Control.Monad.Trans.Resource Control.Lens>
_ <- runResourceT $ runAWST env' $ void $ send (
   changeResourceRecordSets "ZAAAAAAAAAAA" (changeBatch ((change Upsert (resourceRecordSet "www.example.com" A & rrsResourceRecords .~ Just (resourceRecord "127.0.0.1" :| []))) :| [])))

*** Exception: ServiceError (ServiceError' {_serviceAbbrev = Abbrev "Route53", _serviceStatus = Status {statusCode = 400, statusMessage = "Bad Request"}, _serviceHeaders = [("x-amzn-RequestId","92fb66ac-3eda-11e5-80ea-8d0638de187b"),("Content-Type","text/xml"),("Content-Length","438"),("Date","Sun, 09 Aug 2015 21:06:59 GMT"),("nnCoection","close")],
_serviceCode = ErrorCode "InvalidInput", _serviceMessage =
  Just (ErrorMessage "Invalid XML ; cvc-complex-type.2.4.a: Invalid content was found starting with element 'ChangeBatch'. One of '{\"https://route53.amazonaws.com/doc/2013-04-01/\":ChangeBatch}' is expected.")
 _serviceRequestId = Just (RequestId "92fb66ac-3eda-11e5-80ea-8d0638de187b")})
```

The XML that `xml-conduit` generates looks like:

```xml
<?xml version="1.0" encoding="UTF-8"?>
-- here amazonka generates code using the `namespaced` function:
<ChangeResourceRecordSets xmlns="https://route53.amazonaws.com/doc/2013-04-01/">
-- xml-conduit does smart namespace tracking and resets namespaces for all tags down:
  <ChangeBatch xmlns="">
    <OtherStuff/>
  </ChangeBatch>
</ChangeResourceRecordSets>
```

For 0.3.6.1 I implemented a [workaround by hardcoding specialized xml structure](https://github.com/zalora/upcast/blob/8b36fe99cdba3a9ca293d3be68d56bf865fb3845/src/Upcast/Infra/Amazonka/UpsertAlias.hs).

The proper fix i suppose would be to make generated recursively set namespaces for everything and I'm sorry I don't have a proper patch yet.

brendanhay:
By previously using either the `IsString Name` instance or constructing a name with `Nothing` for the XML namespace, the namespace would be rendered as `xmlns=""` thus generating invalid AWS XML.

This PR adds a some what performant fix during rendering of the XML document to ensure names have the root element's namespace, if none was specified.

Fixes #166.

### DIFF
--- a/amazonka-s3/fixture/DeleteObjects
+++ b/amazonka-s3/fixture/DeleteObjects
@@ -4,8 +4,8 @@ path:   /bucketname
 query:  ?delete=
 headers:
   Host: s3.amazonaws.com
-  Authorization: AWS4-HMAC-SHA256 Credential=access/20091028/us-east-1/s3/aws4_request, SignedHeaders=content-md5;host;x-amz-content-sha256;x-amz-date, Signature=444663bb6a12513999d350ae5d1e6fcf7b2a0893ea22582c6db4922edbb9c83b
-  Content-MD5: PdWUHsgAuOlP59d4mBjdGg==
-  X-Amz-Content-SHA256: 593bd30fd702e1eae6fd7d828b80ccdf3283194ef2632440511115e70cdf4ccc
+  Authorization: AWS4-HMAC-SHA256 Credential=access/20091028/us-east-1/s3/aws4_request, SignedHeaders=content-md5;host;x-amz-content-sha256;x-amz-date, Signature=8395d88802592a49a30ca19ee1e0a819a663542eddc9bfe6482e314f6ce09b58
+  Content-MD5: Baa58yLrqn+8X+KF6MwPug==
+  X-Amz-Content-SHA256: 3be5f080da935ef26665db14539913f2a45d0219d47414df168777f155b30070
   X-Amz-Date: 20091028T223200Z
-body: <?xml version="1.0" encoding="UTF-8"?><Delete xmlns="http://s3.amazonaws.com/doc/2006-03-01/"><Object xmlns=""><Key>sample1.text</Key></Object><Object xmlns=""><Key>sample2.text</Key></Object></Delete>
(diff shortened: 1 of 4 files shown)


brendanhay:
Not convinced this is the correct solution. The other option would be to propagate the namespace throughout the generator and therefore have `locationName`s rendered in clark notation, so that the `ToXML` instances have field assignment that would look like:

```
"{http://s3.amazonaws.com/doc/2006-03-01}Object" @= _doObject ...
```

proger:
Apart from clark notation (TIL) you could use xmlns prefixes (like `route53:ChangeBatch`, iirc you can set them as arguments in `RenderSettings`). However I though you were going to change the generator to drag namespaces around (without clark notation, xml-conduit just strips them to avoid repetition which results in generated xml look exactly the same except without the show-stopping `xmlns=""` bit); however this code looks way more terse.

Thank you!

brendanhay:
Yeah, I actually tried using `RenderSettings.rsNamespaces` to add a default namespace when rendering, and assigning the prefix for that namespace whenever creating an element which didn't have one explicitly set. Didn't seem to work as expected though.

brendanhay:
To elaborate on the 'didn't work as expected' point above, for my future self - since it prefixes everything (and retains the prefixes) you end up with some default `aws` or whatever the service name prefix is on _every_ element. a) slight increase in file size for larger documents b) unable to copy/paste the AWS examples into the tests/fixtures without significant modification.

Going to merge this and revisit if it's not sufficient.

PR #185, opened 2015-08-20
Demonstration of incorrect/unintentional exception semantics

brendanhay:
To run, from the project root:

```
$ stack ghci amazonka-examples
```

And then, in GHCi:

```
λ: import Example.ExceptionSemantics
λ: exceptions Frankfurt "non-existent-dynamo-table"
Listing all tables in region eu-central-1
Throwing deliberate IOError
Caught: user error (deliberate!)
Performing table scan of foo
*** Exception: ServiceError (ServiceError' {_serviceAbbrev = Abbrev "DynamoDB", _serviceStatus = Status {statusCode = 400, statusMessage = "Bad Request"}, _serviceHeaders = [("x-amzn-RequestId","..."),("x-amz-crc32","..."),("Content-Type","application/x-amz-json-1.0"),("Content-Length","112"),("Date","Thu, 20 Aug 2015 08:39:28 GMT")], _serviceCode = ErrorCode "ResourceNotFoundException", _serviceMessage = Just (ErrorMessage "Requested resource not found"), _serviceRequestId = Just (RequestId "...")})
```

The `examples/src/Example/ExceptionSemantics.hs` should perform the following:

* List all DynamoDB tables in the supplied region.
* Throw a deliberate `IOError` and catch it, continuing.
* Perform a table scan of the supplied table name, using `send` and `try` to catch `SomeException`.
* Perform a table scan of the supplied table name, using `paginate` and `try` to catch `SomeException`.
* Print the caught error, which should be a `ResourceNotFoundException`.
* Print when the `AWST` is unwrapped.
* Print when the `ResourceT` is unwrapped.

From the GHCi example above, this is not the case. The `ResourceNotFoundException is never caught and the example exits after the first `send Scan` attempt.

PR #186, opened 2015-08-20
Less Free

brendanhay:
This is a sad PR which removes the underlying free monad due to broken intuition regarding exceptions. See #185 for more information.


kim:
You should write a blog post about this

brendanhay:
Haha. The irony - less monads for free, maybe?

proger:
free monads have exceptional cost

joehillen:
+1 for a blog post about this. I'd like to understand why you can't use Free here.

brendanhay:
Nothing preventing use of `Free` other than my own errors/assumptions WRT exception semantics. I'll revisit it sometime.

brendanhay:
Issue #185 mentioned above demonstrates the motivating example.

Briefly, due to the exception semantics around how I'd constructed the `Free` DSL + interpretation(s) causing exceptions to be thrown by `runAWS*`, not at the call site for `send/paginate/await/...`.

This caused some fairly non-intuitive (and undesired) behaviour if, for example, you'd like to use the idiom `catching _Error (send ...)`, this would type-check, but then nothing would actually be caught by `catching` and the expected error would be thrown in `IO`. (Or whatever `MonadThrow` instance `runAWS*` was run in.).

PR #194, opened 2015-08-24
Bringing value back

brendanhay:
In the wise words of Justin Timberlake, there's just too much type hasochism going on in the world today.

This PR removes the `AWSSigner`, `AWSPresigner` and `AWSService` classes, as well as the various associated type/data families `Sg`, `Sv`, and `Meta` in favour of directly encoding/passing values.

As an example, instead of having to define a `AWSService` class, the default service is now passed directly in the `AWSRequest.request` definition, like `request = postJSON dynamoDB`.

This has a number of positive and negative follow-ons. It allows easier modification of the service configuration from the top-level (`amazonka`), and simplifies all the type family astronaut architecture. On the negative side, the specific signing algorithm now must support both signing + presigning, and the metadata can no longer be extracted except for logging purposes.


proger:
Out of curiosity, what impact does this have on compilation times?

brendanhay:
Limited at first glance, I'll try and gather empirical evidence to see if there is an improvement, but past tests have shown the majority of time is due to the excessive inlining (say, for`Text` fields and particuarly `case-insensitive`) and `Generic` instances.

brendanhay:
Just FYI, no noticeable speedups. :smile:

PR #196, opened 2015-08-25
Ensure trailing slashes in paths are preserved

### LINKED ISSUE #195: Trailing '/' is removed from paths
I believe the current (1.1.0 in `master`) URL encoding mechanism drops trailing `/` from the path:
https://github.com/brendanhay/amazonka/blob/develop/core/src/Network/AWS/Data/Path.hs#L50-L51:
```haskell
rawPath :: ToPath a => a -> Path 'NoEncoding
rawPath = Raw . filter (not . BS8.null) . BS8.split sep . toPath
``` 
This function converts paths like this:
```
abc/def/
```
Into this:
```
abc/def
```

As a result S3 service does not work correctly with otherwise valid object keys (e.g. keys with trailing `/`'s).

brendanhay:
Ensures erroneous paths are corrected, but trailing slashes are retained: `/foo///bar/` becomes `/foo/bar/`.

Fixes #195.

### DIFF
--- a/core/src/Network/AWS/Types.hs
+++ b/core/src/Network/AWS/Types.hs
@@ -112,7 +112,6 @@ module Network.AWS.Types
 
     -- ** Seconds
     , Seconds         (..)
-    , _Seconds
     , seconds
     , microseconds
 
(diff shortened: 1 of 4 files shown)


core/src/Network/AWS/Types.hs
```diff
@@ -604,17 +603,16 @@ newtype Seconds = Seconds Int
         , ToText
         )
 
-_Seconds :: Iso' Seconds Int
-_Seconds = iso seconds Seconds
-
 instance ToLog Seconds where
-    build (Seconds n) = build n <> "s"
+    build s = build (seconds s) <> "s"
```
brendanhay: What the. I must have been sleep coding!

brendanhay:
In hindsight, I'm not 100% certain about the decision to correct the adjacent slashes within the path. I'll leave it for now, since it's only the trailing slash behaviour that has been changed, but may revisit.

EduardSergeev:
@brendanhay Thank you for fixing it! My understanding that at least for `S3` adjacent slashes is not an error. Ideally (for us) the S3 path (object key) should not be changed at all. I suspect that for other services this slash handling may be required, but for `S3` it doesn't seem to be.

brendanhay:
Yeah, agreed. I'll take care of it separately. Thanks for raising the issue.

PR #197, opened 2015-08-25
Do not forget to set 'host' parameter

EduardSergeev:
Looks like `Request.host` is not being set so the `host == "localhost"`.


brendanhay:
Thanks, it should of been migrated to `Network.AWS.Types.clientRequest` as opposed to having to set it in either signing algorithm.

PR #201, opened 2015-08-26
Temporary fix for SQS's ReceiveMessage using incorrect enumeration values

### LINKED ISSUE #199: QueueAttributeName is used instead of missing MessageAttributeName
[QueueAttributeName](https://github.com/brendanhay/amazonka/blob/73ad00369ac3024d6ca6ac212ba33f01233a1de4/amazonka-sqs/gen/Network/AWS/SQS/Types/Sum.hs#L22-L36) is used to [represent a set of "would be" MessageAttributes](https://github.com/brendanhay/amazonka/blob/develop/amazonka-sqs/gen/Network/AWS/SQS/ReceiveMessage.hs#L100).
The correct set of `MessageAttributes` is [as per comments](https://github.com/brendanhay/amazonka/blob/develop/amazonka-sqs/gen/Network/AWS/SQS/ReceiveMessage.hs#L166-L177):
 * `All` - returns all values.
 * `ApproximateFirstReceiveTimestamp` - returns the time when the message was first received from the queue (epoch time in milliseconds).
 * `ApproximateReceiveCount` - returns the number of times a message has been received from the queue but not deleted.
 * `SenderId` - returns the AWS account number (or the IP address, if anonymous access is allowed) of the sender.
 * `SentTimestamp` - returns the time when the message was sent to the queue (epoch time in milliseconds).

Looks like Amazon documentation [is incorrect](http://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ReceiveMessage.html) since it enumerates QueueAttributes as valid values for MessageAttributes.

brendanhay:
Overrides the SQS service definition for `ReceiveMessage` by substituting in a fresh shape, in turn referencing a fresh enumeration with the correct values.

Fixes #199. In reference to boto/botocore#602.

### DIFF
--- a/amazonka-sqs/gen/Network/AWS/SQS.hs
+++ b/amazonka-sqs/gen/Network/AWS/SQS.hs
@@ -157,6 +157,9 @@ module Network.AWS.SQS
 
     -- * Types
 
+    -- ** MessageAttribute
+    , MessageAttribute (..)
+
     -- ** QueueAttributeName
     , QueueAttributeName (..)
 
--- a/amazonka-sqs/gen/Network/AWS/SQS/ReceiveMessage.hs
+++ b/amazonka-sqs/gen/Network/AWS/SQS/ReceiveMessage.hs
@@ -97,7 +97,7 @@ data ReceiveMessage = ReceiveMessage'
     { _rmVisibilityTimeout     :: !(Maybe Int)
     , _rmMessageAttributeNames :: !(Maybe [Text])
     , _rmWaitTimeSeconds       :: !(Maybe Int)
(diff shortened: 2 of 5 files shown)


EduardSergeev:
Shouldn't [Message](https://github.com/brendanhay/amazonka/blob/develop/amazonka-sqs/gen/Network/AWS/SQS/Types/Product.hs#L305-L306) use `MessageAttribute` instead of `QueueAttributeName ` as well?

brendanhay:
Sorry, I missed this. Indeed it should, I'll add the additional overrides now.

PR #210, opened 2015-08-31
Case Sensitive MemberIds

### LINKED ISSUE #209: Malformed SNS Subscribe request
I've been getting a rather cryptic `Start of list found where not expected` error when attempting to subscribe to SNS topic:

```Haskell
res <- runTokyoDebug $ AWS.send $
    SNS.subscribe topic "sqs"
        & SNS.subEndpoint .~ Just (
            SNS.endpoint & SNS.eEndpointARN .~ Just arn
        )
```

The request body is:

```
Action=Subscribe&Endpoint.EndpointArn=...&Protocol=sqs&TopicArn=...&Version=2010-03-31
```

while request body from `aws-sdk-go` looks like this:

```
Action=Subscribe&Endpoint=...&Protocol=sqs&TopicArn=...&Version=2010-03-31
```

The only difference I can spot is `Endpoint.EndpointArn=` vs `Endpoint=`.

The full error message is:

```
[ServiceError] {
  service    = SNS
  status     = 400 Bad Request
  code       = MalformedInput
  message    = Just Start of list found where not expected
}
```

brendanhay:
Since SNS has case sensitive shapes (`Endpoint` vs `endpoint`), the internal `CI` wrapper is causing one of the disparate types to clobber the other, leading to incorrect serialisation behaviour.

This PR removes the internal `CI` wrapper. The ordering (due to updated `Hashable` instances) of all exports/record declarations is impacted by this change.

Fixes #209.

### DIFF
--- a/amazonka-sns/gen/Network/AWS/SNS.hs
+++ b/amazonka-sns/gen/Network/AWS/SNS.hs
@@ -57,11 +57,11 @@ module Network.AWS.SNS
     -- ** InternalErrorException
     , _InternalErrorException
 
-    -- ** NotFoundException
-    , _NotFoundException
-
     -- ** InvalidParameterValueException
     , _InvalidParameterValueException
+
+    -- ** NotFoundException
+    , _NotFoundException
 
     -- ** TopicLimitExceededException
     , _TopicLimitExceededException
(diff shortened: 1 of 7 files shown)


brendanhay:
It's worth noting, smart constructor parameter ordering is not effected. The only externally visible change should be the slightly re-ordered exports.

PR #212, opened 2015-09-01
Streaming Signature Version 4 Calculation (Chunked Upload)

### LINKED ISSUE #70: Add option to perform streaming signature calculation for Version 4 signing
Currently the Version 4 signatures are calculated for all services including S3 using the [Authenticating Requests by Using the Authorization Header (Compute Checksum of the Entire Payload Prior to Transmission)](http://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html) algorithm.

This is less than ideal in S3's case, and an alternative [Authenticating Requests Using HTTP Authorization Header (Chunked Upload)](http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html) is offered which is considerably more efficient.

An additional signing option should be added which could be utilised for S3 signing purposes.

brendanhay:
Since Amazonka endeavours to use the V4 signing algorithm as much as possible (for consistency / simplicity), it has the undesired follow-on of requiring the payload's SHA256 to be calculated before sending the request.

This means the functions use to construct `RqBody` from `amazonka-Network.AWS.Internal.Body` need to traverse the file contents to calculate the digest and again to stream the contents to the socket for sending.

It gets worse when attempting to support encryption, as you'd need to encrypt the file contents into a temporary file to perform the SHA256 calculation of the encrypted contents.

This PR adds support for [Version 4 streaming](http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html) calculations.

Fixes #70

### DIFF
--- a/amazonka-cloudsearch-domains/gen/Network/AWS/CloudSearchDomains/UploadDocuments.hs
+++ b/amazonka-cloudsearch-domains/gen/Network/AWS/CloudSearchDomains/UploadDocuments.hs
@@ -77,7 +77,7 @@ import           Network.AWS.Response
 -- /See:/ 'uploadDocuments' smart constructor.
 data UploadDocuments = UploadDocuments'
     { _udContentType :: !ContentType
-    , _udDocuments   :: !RqBody
+    , _udDocuments   :: !Body
     } deriving (Show,Generic)
 
 -- | Creates a value of 'UploadDocuments' with the minimum fields required to make a request.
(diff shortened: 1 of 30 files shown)


brendanhay:
It appears that contrary to the AWS V4 documentation, that `Transfer-Encoding` is in fact _not_ supported .. I've switch to the same mechanism that the Java SDK uses, that is, calculating the `Content-Length` for the metadata + chunks etc.

More details:
https://forums.aws.amazon.com/message.jspa?messageID=604543
https://github.com/aws/aws-sdk-java/blob/3c42afd5a1677259ce77ee4e49ecae0657cec63a/aws-java-sdk-s3/src/main/java/com/amazonaws/auth/AwsChunkedEncodingInputStream.java#L268

PR #217, opened 2015-09-09
ISO8601 timezone formatted incorrectly for amazon

charleso:
`_serviceCode = ErrorCode "MalformedInput", _serviceMessage = Just (ErrorMessage "timestamp must follow ISO8601")`

This only seems to affect sending dates to amazon, the parsing itself works fine (otherwise I suspect this would have been hit ages ago):

```
> fromText "2014-11-07T04:42:13Z" :: Either String ISO8601
Right (Time 2014-11-07 04:42:13 UTC)
```

### DIFF
--- a/amazonka-s3/fixture/CopyObjectResponse.proto
+++ b/amazonka-s3/fixture/CopyObjectResponse.proto
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <CopyObjectResult>
-  <LastModified>2009-10-28T22:32:00</LastModified>
+  <LastModified>2009-10-28T22:32:00Z</LastModified>
   <ETag>"9b2cf535f27731c974343645a3985328"</ETag>
 </CopyObjectResult>
--- a/core/src/Network/AWS/Data/Time.hs
+++ b/core/src/Network/AWS/Data/Time.hs
@@ -98,7 +98,7 @@ class TimeFormat a where
     format :: Tagged a String
 
 instance TimeFormat RFC822    where format = Tagged "%a, %d %b %Y %H:%M:%S GMT"
-instance TimeFormat ISO8601   where format = Tagged (iso8601DateFormat (Just "%X%Q%Z"))
+instance TimeFormat ISO8601   where format = Tagged (iso8601DateFormat (Just "%X%QZ"))
 instance TimeFormat BasicTime where format = Tagged "%Y%m%d"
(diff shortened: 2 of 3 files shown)


charleso:
Looks like some of the tests have hard-coded dates without 'Z'. I'm assuming they have come from real data and that this change is going to break the return formatting too?

brendanhay:
Yeah, the tests are just copy-pasta from documentation examples and there is only a single format for de/serialisation. That said, the parser should really be able to handle either of these cases.

charleso:
Hey @brendanhay. I'm curious where you found the examples without any timezone? In any case, what would you like to see me do here? Tweak the parser to handle no time zone, in which case what would it default to? Or tweak the hard-coded values?

brendanhay:
Anything which is a fixture comes from the related documentation page, in the case of the failing S3 test it'd be:

http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectCOPY.html
https://github.com/brendanhay/amazonka/blob/03e5452ac0a861e29bd7159f35089f67cd74422f/amazonka-s3/fixture/CopyObjectResponse.proto#L3

brendanhay:
I think the quickest fix is to add an additional `TimeFormat`, and then alternatively parse that in the `ISO8601` [parser](https://github.com/ambiata/amazonka/blob/493b33c2abe854956a664668ceb858f669cf003b/core/src/Network/AWS/Data/Time.hs#L112).

Longer term, I'll investigate actually moving away from `time`, using `attoparsec` etc.

brendanhay:
Actually a new `TimeFormat` doesn't need to be added - just the alternative `<|>` case for the adjusted format string.

charleso:
@brendanhay Updated. Let me know if that looks OK. I'm actually no convinced those AWS examples are real, but I'm not willing to break downstream code so I'm 100% behind your suggestion.

brendanhay:
Yeah that is usually the case - the examples are wrong. :sparkles: The fix looks good, thanks for that.

charleso:
One test is failing, the `CopyObjectResult` is a `RFC822` date, but the example used doesn't have a timezone. Sigh.

charleso:
Can copy the line to the `RFC822` `FromText` instance, or fix/tweak the `proto` file. Did you have a preference?

brendanhay:
Lets' just tweak the `proto`.

charleso:
Thanks. Done. The build should pass this time. Let me know if you see any fallout from this.

brendanhay:
:shipit:

PR #226, opened 2015-09-17
Ensure literal/primitive response bodies are correctly handled

### LINKED ISSUE #224: S3 GetBucketPolicy returns Json
Seems to be the only S3 response return Json -- http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGETpolicy.html.

brendanhay:
Extends `amazonka-gen` with a `json` literal type, that can be added via `annex` / `config` overrides. This allows for fixes with S3's `GetObjectPolicy`, Lambda, and Glacier operations that return non-streaming response bodies that are in factor opaque JSON documents.

Fixes #224.


brendanhay:
Looks like `PutBucketPolicy` is also incorrect.

PR #229, opened 2015-09-28
Updating retryService to actually delay

nhibberd:
Previously this code incorrectly caused the very first
attempt to return Nothing (Control.Retry starts at 0).
Now it should capture the first attempt, so it returns
Just and actually retry.

Not really sure with the state of the tests, this could probably do with one. Here is a repl demonstrating the problem
```
>> :m + Control.Retry
>> let retry = limitRetries 5 <> (RetryPolicy $ \n -> if n > 0 then Just 2000000 else Nothing)
>> let action = putStrLn "action" >> return True
>> retrying retry (\_ b -> return b) action
action
True
>> let retry = limitRetries 5 <> (RetryPolicy $ \n -> if n >= 0 then Just 2000000 else Nothing)
>> retrying retry (\_ b -> return b) action
action
action
action
action
action
action
True
```

### DIFF
--- a/amazonka/src/Network/AWS/Internal/HTTP.hs
+++ b/amazonka/src/Network/AWS/Internal/HTTP.hs
@@ -160,7 +160,7 @@ retryService :: Service -> RetryPolicy
 retryService s = limitRetries _retryAttempts <> RetryPolicy delay
   where
     delay n
-        | n > 0     = Just $ truncate (grow * 1000000)
+        | n >= 0    = Just $ truncate (grow * 1000000)
         | otherwise = Nothing
       where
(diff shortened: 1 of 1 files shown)


brendanhay:
:+1: Thanks!

nhibberd:
Worth noting, this is causing a serious issue since upgrading from `0.3.*`. Previously we had a custom retry policy which meant that this didn't actually get hit, the reason that we have a custom retry policy might also need a separate PR, for now I'll probably just raise an issue until I can decide the best way to do it.

brendanhay:
This fix has been released as [1.3.2.1](http://hackage.haskell.org/package/amazonka-1.3.2.1) on Hackage, note that it is only the top-level `amazonka` library that has been released.

Issue #232, opened 2015-10-07
No instance for (resourcet-1.1.6:Control.Monad.Trans.Resource.Internal.MonadResource IO)


joehillen:
```haskell
main = do
  e <- newEnv Oregon Discover
  runAWS e $ send describeInstances
  return ()
```

gives this error:

```
    No instance for (resourcet-1.1.6:Control.Monad.Trans.Resource.Internal.MonadResource
                       IO)
      arising from a use of ‘runAWS’
    In the expression: runAWS e
    In a stmt of a 'do' block: runAWS e $ void $ send describeInstances
    In the expression:
      do { e <- newEnv Oregon Discover;
           runAWS e $ void $ send describeInstances;
           return () }
```

Not a very helpful error message. I can't figure out what is wrong.

brendanhay:
The [result of `runAWS`](http://hackage.haskell.org/package/amazonka-1.3.2.1/docs/Network-AWS.html#v:runAWS) is `MonadResource m => m a`, which needs to be further unwrapped (`m a -> IO a`) using `runResourceT`. This is exported by `Network.AWS` and should be available in scope.

For example:

```haskell
runResourceT . runAWS e $ do
    ...
```

brendanhay:
I'll close this for now. If you have more questions feel free to re-open or create a new issue, you can also use the [Gitter chat](https://gitter.im/brendanhay/amazonka).

joehillen:
My confusion came from the example at the top of that page uses `runAWS` in IO. https://hackage.haskell.org/package/amazonka-1.3.2.1/docs/Network-AWS.html#g:1

brendanhay:
Thanks, I'll fix the docs

PR #233, opened 2015-10-07
Add runResourceT to usage example in Network.AWS

joehillen:
The usage example seems to be wrong. It caused the confusion in #232


brendanhay:
:+1:

PR #245, opened 2015-10-30
Indicate in "await" responses whether requests were fulfilled.

fmap:
Presently there doesn't seem to be much indication whether a waiter met success before `_waitAttempts`; service errors encountered in terminal waits are lifted and thrown (unless that wait happened to somehow succeed in spite of it), but it is not revealed whether valid and terminal requests met `AcceptSuccess`, `AcceptFailure`, or `AcceptRetry`. This patch has that missing result provided as the return type of `await`.

Motivating example; on `develop`, with `i-deadbeef` substituted for some running instance, this exits without anything being thrown, and no indication that the waiter never succeeded:

```haskell
module Main where

import Control.Lens  (set)
import Control.Monad.Trans.AWS
import Network.AWS.EC2 
import Network.AWS.Waiter (_waitAttempts)

waiter = instanceTerminated { _waitAttempts = 1 } 
 `await` set diiInstanceIds ["i-deadbeef"] describeInstances 

main :: IO ()
main = newEnv Singapore Discover
   >>= runResourceT . flip runAWST waiter
```

### DIFF
--- a/amazonka/src/Network/AWS/Internal/HTTP.hs
+++ b/amazonka/src/Network/AWS/Internal/HTTP.hs
@@ -81,7 +81,7 @@ waiter :: ( MonadCatch m
           )
        => Wait a
        -> a
-       -> m (Maybe Error)
+       -> m (Either Error Accept)
 waiter w@Wait{..} x = do
    e@Env{..} <- view environment
    rq        <- configured x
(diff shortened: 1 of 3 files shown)


amazonka/src/Network/AWS/Internal/HTTP.hs
```diff
@@ -98,9 +98,9 @@ waiter w@Wait{..} x = do
 
     result rq = first (fromMaybe AcceptRetry . accept w rq) . join (,)
 
-    exit (AcceptSuccess, _) = return Nothing
```
fmap: I'm not sure what case this was expected to handle, but I tried to preserve its semantics in spirit.

brendanhay:
Thanks, looks good.

brendanhay:
Only thing I can think of, is perhaps having a more communicative type (or at least constructor names) exposed to the user than `AcceptRetry` etc. since it might be somewhat unintuitive that it in fact signals retry attempts are exhausted. I'll ponder it and perhaps do it separately though.

fmap:
Yeah, I can see this interface being somewhat clunky absent knowledge of the internals. New constructors and/or documentation sounds good.

PR #246, opened 2015-11-07
Conditional ToElement Serialisation for Root XML Elements

### LINKED ISSUE #241: PutObjectACL incorrectly uses `putXML` instead of `put`.
It can not provide a body, if it does you will see something like:

```
[ServiceError] {
  service    = S3
  status     = 400 Bad Request
  code       = UnexpectedContent
  message    = Just This request does not support content
  request-id = Just AEC8FA44647E0C83
}
```

The fix to the generated code needs to be something like: https://github.com/ambiata/mismi/commit/7b3c906814466e59d1b0113df623b664eb6e63ac

This is verified to work as expected, but I don't have me head in the generator enough to be able to quickly make the change, I will have a look but perhaps not quickly.

brendanhay:
This provides a fix for request types such as `PutObjectACL`, which are `PUT` or `POST` requests containing an optional request body. If the field to be serialised in the body results in an empty body, previously an empty root element would be sent resulting in an `UnexpectedContent` error.

This fix ensures the request body will no longer be sent unless the field is actual set via the appropriate lens.

Fixes #241.

### DIFF
--- a/amazonka-s3/test/Test/AWS/S3/Internal.hs
+++ b/amazonka-s3/test/Test/AWS/S3/Internal.hs
@@ -13,8 +13,10 @@
 
 module Test.AWS.S3.Internal where
 
+import           Data.Time
 import           Network.AWS.Prelude
 import           Network.AWS.S3
+import           Test.AWS.Fixture
 import           Test.AWS.Prelude
 import           Test.Tasty
 import           Test.Tasty.HUnit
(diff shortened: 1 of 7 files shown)


amazonka-s3/test/Test/AWS/S3/Internal.hs
```diff
@@ -13,8 +13,11 @@
 
 module Test.AWS.S3.Internal where
 
+import           Data.Proxy
```
fmap: These look redundant.

core/src/Network/AWS/Data/XML.hs
```diff
@@ -4,6 +4,7 @@
 {-# LANGUAGE LambdaCase          #-}
 {-# LANGUAGE OverloadedStrings   #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TupleSections       #-}
```
fmap: Likewise.

amazonka-s3/test/Test/AWS/S3/Internal.hs
brendanhay: Thanks

Issue #230, opened 2015-09-28
S3 throttling causes transport errors not service errors


nhibberd:
S3 throttles differently to other services, in that (unless there is special code to handle 100 CONTINUE) it tends to accept connections then terminate them at some random point for "throttling" if required. My understanding is the reason this happens is because downloading largish things from S3 happens and they don't know that something actually needs to be throttled until things are inflight. This unfortunately results in very unclean termination, generally with `NoResponseDataReceived`, `TlsException` or `FailedConnectionException2` based on when it happened to terminate. Below is a retry handler that works for S3 retries (excluding #229 - it had worked on our patched-up 0.3.* era code), it is possibly that the `StatusCodeException` is redundant because of other things, but the other ones are essential to handling heavy S3 workloads. 

```
retryAWS :: Int -> Env -> Env
retryAWS i e = e & envRetryCheck .~ err
  where
    err c _ | c >= i = False
    err c v = case v of
      NoResponseDataReceived -> True
      StatusCodeException s _ _ -> s == status500
      FailedConnectionException _ _ -> True
      FailedConnectionException2 _ _ _ _ -> True
      TlsException _ -> True
      _ -> (e ^. envRetryCheck) c v
```

brendanhay:
Pondering this - I'll probably add a variant of the above directly to S3's `src/` (non-generated code) and then allow overriding the default service configuration retry logic to use an existing function instead of rendering one.

kim:
I would (also) suggest to add `100-continue` to `POST`/`PUT` requests by default, `http-client` handles this automatically. It appears that, if 100-continue` is not used, `http-client` has no way of knowing when to remove a connection from it's pool which was closed on the remote side. Thus, it will reuse half-open connections, stream the request body into the socket buffer and then hang indefinitely. For some reason, using `100-continue` by default also increases throughput significantly, at least within AWS.

brendanhay:
I've added a the `100-continue` header via #235. Revisting the retries - since the `Service` configuration itself only deals with `ServiceError`, not `HttpException`, I'd need to make this logic the default for all service retries. Otherwise, the `Retry` type in core will need to be updated accordingly.

domenkozar:
Are there any reasons not to retry 500 by default?

PR #247, opened 2015-11-07
Retry HttpExceptions Consisting of Transport/Connection Failures by Default

brendanhay:
.. up to 3 times by default. This can be reset or have the limit increased by:

```haskell
-- Retry 'many' times ..
e <- newEnv Rotorua Discover <&> envRetryCheck .~ retryConnectionFailure maxBound
...

-- Retry no HttpExceptions (1.3.5 and earlier behaviour):
e <- newEnv Ruatoria Discover <&> envRetryCheck .~ (\_ _ -> False)
...
```

This is a partial fix for #230. The `StatusCodeException == status500` check has not been included - since whether to include this as the default or not is (imo) subjective and specific to S3 in the mentioned issue. I may introduce a more composeable mechanism for adding `HttpException` retries in the future, such as combining with the `ServiceError` check and use of `Control.Retry`.

The hope is that this change combined with the addition of the `100-continue` header introduced in #235 provide sufficient to address the concerns in #230.

cc @nhibberd

### DIFF
--- a/amazonka/src/Network/AWS/Internal/HTTP.hs
+++ b/amazonka/src/Network/AWS/Internal/HTTP.hs
@@ -98,9 +98,9 @@ waiter w@Wait{..} x = do
 
     result rq = first (fromMaybe AcceptRetry . accept w rq) . join (,)
 
-    exit (AcceptSuccess, _) = return $ Right AcceptSuccess
-    exit (_,        Left e) = return $ Left e
-    exit (accept,        _) = return $ Right accept
+    exit (AcceptSuccess, _) = return (Right AcceptSuccess)
+    exit (_,        Left e) = return (Left e)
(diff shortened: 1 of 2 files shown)


amazonka/src/Network/AWS/Internal/HTTP.hs
```diff
@@ -98,9 +98,9 @@ waiter w@Wait{..} x = do
 
     result rq = first (fromMaybe AcceptRetry . accept w rq) . join (,)
 
-    exit (AcceptSuccess, _) = return $ Right AcceptSuccess
-    exit (_,        Left e) = return $ Left e
-    exit (accept,        _) = return $ Right accept
```
brendanhay: Shadowed variable removed.

PR #248, opened 2015-11-13
Retry 0.7

MichaelXavier:
So the retry library just released a breaking version change 0.7. It is up on hackage right now and I assume that means it will be updated in the stackage nightlies soon. In this PR I just bumped the minimum dependency to 0.7 and made the necessary changes. As you can see from the code, there would have been quite a bit of messy CPP to support older versions of retry. We'll be using our fork in the mean time. I'm not sure what the appropriate path forward is with stackage. Let me know if you need anything from me on this PR.

### DIFF
--- a/amazonka/amazonka.cabal
+++ b/amazonka/amazonka.cabal
@@ -69,7 +69,7 @@ library
         , monad-control       >= 1
         , mtl                 >= 2.1.3.1
         , resourcet           >= 1.1
-        , retry               >= 0.5
+        , retry               >= 0.7
         , text                >= 1.1
         , time                >= 1.2
         , transformers        >= 0.2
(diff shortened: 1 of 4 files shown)


brendanhay:
Thanks @MichaelXavier, can you bump the stackage resolver in `stack-7.10.2.yaml` on your branch so the Travis build can proceed? `stack-7.8.4.yaml` will probably need `retry-0.7` added to the `extra-deps` section, since the resolver is intentionally old to pick up that specific GHC version.

MichaelXavier:
@brendanhay It looks like the last nightly didn't pick it up yet so I just set both to have extra-deps for now.  We'll see if travis takes it.

brendanhay:
:+1:

MichaelXavier:
Looks like that passed! Let me know if you need anything else.

brendanhay:
Alles wunderbar, thanks!

PR #254, opened 2015-11-23
Allows use of AWS_PROFILE environment variable for file credentials

rjosephwright:
The AWS_PROFILE environment variable is standardized across Amazon's
SDKs. The value of this variable overrides the default INI section
when using file credentials.


brendanhay:
:+1:

PR #258, opened 2015-12-09
Exporting the ErrorCode data constructor

ixmatus:
So we can pattern match on the specific error code wrapped by the `ErrorCode` newtype.


brendanhay:
:+1:

ixmatus:
Thank you sir :)

Issue #259, opened 2015-12-15
is the lens dependency necessary?


gregwebs:
What is the lens dependency required for? From a quick look at the code it seems like amazonka is only using type synonyms from `lens`. It would be nice to be able to use microlens (or no lenses) to keep a lighter footprint.

brendanhay:
The `amazonka-*` libraries themselves use `prism/lens` and other optical constructors/operators re-exported by `Network.AWS.Prelude`. Detaching the use of `prism/Prism` in particular was outlined by Edward Kmett [here](https://www.reddit.com/r/haskell/comments/3kbj9r/edward_kmett_the_unreasonable_effectiveness_of/cuxphr7).

brendanhay:
It's also worth mentioning from the above thread, that pulling in `profunctors` to continue to support `Choice/Prism` means the net-win (in terms of compilation footprint) aren't as large as expected, the transitive dependencies are still the same, namely `tagged`, `distributive`, `comonad`, `semigroups`, `nats`, `contravariant`, `StateVar` etc. - you're only shaving off the actual `lens` modules.

gregwebs:
thanks for the explanation. `lens` itself does have a lot of code even without dependencies.

brendanhay:
#261 contains an audit of `lens` usage in all the libraries contained in this repo aside from `amazonka-s3-encryption`. The [Network.AWS.Lens] module now contains the canonical import/exports and illustrates exactly what from `lens` is being used.

There is some low hanging fruit there, with things like `&` and `<&>` that can easily be removed or replaced with `base` and alot of the `Lens'` + `Fold` usage having counterparts in libraries like `microlens`. Some of the more esoteric traversals could be refactored to be more straight forward and `Iso'` + `Prism'` + `Control.Exception.Lens` usage seems to be what remains out of that.

Replacing or refactoring all of the above appears to be a decent amount of work and while I'm interested in cutting down the compilation times + dependencies, it'll be a while (New Year) before I get the time to invest in this particular cleanup I suspect.

brendanhay:
Closing for now. Will probably revisit the `Prism` usage at some point which would make the removal of `lens` more viable.

PR #261, opened 2015-12-17
Hipster Lenses

brendanhay:
Audit of all the `lens` dependency usage sites.

In reference to #259.


brendanhay:
The full overview of all `lens` imports can be found [here](https://github.com/brendanhay/amazonka/blob/bae8a2ff15f19636028b652b0d6a7743fb44d511/core/src/Network/AWS/Lens.hs).

PR #264, opened 2015-12-23
Host header is a requirement for V4 request signing

ixmatus:
Took a while to track this one down but our S3 URLs were being rejected by AWS because the `host` header was not being signed! I took a hint from the `base` function in constructing the `rqHeaders` request object.

No included tests, I didn't see any tests for this specific functionality, but I did hand test it on our own production software and it fixes the regression.


brendanhay:
Apologies for the late reply - can you add the required lens `(%~)` re-export to `Network.AWS.Prelude` on your branch so CI can build?

ixmatus:
@brendanhay no worries! I've pushed the added export. If that's a change in another branch you'd prefer me to integrate, I can do that but I didn't see any newer updates to the develop branch.

brendanhay:
Thanks!

PR #266, opened 2015-12-27
APIGateway fixes

proger:
Turns out, APIGateway has two protocol modes, `Accept: application/hal+json` and `Accept: application/json` where the latter is the one that's actually specified in the botocore spec.

`make -C release MODELS=model/apigateway` generates a thorough diff (updated `ToHeaders` for every request), which is not included.

The only reason I added a whole protocol to `gen` is to set the header because `requestInsts` matches on the protocol. It probably makes sense to remove it to make the patch more lightweight, however earlier versions of the patch included a ton of changes for the timestamp format and for pulling the root resource id by parsing stuff in `_links` before I realised I should try see what botocore does. (｡･ω..･)っ So let me know if you'd like the change to be done differently.

### DIFF
--- a/examples/src/Example/APIGateway.hs
+++ b/examples/src/Example/APIGateway.hs
@@ -1,0 +1,34 @@
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PartialTypeSignatures #-}
+
+{-# OPTIONS_GHC -fno-warn-partial-type-signatures #-}
+{-# OPTIONS_GHC -fno-warn-name-shadowing #-}
+{-# OPTIONS_GHC -fno-warn-unused-imports #-}
+
+module Example.APIGateway where
+
+import           Control.Lens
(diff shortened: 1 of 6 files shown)


examples/src/Example/APIGateway.hs
```diff
@@ -0,0 +1,34 @@
...
+import           Control.Monad.IO.Class
+import           Control.Monad.Trans.AWS
+import           Data.Monoid
+import           Network.AWS.Data
+import           Network.AWS.APIGateway
+import           System.IO
+
+main :: IO ()
+main = do
+  lgr <- newLogger Trace stdout
```
brendanhay: Hate to bikeshed :trollface:  but maybe 4 spaces here, if only because the rest of the project is the same.

brendanhay:
:+1:

examples/src/Example/APIGateway.hs
proger: oops, indeed :)

proger:
Let's merge?

brendanhay:
Kone4no!

PR #274, opened 2016-03-19
AttachmentStatus missing "available" (#273)

misandrist:
I mean this is clearly not in the API docs either, so there's that, too.

The latest API version I can find is [API Version 2015-10-01][APIRef],
and it clearly fails to mention "available" as a state. On the
management console, there are two states, too.

1. State: attached
2. Attachment state: available

I'm just going to guess that this means it's attached to the VPC, and
it's available for routing traffic, but that's definitely a guess.

I've also just tested that this change fixes this error.

Thanks!

[APIRef]: http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_InternetGatewayAttachment.html "Latest API for InternetGatewayAttachment"

### DIFF
--- a/amazonka-ec2/gen/Network/AWS/EC2/Types/Sum.hs
+++ b/amazonka-ec2/gen/Network/AWS/EC2/Types/Sum.hs
@@ -181,6 +181,7 @@ data AttachmentStatus
     | Busy
     | Detached
     | Detaching
+    | IsAvailable
     deriving (Eq,Ord,Read,Show,Enum,Bounded,Data,Typeable,Generic)
 
 instance FromText AttachmentStatus where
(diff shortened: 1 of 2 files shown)


fmap:
You probably want to override the service definition here: https://github.com/brendanhay/amazonka/blob/develop/gen/annex/ec2.json#L7. The module you touched is generated from it.

misandrist:
I just updated the json file and force-pushed to my branch.

brendanhay:
Thank you for raising the issue and providing an example!

The fix is slightly more complicated, since everything under the `amazonka-*/gen` directories is auto-generated from the service definitions under `gen/model`. The fix involves overriding the service definition through the use of `gen/config/ec2.json` (custom amazonka generator configuration) and `gen/annex/ec2.json` (verbatim merged json service definition) to ensure that the naming is both unique and stable. See #275 for the configuration details.

brendanhay:
Closing in favour of #275, which has been merged.

PR #275, opened 2016-03-20
Override EC2 AttachmentStatus to support 'available'

brendanhay:
Adds additional annex+config overrides to support `AttachmentStatus.available`. 

Since the constructor name `Available` introduced by this change would be non-unique, the prefixing steps in the generator cause cascade of renaming to occur unless  the `enumPrefix` is configured. Both `VolumeAttachmentState` and `AttachmentStatus` now have `enumPrefix`es set to prevent this.

See #274.


misandrist:
Brilliant, thank you!

PR #301, opened 2016-07-05
Add upper bounds to http-conduit

phadej:
Also I advice to add upper bounds for all dependencies. To prevent large breakages in the future.

screen shot 2016-07-05 at 12 33 51

With https://github.com/hvr/hackage-cli is handy (scrioptably) make revisions to package when you only need to bump bounds


brendanhay:
Thanks

PR #308, opened 2016-08-17
Discover Region from environment variables

brendanhay:
This follows the official SDK(s) behaviour by retrieving the current region from either the environment variable `AWS_REGION` or by retrieving the [Instance Identity Document](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html).

* **Breaking change**: Region is now longer a parameter to `newEnv`. The region will be discovered in the fashion described above or will default to `us-east-1` per other SDK behaviour. The region can be explicitly specified via the `envRegion` lens or the `within` modifier. The examples and documentation have been updated to use the latter.
* **New function**: Instance Identity Documents are now directly retrievable in a deserialised form via `Network.AWS.EC2.Metadata.identity`.

### DIFF
--- a/amazonka/src/Network/AWS/Auth.hs
+++ b/amazonka/src/Network/AWS/Auth.hs
@@ -331,7 +331,7 @@ getAuth m = \case
 -- cannot be read, but not if the session token is absent.
 --
 -- /See:/ 'envAccessKey', 'envSecretKey', 'envSessionToken'
-fromEnv :: (MonadIO m, MonadThrow m) => m (Auth, Maybe Region)
+fromEnv :: (Applicative m, MonadIO m, MonadThrow m) => m (Auth, Maybe Region)
 fromEnv =
     fromEnvKeys
         envAccessKey
@@ -344,7 +344,7 @@ fromEnv =
 --
 -- Throws 'MissingEnvError' if either of the specified key environment variables
 -- cannot be read, but not if the session token is absent.
-fromEnvKeys :: (MonadIO m, MonadThrow m)
+fromEnvKeys :: (Applicative m, MonadIO m, MonadThrow m)
             => Text       -- ^ Access key environment variable.
             -> Text       -- ^ Secret key environment variable.
             -> Maybe Text -- ^ Session token environment variable.
--- a/amazonka/src/Network/AWS/Auth.hs
+++ b/amazonka/src/Network/AWS/Auth.hs
@@ -65,6 +65,7 @@ import           Control.Concurrent
 import           Control.Monad
 import           Control.Monad.Catch
 import           Control.Monad.IO.Class
+import           Control.Monad.Trans.Maybe  (MaybeT (..))
 import qualified Data.ByteString.Char8      as BS8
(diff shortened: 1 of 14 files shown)


amazonka/src/Network/AWS/Auth.hs
```diff
@@ -98,10 +99,14 @@ envSecretKey = "AWS_SECRET_ACCESS_KEY"
...
 envSessionToken = "AWS_SESSION_TOKEN"
 
--- | Credentials profile environment variable.
+-- | Default credentials profile environment variable.
 envProfile :: Text -- ^ AWS_PROFILE
 envProfile = "AWS_PROFILE"
 
+-- | Default region environment variable
+envRegion :: Text -- ^ AWS_REGION
+envRegion = "AWS_REGION"
```
charleso: @brendanhay I'm curious where this is documented, our understanding is that `AWS_DEFAULT_REGION` is the official environment variable.

http://docs.aws.amazon.com/cli/latest/topic/config-vars.html#general-options

brendanhay: The documentation you linked is for the Python CLI.

All of the offical SDKs use `AWS_REGION` as follows:

* https://github.com/aws/aws-sdk-js/blob/5b3a8a69f876f8b4f52e422e0a076cb2e3855c48/lib/aws.js#L56
* https://github.com/aws/aws-sdk-java/blob/ebd165684ae3857c71796512824eb850d22d8830/aws-java-sdk-core/src/main/java/com/amazonaws/SDKGlobalConfiguration.java#L191-L194

And the Go SDK has a comment about `AWS_REGION` vs `AWS_DEFAULT_REGION`:

* https://github.com/aws/aws-sdk-go/blob/f34b74c96bfd27df35643adeb14d8431ca047df5/aws/session/env_config.go#L32-L41

> AWS_DEFAULT_REGION is only read if AWS_SDK_LOAD_CONFIG is also set,  and AWS_REGION is not also set.

brendanhay: Just to add - this is an unfortunate misconception since really the only documentation for environment variables links to the CLI documentation. Most of the actual SDK documentation is buried in Java/C#/etc. class docs.

amazonka/src/Network/AWS/Auth.hs
```diff
@@ -175,11 +180,12 @@ data Credentials
     | Discover
       -- ^ Attempt credentials discovery via the following steps:
       --
-      -- * Read the 'envAccessKey' and 'envSecretKey' from the environment if they are set.
+      -- * Read the 'envAccessKey', 'envSecretKey', and 'envRegion' from the environment if they are set.
```
blender: No chance that AWS_REGION be read from `~.aws/config` as part of the discovery as well? Or is that just a CLI thing?

brendanhay: It's something I'm working towards - it is a CLI thing but the other SDKs all support it via the `AWS_SDK_LOAD_CONFIG` flag.

PR #309, opened 2016-08-17
S3 Encryption Release

brendanhay:


endgame:
Is there anything here that's worth keeping, or is it closeable given the recent regeneration of `amazonka-s3-encryption` on `develop`?

brendanhay:
No, I think this can be closed as it seems to mostly be stylistic changes. (`amazonka-s3-encryption` is hand-written, FWIW)

PR #330, opened 2016-11-24
Fix aws-lambda auth errors

ondrap:
When there is a `Expect: 100-continue` header sent to aws lambda servers, the signature somehow contains only `Expect:` and it thus fails to validate. The solution seems to be removing the `Expect:` header from the list of signed headers.

### DIFF
--- a/core/src/Network/AWS/Sign/V4/Base.hs
+++ b/core/src/Network/AWS/Sign/V4/Base.hs
@@ -267,3 +267,4 @@ normaliseHeaders = Tag
     . nubBy  ((==)    `on` fst)
     . sortBy (compare `on` fst)
     . filter ((/= "authorization") . fst)
+    . filter ((/= "expect") . fst)


brendanhay:
Thanks for your PR @ondrap.

I need a little more context around the comment "the signature somehow contains only Expect: ..", what do you mean specifically by this?

Currently when a `POST` or `PUT` is issued an `Expect: 100-continue` header is added. These headers are then sorted into a canonical form per the [signing algorithm](http://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html#d0e12707). If the `Expect` header is sent but is not part of the canonical headers then the request signature should be invalid - hence I need to understand what is actually going on and why your fix actually works!

ondrap:
See this page: http://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html

> For the purpose of calculating an authorization signature, only the host and any x-amz-* headers are required; however, in order to prevent data tampering, you should consider including all the headers in the signature calculation.

The problem is that when calling the Lambda, the server that does the authentication somehow receives the request with empty `Expect:` header and the authentication fails. The requests for the Lambda are limited to 30MB, so in the Amazon they probably strip the `100-Continue` on some of their proxy servers first. Not signing the `Expect` header doesn't seem to me a security problem.

ondrap:
However, looking at it - as the `100-continue` does not seem to be in use anyway, a further enhancement may be not sending it anyway as it for the lambda `Invoke` method as it only adds to the latency.

brendanhay:
It currently is required for S3 for any non-trivial file upload to behave predictably.

I'm currently considering how to add the header only for those requests, and not as a blanket case for all services.

PR #371, opened 2017-03-09
ReplicationStatus parser: should be '"COMPLETED" instead of "COMPLETE"

albertov:
Hi,

As mentioned here http://docs.aws.amazon.com/AmazonS3/latest/dev/crr-status.html.
Without this I get an exception when getting an object from a replicated bucket.

Cheers,
Alberto

### DIFF
--- a/amazonka-s3/gen/Network/AWS/S3/Types/Sum.hs
+++ b/amazonka-s3/gen/Network/AWS/S3/Types/Sum.hs
@@ -737,16 +737,16 @@ data ReplicationStatus
 
 instance FromText ReplicationStatus where
     parser = takeLowerText >>= \case
-        "complete" -> pure Complete
-        "failed" -> pure Failed
-        "pending" -> pure Pending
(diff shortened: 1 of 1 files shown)


brendanhay:
Hi @albertov, thanks for the PR.

The code for each library under `amazonka-*/gen` is generated from the AWS service definitions, which for [S3 have the incorrect enumeration value](https://github.com/boto/botocore/blob/5cdc31ab591ead61cdf6fdddb19eaf3b8c90ec97/botocore/data/s3/2006-03-01/service-2.json#L5660-L5663).

Rather than making changes to these files I need to notify the upstream `botocore` project that their service definition is wrong, and create a `gen/annex/s3.json` or `gen/config/s3.json` entry to override the enumeration value until upstream is fixed.

I've done the latter in #372.

albertov:
Hi @brendanhay,

I should have known better after seeing that the file lives inside a directory called "gen"..
Thanks for fixing it properly!

Alberto

PR #377, opened 2017-03-31
Bumping static version to 1.5.0

shlevy:
This way packages needing recent changes from git can set their version bounds appropriately and be ready for the next release.

### DIFF
--- a/amazonka-s3-encryption/amazonka-s3-encryption.cabal
+++ b/amazonka-s3-encryption/amazonka-s3-encryption.cabal
@@ -1,5 +1,5 @@
 name:                  amazonka-s3-encryption
-version:               1.4.5
+version:               1.5.0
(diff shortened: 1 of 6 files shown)


shlevy:
@brendanhay

shlevy:
Not sure how to update the versions of generated packages.

brendanhay:
I'm not sure I understand the initial premise - I assume you'll be pinning to a specific git commit anyway? The release will be developed in an independent feature branch(es), so you'd still be better off having the upper bound set to `< 1.5`.

shlevy:
The idea is to have a bound that will work now *and* when the next release comes out. A bound of `>=1.4.5` doesn't really work now because `1.4.5` doesn't have stuff I need, I just happen to be completely in control of which "1.4.5" I'm using so it's fine.

Not a huge deal, of course. Just a slight convenience.

PR #381, opened 2017-04-10
Do not set binary mode for handle in ‘newLogger’

### LINKED ISSUE #380: The newLogger function breaks output of Unicode characters
Hey there, I must say it's a very pleasant piece of software to work with, thank you. However, recently I found myself investigating a rather strange situation where console output that prints Unicode characters (Cyrillic script in my case) produced rather strange results (i.e. printed garbage). I started investigating this and have found that the reason is in `newLogger` function, which I use like this:

```haskell
logger <- newLogger logLevel stdout
```

It sets output mode of `stdout` to binary and this is not good:

https://www.stackage.org/haddock/nightly-2017-04-10/amazonka-1.4.5/src/Network.AWS.Internal.Logger.html#newLogger

I don't understand why for logging purposes we need to set output mode to binary? Also, it looks like logging to `stdout` is not uncommon and changing its settings in this way should be at least documented. What do you think?

mrkkrp:
Fixes #380.

### DIFF
--- a/amazonka/src/Network/AWS/Internal/Logger.hs
+++ b/amazonka/src/Network/AWS/Internal/Logger.hs
@@ -43,7 +43,6 @@ import           System.IO
 -- should be used in production code.
 newLogger :: MonadIO m => LogLevel -> Handle -> m Logger
 newLogger x hd = liftIO $ do
-    hSetBinaryMode hd True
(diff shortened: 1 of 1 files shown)


brendanhay:
Thanks :+1:

Issue #382, opened 2017-04-21
Support for V2 header signer


BBGWolfgang:
In addition to the AWS V2 query signer authentication scheme, please add the V2 header signer authentication scheme.

Specification in http://docs.aws.amazon.com/AmazonS3/latest/dev/RESTAuthentication.html
Implementation example in https://github.com/boto/boto/blob/d7253d8f4dc5fa05733422196bfe8f1e09090ba6/boto/auth.py#L161.

brendanhay:
Would you like to elaborate on your use-case?

The V2 signing mechanisms were [previously supported](https://github.com/brendanhay/amazonka/blob/22386867cb728971a746930e879929c49d3996d4/core/src/Network/AWS/Sign/V2.hs), and were removed when the incremental signing support for V4 was implemented to avoid having to support V2 + V4, when V4 is recommended by Amazon as the signing mechanism to use moving forward.  I understand this doesn't work with S3-comptable layers such as Eucalyptus et al.

BBGWolfgang:
Correct on the S3 compatible layers - our use case is Ceph and currently its [authentication method](http://docs.ceph.com/docs/master/radosgw/s3/authentication/) is closely aligned with the V2 header signer.

#383 is also not for the [V2 query signature method](https://github.com/brendanhay/amazonka/blob/22386867cb728971a746930e879929c49d3996d4/core/src/Network/AWS/Sign/V2.hs), this is for the header signing method as outlined above.

PR #383, opened 2017-04-21
AWS signer for V2 Header authentication

BBGWolfgang:
#382 - Implementation of V2 header signer given specifications outlined in issue and closely aligned with boto.
Note: 
- Expiry header is not honored.
- Temporary security token is not honored.


brendanhay:
Thanks, looks good. I'm currently traveling and will merge it when I make it back to civilisation.

BBGWolfgang:
Sounds good

PR #386, opened 2017-06-04
Update reference to fast-logger

koterpillar:
http://hackage.haskell.org/package/fast-logger is the correct package name for 
`System.Log.FastLogger`, not `FastLogger`.


brendanhay:
👍

PR #388, opened 2017-06-10
Add support for the ECS credential discovery method for tasks

### LINKED ISSUE #385: Support for ECS credentials discovery method
ECS allows providing roles for tasks, as documented in [IAM Roles for Tasks](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html). This seems to expect that a request be made to `169.254.170.2` if the `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` environment variable is set. It would be nice if amazonka’s credentials discovery method attempted this.

lexi-lambda:
This fixes #385 by adding support for credentials discovery by communicating with the ECS container agent, which allows using the credentials provided by [IAM roles associated with particular tasks](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html). It reuses the existing machinery for refreshing credentials in the background from `fromProfileName`.

I’ve done my best to figure out the existing code style and cover the relevant bases, but I wouldn’t be surprised if there are things I’ve overlooked, so obviously feel free to make tweaks to what I’ve written yourself or suggest changes.

### DIFF
--- a/amazonka/src/Network/AWS/Auth.hs
+++ b/amazonka/src/Network/AWS/Auth.hs
@@ -51,6 +51,7 @@ module Network.AWS.Auth
     , fromFilePath
     , fromProfile
     , fromProfileName
+    , fromContainer
 
     -- ** Keys
     , AccessKey    (..)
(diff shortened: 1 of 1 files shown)


brendanhay:
Looks good, thanks for the PR!

Issue #421, opened 2017-11-13
Codename O'Farrell


brendanhay:
@charleso Most of the discussed build improvements/simplifications are in #420. The next steps in no particular order are:

* Currently `amazonka-gen` requires `config/<boto_service_name>.json` file to exist. This should be turned into a warning so the generator will run through to completion even if there is a new/unseen service definition present.
* Create a 2-stage `generate` + `build` CI job where the generator runs in the first stage, and then the generated code is built in the second dependent stage. I'm unsure if this should be separate from the usual push + build CI workflow, or if in fact generated code should _only_ be committed back by a successful CI/generator run? It's possible the latter option while purer in principle, would lie upon the path to madness.
* Run the 2-stage build as a Travis Cron job (say, nightly). This should cover most `botocore` updates and give people a `nightly` branch they can use to get the latest version.
* The [documentation builds](http://brendanhay.nz/amazonka-doc/develop) have been disabled. `stack haddock` for the entire project is both consuming too much memory and taking too long to run - and then being killed by CI timeout. The most recent approach attempted to build documentation per `amazonka-*` library and then force push the updates to the GitHub pages repository. This was complicating the build scripts and just a bit shit in the general sense. I'd like to find another, possibly out of band solution to have up to date `develop` documentation (or the `nightly` branch) for people to refer to.
* Add an additional [override](https://github.com/brendanhay/amazonka/blob/afeb929c758c772fe83692a54978cf7bb05a9de5/gen/src/Gen/Types/Config.hs#L83) to `config/<boto_service_name>.json` which allows an `Other/Unknown` field to be emitted for an enum. The intent being only _some_ enum types could be extended deliberately with a field that will always succesfully encode/decode, such as `InstanceType = Unknown !Text | ...`. 
* Generating a nice `CHANGELOG.md` as part of the release process. It is currently (mostly) hand-rolled to try and decipher some of the upstream changes and keep it mostly to the point. A better policy regarding PR naming and my own personal branch/commit/squash messaging strategy would also go a long way.

Looking at Travis' staged build + cron documentation this should be possible - but a move to Gitlab CI could also be an option, as it has better support for pipelines + cron/triggered builds.

Additional steps that need to be documented:

* Running the generator (which is literally `make -C gen`).
* Releasing the project (`git flow release start ...; ./script/set-version 1.5.0 1.5.0; make -C gen`).

If you can help out with any of these - please feel free to open an issue or PR and go for it.

charleso:
[...]

PR #422, opened 2017-11-15
Remove linux-isms in the set-version script.

tmcgilchrist:
https://github.com/brendanhay/amazonka/issues/421

### DIFF
--- a/script/set-version
+++ b/script/set-version
@@ -1,14 +1,14 @@
-#!/usr/bin/env bash
+#!/bin/sh -eux
 
 set -euo pipefail
 
-pvp="$1"
-semantic="$2"
-
-if [ -z "$pvp" ] || [ -z "$semantic" ]; then
+if [ $# -ne 2 ]; then
     echo "Usage: bump-version <VERSION> <DEPENDENCY>"
     exit 1
 fi
+
+pvp="$1"
+semantic="$2"
 
 specs=(
  amazonka/amazonka.cabal
@@ -22,9 +22,10 @@ specs=(
 echo "Bumping version to $pvp and dependencies to == $semantic ..."
 
 for cabal in ${specs[@]}; do
-    sed -i"" "s/^\(version: *\).*$/\1$pvp/" $cabal
-    sed -i"" "s/^\(.*amazonka-[a-z0-9]* *== *\)\([0-9\.]\+\*\)$/\1$semantic.*/" $cabal
-    sed -i"" "s/^\(.*amazonka *== *\)\([0-9\.]\+\*\)$/\1$semantic.*/" $cabal
+    sed -e "s/^\(version: *\).*$/\1$pvp/;"  \
+        -e "s/^\(.*amazonka-[a-z0-9]* *== *\)\([0-9\.]\+\*\)$/\1$semantic.*/" \
+        -e "s/^\(.*amazonka *== *\)\([0-9\.]\+\*\)$/\1$semantic.*/" $cabal > $cabal.new
+    mv $cabal.new $cabal
 done
 
 echo "Done."


tmcgilchrist:
This change is fixing the linux-isms so the script can be run on OSX with the BSD version of tools.

script/set-version
```diff
@@ -1,30 +1,30 @@
...
-semantic="$2"
-
-if [ -z "$pvp" ] || [ -z "$semantic" ]; then
+if [ $# -ne 2 ]; then
     echo "Usage: bump-version <VERSION> <DEPENDENCY>"
     exit 1
 fi
 
+pvp="$1"
+semantic="$2"
```
tmcgilchrist: The check around arguments wasn't working in my shell, I think checking the length of arguments and printing usage before failing is the more robust solution.

script/set-version
```diff
@@ -1,30 +1,30 @@
...
+
 specs=(
  amazonka/amazonka.cabal
  core/amazonka-core.cabal
  examples/amazonka-examples.cabal
  gen/amazonka-gen.cabal
  test/amazonka-test.cabal
- amazonka-s3-encryption/amazonka-s3-encryption.cabal
```
tmcgilchrist: I had a query about this set of cabal files and why these particular ones are included.
My intuition is that it should include all cabal files. @brendanhay

script/set-version
```diff
@@ -1,30 +1,30 @@
...
 echo "Bumping version to $pvp and dependencies to == $semantic ..."
 
 for cabal in ${specs[@]}; do
-    sed -i"" "s/^\(version: *\).*$/\1$pvp/" $cabal
-    sed -i"" "s/^\(.*amazonka-[a-z0-9]* *== *\)\([0-9\.]\+\*\)$/\1$semantic.*/" $cabal
-    sed -i"" "s/^\(.*amazonka *== *\)\([0-9\.]\+\*\)$/\1$semantic.*/" $cabal
+    sed -e "s/^\(version: *\).*$/\1$pvp/;"  \
+        -e "s/^\(.*amazonka-[a-z0-9]* *== *\)\([0-9\.]\+\*\)$/\1$semantic.*/" \
+        -e "s/^\(.*amazonka *== *\)\([0-9\.]\+\*\)$/\1$semantic.*/" $cabal > $cabal.new
+    mv $cabal.new $cabal
```
tmcgilchrist: Echoing out to a new file then using mv gets around the incompatibilities with `sed` and inlace edits.

script/set-version
brendanhay: The `set-version` script deals with the hand-rolled `cabal` files, not the generated ones. The flow is to run `set-version` to bump the listed `cabal` file versions, and then run the generator which will discover those versions via:

- [cabal.ede](https://github.com/brendanhay/amazonka/blob/afeb929c758c772fe83692a54978cf7bb05a9de5/gen/template/cabal.ede#L58) - Template for generated `amazonka-*/amazonka-*.cabal` files.
- [Makefile](https://github.com/brendanhay/amazonka/blob/afeb929c758c772fe83692a54978cf7bb05a9de5/gen/Makefile#L9-L15) - which passes the versions from `amazonka` and `amazonka-core` into the generator.

Part of the reason for this (and why `set-version` takes two version numbers) is to allow dependency versions to differ/go out of sync. So far I've only used this a couple of times when pushing out a hotfix to either `amazonka-core` or `amazonka-ec2`, separately.

script/set-version
brendanhay: :+1:

script/set-version
brendanhay: Any reason to remove `amazonka-s3-encryption`? It similarly needs to be version bumped and is somewhat confusing, not auto-generated.

tmcgilchrist: No reason, I'll add it back in.

tmcgilchrist: 👍 thanks for the explanation.

PR #428, opened 2017-11-28
Fixing two types in the instance identity document

dudebout:
The fields devpayProductCodes and billingProducts in the instance
identity document can be null, or an array of strings. For reference,
see this test in the AWS Java SDK:
https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/test/java/com/amazonaws/util/EC2MetadataUtils_TT0049160280Test.java


brendanhay:
Thanks!

PR #429, opened 2017-12-05
Remove duplicate case-sensitive files.

tmcgilchrist:
This was causing issues on OSX's case-insensitive file system.


nhibberd:
Would appreciate this 👍

tmcgilchrist:
Thanks

PR #437, opened 2018-01-09
Get DescribeStacks working

dfithian:
I know this isn't what I'm supposed to do, but I'm filing for visibility. Happy to try to edit the configuration JSON but this is my first experience with this project so I'm unsure if it's as straightforward as the [other fix](https://github.com/brendanhay/amazonka/pull/398).


dfithian:
@brendanhay Have you seen this?

brendanhay:
Apologies @dfithian - I have, but the changes have been made to the generated code and will be overwritten - we'll need to add an override to correct the service definition's type metadata/annotations. Examples can be found under `gen/config/ec2.json` and `gen/annex/ec2.json` - otherwise I'll try and get it in the upcoming release myself.

brendanhay:
Fixed in #459

PR #431, opened 2017-12-12
Add CPP for backwards compatible http-client + http-conduit

nhibberd:
Have tested with

- http-client 0.4.18.1
- http-client 0.4.31.1
- http-client 0.5.7.1

on `amazonka` and `amazonka-core`

Updates based on https://github.com/brendanhay/amazonka/commit/0656b5a12001a1d32cca65cf69f6327d3fcef858

PR #441, opened 2018-01-20
Switch to non-deprecated function from http-conduit

bitc:
http-conduit 2.3.0 (Released Jan 16, 2018) removed the deprecated function `conduitManagerSettings`. 

Instead we should use `tlsManagerSettings` (which was introduced in http-conduit 2.1.7, released Jul 21, 2015).

### DIFF
--- a/amazonka/amazonka.cabal
+++ b/amazonka/amazonka.cabal
@@ -68,7 +68,7 @@ library
         , directory           >= 1.2
         , exceptions          >= 0.6
         , http-client         >= 0.4 && < 0.6
-        , http-conduit        >= 2.1.4 && < 3
+        , http-conduit        >= 2.1.7 && < 3
         , http-types          >= 0.8
         , ini                 >= 0.3.5
         , mmorph              >= 1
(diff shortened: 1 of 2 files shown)


charleso:
As a downstream user of amazonka who still has to support an old version of ghc for a little longer can I ask that we try to avoid breaking compatibility for the sake of it, using CPP where necessary:

https://github.com/brendanhay/amazonka/pull/431

I know it's a pain, for a library as important as amazonka I argue it's worth a little extra pain/effort to make sure it works for the widest range of versions. Thanks.

bitc:
CPP could be used here, but is it really necessary?
All I've done in this patch is bump the minimum required version of `http-conduit` from 2.1.4 -> 2.1.7. Note that http-conduit 2.1.7 was released way back in July 2015

phadej:
@bitc could you rebase this to current `master`, there is (hopefully simple to resolve) confict, and `master` should be green.

This fixes #446

bitc:
@phadej I've done the rebase :+1:

phadej:
@brendanhay this is green now

brendanhay:
Fixed in #454 and #459

brendanhay:
:man_facepalming:

PR #444, opened 2018-01-24
Strip 'Expect' header when presigning URLs

### LINKED ISSUE #363: Presigning of S3 PutObject Erroneously Includes Expect Header
This is due to a somewhat recent change I made to [add plugins](https://github.com/brendanhay/amazonka/blob/e0c8fea840b44b6173fc59d4715c2b29b2acee18/gen/config/s3.json#L13) directly to individual [request types](https://github.com/brendanhay/amazonka/blob/e0c8fea840b44b6173fc59d4715c2b29b2acee18/amazonka-s3/gen/Network/AWS/S3/PutObject.hs#L303).

Two possible fixes come to mind:

1. Customise the plugin application so somehow request construction can 'choose' headers and other request metadata depending on signing
2. Filter headers according to a whitelist specifically for presigning.

mossprescott:
Fixes #363.

This is more or less the _simplest_ thing that can be done to make
pre-signed PUT urls work, and it exposes the header transformation
at the lowest level, so that it's possible to build some other
behavior on top of that. However, this is neither as targeted nor
as flexible as the solutions proposed in the issue.

This is the approach from the workaround I posted on that issue, but simpler because I didn't thread the transformation of headers all the way up through the layers. I'm submitting this for feedback; happy to rework this branch to make this more robust, flexible, or whatever.

I've tested it by running the following:

```haskell
main = do
  now <- getCurrentTime
  let expireSeconds = 300
  env <- newEnv Discover
  urlBytes <- runResourceT $ runAWS env $ presignURL now expireSeconds $
    putObject testBucket testKey ""
  putStrLn $ unpack $ decodeUtf8 urlBytes
```

```bash
curl -X PUT "https://s3.amazonaws.com/..." -d "foo"
```

### DIFF
--- a/amazonka/amazonka.cabal
+++ b/amazonka/amazonka.cabal
@@ -69,6 +69,7 @@ library
         , exceptions          >= 0.6
         , http-client         >= 0.4 && < 0.6
         , http-conduit        >= 2.1.4 && < 3
+        , http-types          >= 0.8
         , ini                 >= 0.3.5
         , mmorph              >= 1
         , monad-control       >= 1
--- a/amazonka/src/Network/AWS/Presign.hs
+++ b/amazonka/src/Network/AWS/Presign.hs
@@ -18,10 +18,12 @@ module Network.AWS.Presign where
 import           Control.Monad
 import           Control.Monad.IO.Class
 import           Network.AWS.Data.Time
-import           Network.AWS.Lens       ((%~), (&))
+import           Network.AWS.Lens         ((%~), (&))
 import           Network.AWS.Prelude
-import           Network.AWS.Request    (requestURL)
+import           Network.AWS.Request      (requestURL)
 import           Network.AWS.Types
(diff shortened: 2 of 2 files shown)


MaxGabriel:
I reached out to @mossprescott after I saw his Github comment, since we need client S3 uploads at work. I can confirm that both his code from #363 and this PR fixes the issue 👍

mossprescott:
The GHC 7.10 build is broken here, but also on master/develop, so I'm not sure what if anything I can do about that.

brendanhay:
Thanks for taking the time to make a PR, it looks like a good balance of backwards compatibility, flexibility, and simplicity! :shipit:

MaxGabriel:
@brendanhay Any chance you could do a release with this fix? I understand if not—I imagine it's a huge pain to do releases when this many packages are released in lock step, and I can keep using Github to pull in Amazonka.

PR #456, opened 2018-04-09
Add Semigroup instances for GHC8.4 compatibility

### LINKED ISSUE #451: Compile error on GHC 8.4.1
Since `Semigroup` is a superclass of `Monoid` in GHC 8.4.1, the build fails with:

```src/Network/AWS/Data/Query.hs:44:10: error:
    • No instance for (Semigroup QueryString)
        arising from the superclasses of an instance declaration
    • In the instance declaration for ‘Monoid QueryString’
   |
44 | instance Monoid QueryString where
   |          ^^^^^^^^^^^^^^^^^^
```

We should have a `Semigroup` instance for it. However adding it will drop GHC 7 compatibility since it's only introduced with base 4.9, however we can either use CPP to conditionally derive it (ugly), or import `semigroups` package on older GHC's.

utdemir:
This fixes #451 . I checked that it compiles on GHC `8.0.2`, `8.2.2` and `8.4.1`.


nhibberd:
What about with 7.10.3

utdemir:
@nhibberd I don't have a working GHC installation so I don't know. I don't see a reason why it won't work tho.

utdemir:
@nhibberd Travis tests have passed, and I saw that you're testing on GHC 7.10.3 too, so I guess it's fine.

utdemir:
Thank you @brendanhay , rebased.

PR #461, opened 2018-05-19
Add a `MonadUnliftIO` instance for `AWST'`

fumieval:
This is needed to make it work with a more recent version of conduit.


brendanhay:
Thanks 👍

domenkozar:
@brendanhay :pray:

PR #467, opened 2018-05-30
Restored old behavior of sinkBody, see #466

### LINKED ISSUE #466: S3 Example broken
This line now breaks with a `ConnectionClosed`. It worked before with lts-9.X, doesn't anymore with nightly stackage.

From what I know, it's an issue with the connection being closed too fast in this part of the code. Instead, the `MonadResource` instance from outside should be used.
https://github.com/brendanhay/amazonka/blob/develop/examples/src/Example/S3.hs#L76

reactormonk:
Fixes #466, but I haven't actually checked if it works yet.

### DIFF
--- a/amazonka/src/Network/AWS/Internal/Body.hs
+++ b/amazonka/src/Network/AWS/Internal/Body.hs
@@ -26,8 +26,8 @@ getFileSize :: MonadIO m => FilePath -> m Integer
 getFileSize path = liftIO (withBinaryFile path ReadMode hFileSize)
 
 -- | Connect a 'Sink' to a response stream.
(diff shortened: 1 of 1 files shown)


utdemir:
I just checked it using this branch, still fails with the same error:

```test :: IO ()
test = do
  env <- newEnv Discover
  runResourceT . runAWS env $ do
    liftIO $ print "sending"
    gors <- send $
      getObject
        (S3.BucketName "somebucket")
        (ObjectKey "index.html")
    liftIO $ print $ gors ^. gorsResponseStatus
    chunks <- sinkBody (gors ^. gorsBody) sinkList
    liftIO $ print chunks```

PR #469, opened 2018-06-10
Mark appropriate fields as required on multipart glacier uploads.

LeifW:
These are apparently not marked as required in the service file, but
they are marked as required in the docs for these things, e.g.
https://docs.aws.amazon.com/amazonglacier/latest/dev/api-multipart-initiate-upload.html

I don't know the difference between marking a field as required in the
shapes of gen/annex/foo.json, vs marking it as required in the
typeOverrides of gen/config/foo.json. This current approach seems to
give the desired effect, though.

There's likely plenty of other request types where the service files are
not entirely accurate, but I just did this for the types I'm currently
working with (the ones necessary to do multipart uploads).

### DIFF
--- a/gen/annex/glacier.json
+++ b/gen/annex/glacier.json
@@ -1,5 +1,32 @@
 {
     "metadata": {
         "serviceAbbreviation": "Glacier"
+    },
+    "shapes": {
+      "InitiateMultipartUploadInput": {
+        "required": [
+          "accountId",
+          "vaultName",
+          "partSize"
+        ]
+      },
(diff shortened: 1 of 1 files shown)


brendanhay:
FWIW `typeOverrides` predates the `annex/*` files and yields somewhat more predictable results given the adhoc JSON object merge strategy employed by `annex`.

brendanhay:
(They also are more robust in the presence of service description updates since you don't have to continually keep the annex files in step.)

PR #470, opened 2018-06-10
Support building amazonka-gen on GHC 8.4, and the new unexceptionalio lib.

LeifW:
The mappend = (<>) lines are redundant on GHC 8.4.
The migration pages suggests to conditionally add them with CPP, but the
other PR upgrading things to GHC 8.4 didn't do that either.
https://prime.haskell.org/wiki/Libraries/Proposals/SemigroupMonoid#Writingcompatiblecode

### DIFF
--- a/gen/amazonka-gen.cabal
+++ b/gen/amazonka-gen.cabal
@@ -99,6 +99,6 @@ executable amazonka-gen
         , text-regex-replace
         , time                 >= 1.5
         , transformers
-        , unexceptionalio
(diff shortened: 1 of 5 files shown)


LeifW:
Apparently stackage lts is still on unexceptionalio-0.3.0. I'll revert that change for now.

domenkozar:
Seems like it fails on GHC 8.2 :(

brendanhay:
Thanks - now (3 years later) fixed in `develop`.

PR #471, opened 2018-06-11
Add PrimMonad instance.

LeifW:
This way operations of e.g. Vector can be used in AWST.

Not really sure if this instance is the thing to do, but since AWST' is
just a newtype wrapper for ReaderT, I copied the ReaderT instance. It
seems to be working for me.


LeifW:
Duplicate of #510, which has since been merged. Closing.

PR #480, opened 2018-07-11
Read the configuration file to find the default region to use

### LINKED ISSUE #479: Region not read from file
The region is not getting picked up from the configuration file (`~/.aws/config`) despite being set, preventing file based environments from changing regions. `Network.AWS.Auth.fromFilePath` only looks at the credentials file and always leaves the region unset.

mfine:
When reading environment material from file, the default configured region is not picked up. Add support for reading the configuration file and picking up the region to use.

Fixes #479.

/cc @brendanhay


endgame:
Is this solving the same problem as #612 ? If so, which is the better implementation?

PR #482, opened 2018-07-16
Drop some MonadThrow and MonadCatch constraints

snoyberg:
This is arguable, and simply closing is acceptable. My argument for
making this change is that, when `MonadIO` is already in the
constraints, adding in either `MonadThrow` or `MonadCatch` simply adds
ambiguity as to where exceptions may end up being thrown. Reducing to
just `MonadIO` makes it clear that all exceptions will be thrown as
runtime exceptions.

There may be other places in the library this approach applies, but this
is where I noticed it.


brendanhay:
The constraints are a bit silly, agreed. So silly in fact, I've been unable to reply out of shame for the past 3 years.

I'll pull the PR and make some modifications to rebase it against the current `develop` which has diverged a little. Thanks!

Issue #477, opened 2018-07-06
Test against AWS?


LeifW:
I was thinking it would be handy to test against AWS. One benefit is hopefully people could know in advance whether what's in the repo works or not.
e.g.currently amazonka-glacier is missing an auth header, and sends the wrong format of requests for InitiateJob requests. amazonka-sdb sends the auth in the wrong style.
I could contribute some tests if this would be desired, e.g. for sdb a simple "list domains" request, and for glacier perhaps a "list jobs" request or similar. Both would check that the auth and the request format is at least correct. I believe the first 10,000 such requests each month are free for those services.

naushadh:
An AWS mock could be used instead of relying on a live account/credential. Ex: https://github.com/localstack/localstack

endgame:
Integration tests that boot a mock stack would be cool because we could CI them; I'd be more cautious about the bill that you'd get running integration tests against a live AWS account: since you need to test everything, the keys would have nearly unrestricted permissions, which would be a big risk.

A suite of manually-invoked tests masquerading as examples would be useful when trying to fix specific services and when developing tweaks to the files that live in `config/`. I'd look at a PR that creates a new package to hold these manual tests. Making them `executable` instead of `test-suite` in the cabal file is probably best so they don't get run by CI.

I'll close this issue now, because I think that answers the OP question, and otherwise it would stay open forever because there's no concept of what "done" would be otherwise.

brendanhay:
Smoke tests that masquerade as examples would indeed be great - I would leave it up to individuals to run them, at their own expense, though. This was more or less what `./examples` was supposed to be - it just never got a whole lot of love.

PR #493, opened 2018-10-01
Close connections immediately (when we know that we can do that)

### LINKED ISSUE #490: Open connections are leaking
I do a bunch of `SQS.deleteMessage`s in a loop, and experience long waits (5-10 seconds) after getting 99 open connections. This happens with https://github.com/iain/fake_sqs (I haven't tested with real SQS).

This behavior is introduced by a recent commit (made after 1.6.0 was released): https://github.com/brendanhay/amazonka/commit/2688190f659e4d87fd6862b52b04be706b17d251.

My guess is that in the absence of finalizers in conduit-1.3, nothing is closing connections open by `Response`s. The following diff fixes the behavior of `deleteMessage` for me:

```diff
diff --git a/core/src/Network/AWS/Response.hs b/core/src/Network/AWS/Response.hs
index 4791de6a3c..907b0faa05 100644
--- a/core/src/Network/AWS/Response.hs
+++ b/core/src/Network/AWS/Response.hs
@@ -45,8 +45,9 @@ receiveNull :: (MonadResource m, MonadThrow m)
             -> Proxy a
             -> ClientResponse
             -> m (Response a)
-receiveNull rs _ = stream $ \_ _ x ->
-    liftResourceT (x `connect` pure (Right rs))
+receiveNull rs _ = stream $ \_ _ x -> do
+    sinkLBS x
+    pure (Right rs)

 receiveEmpty :: (MonadResource m, MonadThrow m)
              => (Int -> ResponseHeaders -> () -> Either String (Rs a))
```

It's still slower than it used to be (I've no idea why), but at least it doesn't leak connections anymore and I get no stuck requests.

Other `receiveBlah` functions might also be affected.

neongreen:
Fixes #490.

I'm going to test this in https://github.com/wireapp/wire-server so please don't merge that yet. @LeifW @naushadh @dsturnbull would you also like to take a look?

### DIFF
--- a/core/src/Network/AWS/Response.hs
+++ b/core/src/Network/AWS/Response.hs
@@ -12,7 +12,26 @@
 -- Stability   : provisional
 -- Portability : non-portable (GHC extensions)
 --
-module Network.AWS.Response where
+-- Functions contained in this module fully consume the body and thus close
+-- the connection. This is needed to avoid hitting this issue:
+-- <https://github.com/brendanhay/amazonka/issues/490>.
+--
+-- The only exception is 'receiveBody', which passes a streaming response
+-- body to a callback and thus is not allowed to close the connection. Users
+-- of streaming functions are advised to be careful and consume the response
+-- body manually if they want the connection to be closed promptly.
+--
+-- Note that using 'runResourceT' will always close the connection.
+--
+module Network.AWS.Response
+    ( receiveNull
+    , receiveEmpty
+    , receiveXMLWrapper
+    , receiveXML
+    , receiveJSON
+    , receiveBytes
+    , receiveBody
+    ) where
 
 import Control.Applicative          (pure)
 import Control.Monad.Catch
@@ -21,6 +40,7 @@ import Control.Monad.Trans.Resource
 
 import Data.Aeson
 import Data.Conduit
+import Data.Conduit.List (sinkNull)
 import Data.Monoid
 import Data.Proxy
 import Data.Text    (Text)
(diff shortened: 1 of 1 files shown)


neongreen:
So there's a choice of two commits now:

* one does `sinkNull`
* the other closes the connections without sinking anything

The latter is several times faster in my benchmarks (five times faster for SQS's `deleteMessage` with local fake SQS), but I'm not sure it's correct. On the other hand, I also don't see any reason why it wouldn't be correct, but it still seems suspicious.

naushadh:
I too would prefer the faster way, correctness provided. Perhaps long term we really do need some actual testing against AWS (or a mock of it) (https://github.com/brendanhay/amazonka/issues/477) and/or lift connection handling into the type system.

neongreen:
This is in production in https://github.com/wireapp/wire-server now and things haven't broken yet, so I'm in favor of merging it. cc @MichaelXavier?

dzhus:
Same here – *seems* ok to merge

mheinzel:
I rebased this change, which we have been running in production in https://github.com/wireapp/wire-server for over a year at this point.

Some CI builds failed because of `happy`, one because of ["No output has been received in the last 10m0s"](https://travis-ci.org/brendanhay/amazonka/jobs/648467830), but this seems to happen on master, too.

If there is any interest in merging this, that would be great. If not, we'll keep using our fork.
Let me know if there is anything I could do to help make this happen.

brendanhay:
LGTM

mheinzel:
Wow, thanks for the quick response!

PR #497, opened 2018-10-10
Repair time serialization tests broken by PR #496

penntaylor:
Forgot to update the tests along with PR #496.


penntaylor:
The core / time tests appear to be passing with stack `8.2.2` but failing with `8.0.2` and `7.10.3`. I don't immediately see what's causing that. Rebuilding locally with `8.0.2` and attempting to sort it out.

penntaylor:
Figured out the issue: lts-9.0 resolver version of Data.Time.Format doesn't support the width modifier. Closing this PR; will open a new one later today that simply omits fractional seconds from the time stamps. I was hesitant to do that initially, but it seems like the only reasonable way to support older versions of Data.Time.Format.

PR #498, opened 2018-10-29
Add STANDARD_IA storage class support to ObjectVersionStorageClass

newhoggy:


domenkozar:
I've updated S3 model in https://github.com/brendanhay/amazonka/pull/527

PR #502, opened 2018-11-14
Remove fractional seconds from serialization of iso8601 timestamps

penntaylor:
Finally fixing the mess I created with #496 and #497.


penntaylor:
Note the failed checks all appear to be related to timeouts while installing ghc. The `amazonka-core` tests all passed.

PR #504, opened 2018-12-10
Add Nix expressions

sevanspowell:
- Add commands for generating Nix expressions for Amazonka to the Makefile.
- Use "make nix" to generate all the Nix files and "make nix-clean" to remove
  all the Nix files. "make nix-clean" and "make nix" should be run when changing
  the cabal file for a package, and probably should be run before submitting a 
  PR.
- This PR is very useful for people using Nix who want to easily target specific
  commits. For example grabbing a bugfix while waiting for the next stable
  release.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,3 +29,20 @@ core:
 
 $(SERVICES):
 	stack build --fast $@
+
+# Nix
+define clean-nix
+	rm -f $(strip $1)/default.nix;
+endef
+
+.PHONY: nix-clean
+nix-clean:
+	$(foreach l,$(LIBRARIES),$(call clean-nix, $l))
+
+define mk-nix
+  cd $(strip $1); make nix; cd ..;
+endef
+
+.PHONY: nix
+nix:
+	$(foreach l,$(LIBRARIES),$(call mk-nix, $l))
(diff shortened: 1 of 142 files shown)


dudebout:
Out of curiosity, how do you use these files to target a specific commit? How is that different from overriding the source?

utdemir:
Just for reference, I usually do something like this:

```
let
amazonkaSrc = pkgs.fetchFromGitHub {
   owner = "brendanhay"; repo = "amazonka";
   rev = "1.6.1";
   sha256 = "0i8clxy6nkly5pwlzq9ly6kc86h0spwwvnqsp6z090pjjqn5djna";
};
amazonkaSub = folder: "${amazonkaSrc}/${folder}";  
in
{
   ...
   amazonka =
    se.callCabal2nix "amazonka"  (amazonkaSub "amazonka") {};
  amazonka-test =
    se.callCabal2nix "amazonka-test"  (amazonkaSub "test") {};
  amazonka-core =
    se.callCabal2nix "amazonka-core"  (amazonkaSub "core") {};
  amazonka-s3 =
    se.callCabal2nix "amazonka-s3"  (amazonkaSub "amazonka-s3") {};
  amazonka-sqs =
    se.callCabal2nix "amazonka-sqs"  (amazonkaSub "amazonka-sqs") {};
  ...
}
```

This is a bit less than ideal since it uses import-from-derivation functionality, but I did not have any issues yet. But I'd prefer having `.nix` files in repository so I can use them instead.

PR #505, opened 2018-12-30
Adds paginators to AWS Secrets Manager API.

andreyk0:


brendanhay:
These will need to be added to the `gen/annex` configuration - otherwise they will be overwritten by subsequent model refreshes.

PR #510, opened 2019-01-15
Add an instance for PrimMonad m => PrimMonad (AWST' r m)

axman6:
Adds an instance for PrimMonad AWST', which is used for low level memory access when IO or ST is the base monad in a transformer stack. The particular use case we came across was after the release of conduit 1.3 ~broke everything~ changed how IO is handled by using MonadUnliftIO, the conduit-extra changed its Zlib compression functions from needing

```Haskell
(MonadBaseControl b m, PrimMonad b) => ...
```

to needing 

```Haskell
PrimMonad m => ...
```

This change doesn't actually change the packages needed to compile `amazonka` as primitive was already a transitive dependency.

### DIFF
--- a/amazonka/amazonka.cabal
+++ b/amazonka/amazonka.cabal
@@ -74,6 +74,7 @@ library
         , mmorph              >= 1
         , monad-control       >= 1
         , mtl                 >= 2.1.3.1
+        , primitive           >= 0.6
         , resourcet           >= 1.1
         , retry               >= 0.7
         , text                >= 1.1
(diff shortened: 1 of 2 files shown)


axman6:
Ha, looks like this is a dupe of #471, both are nearly identical so happy with either (I did search for an issue asking about PrimMonad but didn't search PRs, oops 😬 ) @LeifW

axman6:
Thanks @brendanhay! ❤️

brendanhay:
Apologies @LeifW - I was working in descending PR order 🤷‍♂

PR #525, opened 2019-04-05
Fix cloudformation resource states

mbj:


mbj:
sorry for the noise, did not meant to PR this early.

PR #526, opened 2019-04-10
Bump the upper bound in the dependency on http-client

23Skidoo:
The changelog doesn't list any silently breaking semantic changes.

### DIFF
--- a/amazonka/amazonka.cabal
+++ b/amazonka/amazonka.cabal
@@ -67,7 +67,7 @@ library
         , conduit-extra       >= 1.1
         , directory           >= 1.2
         , exceptions          >= 0.6
-        , http-client         >= 0.4 && < 0.6
+        , http-client         >= 0.4 && < 0.7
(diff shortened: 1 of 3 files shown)


23Skidoo:
Would be also nice if you could make a Hackage revision.

domenkozar:
cc @MichaelXavier :)

23Skidoo:
Gentle ping.

MichaelXavier:
I still don't know how to do a release for this project but I can merge things that pass like this. Also @brendanhay I might suggest open upper bounds on some core packages given that releases are such a huge pain. Not a lot of sense IMO of being conservative with upper bounds all the time.

borsboom:
@MichaelXavier This change looks like it could be done with a hackage revision, shouldn't need a new release.

domenkozar:
Note: revision changes are blocking https://github.com/commercialhaskell/stackage/pull/4560

PR #527, opened 2019-04-18
Amazonka s3 model update

domenkozar:
I've updated the S3 model because I was getting runtime errors in production due to missing INTELLIGENT_TIERING StorageClass sum type.

I've split the commits in two, while README instructs not to commit generates files, they have to be eventually generated and with this PR I can also use the fork myself.

### DIFF
--- a/amazonka-s3/amazonka-s3.cabal
+++ b/amazonka-s3/amazonka-s3.cabal
@@ -64,6 +64,7 @@ library
         , Network.AWS.S3.DeleteObject
         , Network.AWS.S3.DeleteObjectTagging
         , Network.AWS.S3.DeleteObjects
+        , Network.AWS.S3.DeletePublicAccessBlock
         , Network.AWS.S3.GetBucketACL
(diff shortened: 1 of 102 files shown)


MichaelXavier:
@domenkozar looks like there's some significant compile errors on some of the branches. Are you able to look into those?

domenkozar:
Sorry for radio silence, I'll get back to this soon (maybe at ZuriHac)

Fuuzetsu:
It's been quite a while and INTELLIGENT_TIERING basically can not be used with `amazonka` at all due to this missing.

brendanhay:
There is an in progress branch with a `2.0.0.0` release being prepared that supersedes this. It also includes the changes from #597 - `INTELLIGENT_TIERING` will become a pattern synonym, which should avoid this issue in future.

Fuuzetsu:
Great. In meantime if someone wants to take S3 changes I have a branch at https://github.com/tsurucapital/amazonka/tree/update-api for now

PR #500, opened 2018-11-02
Deduplicate CreateCertificateFromCSR fixtures

m5:
When this repository is cloned on OSX, `CreateCertificateFromCSR.yaml` is overwritten by `CreateCertificateFromCsr.yaml`.

Nix uses a hash checksum to verify the repository, and with those files missing, a hash generated on OSX won't match one generated on linux.

https://github.com/NixOS/cabal2nix/issues/382

The conflicting files are all empty, and `CreateCertificateFromCSR` seems to be the preferred variant.

PR #530, opened 2019-05-21
Remove duplicate files on case-insensitive filesystem.

fiadliel:
When checking out this project on MacOS, the following warning appears:
> warning: the following paths have collided (e.g. case-sensitive paths
> on a case-insensitive filesystem) and only one from the same
> colliding group is in the working tree:
>
>   'amazonka-iot/fixture/CreateCertificateFromCSR.yaml'
>   'amazonka-iot/fixture/CreateCertificateFromCsr.yaml'
>   'amazonka-iot/fixture/CreateCertificateFromCSRResponse.proto'
>   'amazonka-iot/fixture/CreateCertificateFromCsrResponse.proto'

In addition, `nix-prefetch-git` returns a different SHA256 hash on Linux, or MacOS, because the state of the checked out directory is different (two files are missing on MacOS because of the collision).

This commit keeps the files with the capitalization from tests.


fiadliel:
Duplicate of #500

PR #535, opened 2019-07-26
Add regions: HK, Osaka, Paris, Stockholm, Ningxia, GovCloudEast.

mgajda:
Bring Amazonka up-to-date with current list of regions.
Added:
* Hong Kong (2019)
* Osaka
* Paris (2017)
* Stockholm (2018)
* Ningxia 
* GovCloudEast

This fixes: #520
(Note that alternative fix in a huge PR #522 got stalled.)

PR #541, opened 2019-08-11
Added new regions to ELB, RedShift, Route53.

mgajda:
This is update of `725340d92f8109bf5ea620a224fe613df9fba8f1` which added new regions.
It adds ELB, RedShift, and Route53 account ids for the new regions.


nhibberd:
Fixes for #535

PR #544, opened 2019-08-23
Update service definitions for cloudwatch-events

MichaelXavier:
This add some paginators and some networking features that appear to
be required for using this service with Fargate.

I tried to follow the advice of `gen/README.md` and only committed the changes to the service files. LMK if this wasn't right. We're due for a dozen or two other service updates but I wanted to make my MR minimal.

### DIFF
--- a/gen/model/events/2015-10-07/paginators-1.json
+++ b/gen/model/events/2015-10-07/paginators-1.json
@@ -1,3 +1,22 @@
 {
-  "pagination": {}
+  "pagination": {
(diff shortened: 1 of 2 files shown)


MichaelXavier:
Tested this out today and it looks like it works. @brendanhay I've got my own fork with the codegen applied. How should we go about doing the codegen for this project? Given that releases are kind of painful, we'd probably want to do an across-the-board update of service definition files.

Issue #537, opened 2019-07-28
Setting any value to PutObject's `poContentEncoding` breaks signing


mgajda:
Attempting to run this snippet:
```
{-# LANGUAGE OverloadedStrings #-}
module Main where

import Network.AWS
import Network.AWS.S3
import Network.AWS.S3.PutObject
import Control.Lens

bucketName = "test.migamake.com"

main = do
  env  <- newEnv Discover
  runResourceT $ runAWS env $
    within Frankfurt $ do
      content <- chunkedFile 8192 "public/index.html"
      send $ set poContentType     (Just "text/html")
           $ set poContentEncoding (Just "identity")
           $ putObject bucketName (ObjectKey "index.html") content
```
Results in signing error:
```
*** Exception: ServiceError (ServiceError' {_serviceAbbrev = Abbrev "S3", _serviceStatus = Status {statusCode = 403, statusMessage = "Forbidden"}, _serviceHeaders = [("x-amz-request-id","C211E7670B81BA51"),("x-amz-id-2","VdRQD1mnawJ2iXGc+CYJnPNsH8ZPtIiWcUqZUieBuOh4SBZXZNEjCSZnpQvXS7/UTElB3c68Qbw="),("Content-Type","application/xml"),("Transfer-Encoding","chunked"),("Date","Sun, 28 Jul 2019 08:50:02 GMT"),("Connection","close"),("Server","AmazonS3")], _serviceCode = ErrorCode "SignatureDoesNotMatch", _serviceMessage = Just (ErrorMessage "The request signature we calculated does not match the signature you provided. Check your key and signing method."), _serviceRequestId = Just (RequestId "C211E7670B81BA51")})
```
This breaks workaround for #536.

Note that normally there may be multiple content encodings put into a single stance, but it seems that now we have `aws-chunked` hardwired and it interferes with attempt to assign value here.

I prepared a small example of this issue:
https://github.com/mgajda/aws-empty-content-encoding-issue/tree/assign-content-encoding

mgajda:
There is workaround of not using `chunkedFile`:
```
  content <- toBody <$> Data.ByteString.Lazy.readaFile "public/index.html"
```

mgajda:
It seems that:
1. This parameter is incorrectly modeled as a String, instead of comma-separated ordered list of strings.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding
[...]

PR #547, opened 2019-09-19
remove the content-length header when doing V4 streaming signatures

dsturnbull:
Addresses https://github.com/brendanhay/amazonka/issues/546 and possibly https://github.com/brendanhay/amazonka/issues/537

Thanks to @jchia for noticing the difference with the other aws library and @Kheldar noticing something similar in boto3.

Simply removes `content-length` from the signature calculation. The header is still sent with the request.

### DIFF
--- a/core/src/Network/AWS/Sign/V4/Base.hs
+++ b/core/src/Network/AWS/Sign/V4/Base.hs
@@ -267,3 +267,4 @@ normaliseHeaders = Tag
     . nubBy  ((==)    `on` fst)
     . sortBy (compare `on` fst)
     . filter ((/= "authorization") . fst)
+    . filter ((/= "content-length") . fst)


ystael:
As I mentioned on the issue https://github.com/brendanhay/amazonka/issues/546, it appears from my reading of the documentation that the Content-Length header is required to be included in the signature: see the table under "Calculating the Seed Signature" on https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html. I have raised this issue on our AWS support ticket and I hope we will find out soon whether the implementation or the documentation will be fixed.

jchia:
@ystael The documentation says the user "must use" `content-length`, so if simply removing `content-length` works, as implied by the PR, then the documentation must be wrong and needs fixing, regardless of whether the implementation is wrong.

dsturnbull:
Since this is a sudden change I think it’s more likely a bug in s3. We theorise amazon have added some compression between their ingress and api nodes that changes the length header.

dsturnbull:
Also just to keep @ystael and @jchia in the loop, we've supplied trace logs of pre and post patch requests to AWS support

ystael:
I have received the following response from AWS support:

> On September 16, 2019, S3 introduced changes to the REST.PUT.OBJECT API to increase compliance with HTTP standards. This included a change to enforce RFC 7230 where "A sender must not send a Content-Length header field in any message that contains a Transfer-Encoding header field." In making this change, we did not also add this enforcement to our S3 documentation. We will be making changes to the S3 documentation to be consistent with the new behavior.

So it sounds like this PR needs to change to remove `Content-Length` from the request, not merely from the signature calculation.

newhoggy:
Our experience is that just removing from the signature is sufficient.

Perhaps we could merge this and have a separate PR to also remove it from the request.

domenkozar:
interesting that I'm not seeing this error, could someone try with https://github.com/brendanhay/amazonka/pull/527

dsturnbull:
@ystael what amazon has told us (we had the same message) makes no sense to me. The initial request does not have a Transfer-Encoding header set, so we should be fine to have Content-Length in it. Removing that from the signed signature is what works.

ystael:
I agree that it doesn't make sense. Support claims on our ticket that the server side logs show Transfer-Encoding is included. Are there other requests where it is?

PR #551, opened 2019-10-20
Add MonadFail instance for AWST

utdemir:
Starting from GHC 8.8, as per [MonadFail proposal](https://prime.haskell.org/wiki/Libraries/Proposals/MonadFail); `Monad` typeclass does not have a `fail` method; which instead is provided by `MonadFail` typeclass.

This PR adds `MonadFail` instance to `AWST` conditionally based on `base` version.


utdemir:
Would it be possible to cut a new release after this commit? Otherwise, I think more people will have to maintain orphan instances once they migrate to GHC 8.8.

PR #552, opened 2019-10-29
Update cloudwatch/2010-08-01

puffnfresh:
I want access a few new CloudWatch APIs. I copied in a new service-2.json but that has caused a lot of changes, e.g. it looks like CloudWatch Events has been renamed to EventBridge.

I think I made changes in the right places and ran the correct generation scripts so that the code is now consistent. Please let me know if there's a better way to do this.


brendanhay:
I usually request that people make changes to the scripts + configuration in a PR without the emitted code - just so I can reason through the configuration changes, otherwise it gets far to unwieldy. If you could remove the generated code, I'll take a look.

brendanhay:
`develop` now contains code generated from the latest version of the `2010-08-01` CloudWatch service description.

PR #562, opened 2020-03-13
Update UnliftIO instance for unliftio-core-0.2

cfraz89:
Since unliftio suddenly moved `askUnliftIO` out of `MonadUnliftIO`, this restores compatibility by implementing `withRunInIO` instead

PR #563, opened 2020-03-13
Implement instance MonadUnliftIO (AWST' r m) in terms of withRunInIO …

akrmn:
…rather than askUnliftIO, since the latter is being removed from the MonadUnliftIO class


akrmn:
just saw #562, so closing in favor of that PR

PR #564, opened 2020-03-13
Generalized MonadAWS instance for AWST beyond ResourceT IO

akrmn:
The current instance `MonadAWS AWS` is more specific than it needs to be, since `liftAWS :: AWS a -> m a` can be implemented for `AWST m` other than `AWST (ResourceT IO)`, so long as `m` satisfies `(MonadResource m, MonadCatch m)`

### DIFF
--- a/amazonka/src/Control/Monad/Trans/AWS.hs
+++ b/amazonka/src/Control/Monad/Trans/AWS.hs
@@ -40,6 +40,7 @@ module Control.Monad.Trans.AWS
     , newEnv
     , Env
     , HasEnv       (..)
+    , askEnv
 
     -- ** Credential Discovery
     , Credentials  (..)
@@ -183,7 +184,7 @@ import Network.AWS.Env
 import Network.AWS.Internal.Body
 import Network.AWS.Internal.HTTP
 import Network.AWS.Internal.Logger
-import Network.AWS.Lens            (catching, throwingM, trying, view)
+import Network.AWS.Lens            (catching, throwingM, trying, view, (^.))
 import Network.AWS.Pager           (AWSPager (..))
 import Network.AWS.Prelude         as AWS
 import Network.AWS.Request         (requestURL)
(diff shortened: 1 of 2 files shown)


akrmn:
Hi @brendanhay, may I know your thoughts on this PR?

brendanhay:
Thoughts: looks good, I can't see any potential for breaking existing code. I'm wondering out of interest how you actually plan to use this?

akrmn:
Hey! My idea is to use `AWST` anywhere on my monad stack, rather than being limited to having `AWS` at the bottom.

LeifW:
I didn't use AWS / AWST - you don't need anything AWS-specific in your monad stack besides `Env` in a reader: https://github.com/LeifW/glacier-backup/blob/master/src/Network/AWS/Glacier/LiftedGlacierRequests.hs#L26

akrmn:
Thanks for your suggestion, @LeifW. The problem is that we're actually moving away from the `HasX` approach in our codebase. Also, I think this makes sense by analogy with the `mtl` instances of the `transformers` newtypes:
```
MonadReader r (ReaderT r m)
MonadWriter w (WriterT w m)
MonadState s (StateT s m)
```
then surely
```
MonadAWS (AWST m)
```

brendanhay:
I don't see an issue with the approach in the PR since there doesn't appear to be a risk from backwards incompatibility. 

Personally I use a variant of what @LeifW suggested by having a concrete `ReaderT Env` or similar somewhere in my stack, without the constraints usually. The `MonadAWS` monad homomorphism exists to let you get from `AWS a` to whatever lets you suitably run an AWS action by supplying an `Env` and releasing any resources, say:

```haskell
type App = ReaderT Env IO

instance MonadAWS App where
    liftAWS aws = do
        env <- ask
        runResourceT (runAWS env aws)
```

.. or something along those lines, modulo some morning coffee.

akrmn:
Which matches pretty well with my implementation of `instance MonadAWS (AWST m)`, except it isn't necessary to define any instances for the monad stack, since `runAWST` would "trade" the `MonadAWS m` constraint for `MonadResource m` and `MonadCatch m` (which can be discharged later on with, e.g., `runResourceT` and `runExceptT`.)

https://github.com/brendanhay/amazonka/blob/38381bf9a0ed81850286ec4ba1ad08e9f58b5da8/amazonka/src/Network/AWS.hs#L203-L206

akrmn:
Hi! I was wondering if there's anything else you'd like to see in the PR before merging. Thanks!

PR #565, opened 2020-03-13
correct Stockholm in FromText Region

### LINKED ISSUE #560: Stockholm in FromText Region instances is wrong
https://github.com/brendanhay/amazonka/blob/29a6806fb01b22d0a58d455644cbea777ce41b43/core/src/Network/AWS/Types.hs#L720

Should this string be `eu-north-1` like in the reverse case?

https://github.com/brendanhay/amazonka/blob/29a6806fb01b22d0a58d455644cbea777ce41b43/core/src/Network/AWS/Types.hs#L748

akrmn:
This fixes #560


brendanhay:
:man_facepalming: :+1:

PR #567, opened 2020-03-17
Implement instance MonadUnliftIO (AWST' r m) in terms of withRunInIO

akrmn:
Moved implementation of instance behind CPP conditionals to check version of `unliftio-core`.


akrmn:
This PR is like #562 from @cfraz89 or #563 from me, but the CPP conditionals should get the repo to build on any version of `unliftio-core`.

brendanhay:
Thanks for the PR @akrmn - can you take a look at the conflicts and modify if applicable?

akrmn:
Hey! I've rebased now. It seems that there was a similar commit some time after I opened the PR, but I think the `INLINE` pragmas and the simpler implementation of `withRunInIO` are still worth considering. Behaviour should be the same, with a bit less wrapping/unwrapping.

akrmn:
Thanks for merging this one!

Issue #569, opened 2020-03-24
Failure parsing ObjectStorageClass from value: 'intelligent_tiering'. Accepted values: glacier, reduced_redundancy, standard, standard_ia"


k-bx:
Got it while trying to list the common crawl news bucket `s3://commoncrawl/crawl-data/CC-NEWS/2020/03/`: `Failure parsing ObjectStorageClass from value: 'intelligent_tiering'. Accepted values: glacier, reduced_redundancy, standard, standard_ia"`

I guess we need to add the intelligent_tiering somewhere.

dustin:
Also, `onezone_ia`

    Failure parsing ObjectStorageClass from value: 'onezone_ia'. Accepted values: glacier, reduced_redundancy, standard, standard_ia"})

endgame:
Closing as dupe of #609

PR #570, opened 2020-03-24
Add intelligent tiering S3 type

k-bx:
Issue #569


k-bx:
This PR passed btw, but github isn't showing an updated status

PR #572, opened 2020-04-10
Fixes to compile on ghc-8.10.1

newhoggy:
* Bump upper bound of `haskell-src-exts`.
* Fix MonadUnlifIO compiler errors on ghc-8.10.1

### DIFF
--- a/amazonka/src/Control/Monad/Trans/AWS.hs
+++ b/amazonka/src/Control/Monad/Trans/AWS.hs
@@ -249,8 +249,12 @@ instance MonadBaseControl b m => MonadBaseControl b (AWST' r m) where
     restoreM     = defaultRestoreM
 
 instance MonadUnliftIO m => MonadUnliftIO (AWST' r m) where
+#if MIN_VERSION_unliftio_core(0,2,0)
+    withRunInIO inner = (AWST' $ (\(UnliftIO f) -> UnliftIO $ f . unAWST) <$> askUnliftIO) >>= \u -> liftIO (inner (unliftIO u))
+#else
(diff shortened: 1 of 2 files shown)


amazonka/src/Control/Monad/Trans/AWS.hs
```diff
@@ -249,8 +249,7 @@ instance MonadBaseControl b m => MonadBaseControl b (AWST' r m) where
     restoreM     = defaultRestoreM
 
 instance MonadUnliftIO m => MonadUnliftIO (AWST' r m) where
-    askUnliftIO = AWST' $ (\(UnliftIO f) -> UnliftIO $ f . unAWST)
-        <$> askUnliftIO
+    withRunInIO inner = (AWST' $ (\(UnliftIO f) -> UnliftIO $ f . unAWST) <$> askUnliftIO) >>= \u -> liftIO (inner (unliftIO u))
```
newhoggy, line 252: Basically inlined the old definition of `withRunInIO`: https://hackage.haskell.org/package/unliftio-core-0.1.2.0/docs/src/Control.Monad.IO.Unlift.html#withRunInIO

There may be a better way to write this.

newhoggy:
If this works, I'd like a new release to support `ghc-8.10.1`

brendanhay:
:+1:

TomMD:
Could we get a release to hackage with this merge?

ocharles:
Another request for a release :pray:

chris-martin:
`unlift-io-0.2` is still unsupported on Hackage - Is there anything we can do to help make this happen? I think you've got a lot of people willing to help here.

PR #575, opened 2020-04-29
Compile gen on lts-14.27

rossabaker:
Starts from https://github.com/brendanhay/amazonka/pull/470


rossabaker:
Oops.  Meant to incubate this in another repo.  Sorry for the noise.

PR #577, opened 2020-04-30
Support for ghc-8.6.5

rossabaker:
* Extends the work of #470
* Makes ghc-8.6.5 the default
* Relying on Travis CI to test on older ghc


brendanhay:
Haha, oops! I clicked ready for review wondering what the button did. :1st_place_medal:

rossabaker:
I was going to click the same once I knew it wouldn't be a CI disaster. :smile:

One job failed due to a gpg timeout.  I've restarted that one.  The overall build took about 12 hours.  Much of it was populating new caches, but subsequent builds will be heavier than old builds with the addition of a new GHC.  I think we should clean some old ones out of the Travis matrix before we merge.

I propose that we test on 8.6.5 and 8.8.3 (the latest two LTS) and 8.10.1.  I'll append a commit here to clean up the travis generator, and then follow up with PRs to support the newer ghc.

newhoggy:
👋

rossabaker:
We've been running a fork based on this for a while.  If this looks good, I'll proceed to 8.8.3.

peterbecich:
Please see https://github.com/brendanhay/amazonka/pull/606 for GHC 8.8.4

PR #578, opened 2020-05-01
Add `ignoredPaginators` to service definition.

rossabaker:
This lets one skip paginators defined in `paginators-1.json`, which
can be used to skip a paginator that fails to compile after a model
definition update.


brendanhay:
:shipit:

PR #579, opened 2020-05-12
Replace hindent with brittany

rossabaker:
hindent is unmaintained and fails to compile on lts-15.12 without some extra effort.

brittany parses in `IO`.  To avoid that ripple, we move the indentation step into the Makefile, next to stylish-haskell.

### DIFF
--- a/gen/Makefile
+++ b/gen/Makefile
@@ -4,6 +4,7 @@ MODEL_DIR := model
 OUT_DIR   := ..
 BIN_DIR   := ../bin
 GENERATE  := $(BIN_DIR)/amazonka-gen
+BRITTANY  := $(BIN_DIR)/brittany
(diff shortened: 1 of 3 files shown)


rossabaker:
It may be possible to adjust the defaults to be more like hindent to minimize the diff on the generated code.I doubt it would be perfect, and haven't tried.

brendanhay:
:+1: I've got a brittany config file that I'll add separately - I'm not worried about minimizing the diff at this point, it's just great to have this change done.

rossabaker:
I just discovered that brittany isn't working on 8.10.1 yet. (Neither is hindent, so we're no worse off.)

WIP: https://github.com/lspitzner/brittany/pull/305

PR #582, opened 2020-05-18
Import stuff from base instead of mtl

Bodigrim:
There is an intention to stop re-exporting functions from `base` in future versions of `mtl`: https://github.com/haskell/mtl/pull/74
This PR brings `amazonka` in line with the proposed change, because at the moment `Control.Monad.Trans.AWS` imports `MonadPlus` and `Network.AWS.Internal.HTTP` imports `join` (both originally from `Control.Monad`) from `Control.Monad.Reader`.


brendanhay:
Thank you

PR #591, opened 2020-07-31
Separate product and sum into own modules

rossabaker:
Large modules, such as `Network.AWS.ECS.Types`, take too long to compile.  This generates each generated sum and product into their own module, which are then re-exported in `Types`.  There are two advantages:

1. Smaller modules promote parallel compilation, so wall time drops.
2. Compilation of large modules is superlinear, so total CPU time drops. 

This change should be compatible.


rossabaker:
I wouldn't be surprised if bin packing improved it further.  Also wouldn't be suprised if the optimal bins depended on the cores.  The important thing is that this splits up the `EC2.Types` monstrosity. :smile:

PR #597, opened 2020-08-31
Generate sum types as pattern synonyms

rossabaker:
Sometimes, AWS introduces new values to its enums, which causes crashes when the Amazonka-generated sum types haven't caught up.  The encoding in this PR instead uses pattern synonyms over `CI String` for the known values, while permitting an unknown value.  For example:

```haskell
data PutMode = PutMode' (CI Text)
                 deriving (Eq, Ord, Read, Show, Data, Typeable,
                           Generic)

pattern Merge :: PutMode
pattern Merge = PutMode' "merge"

pattern Overwrite :: PutMode
pattern Overwrite = PutMode' "overwrite"

{-# COMPLETE
  Merge,
  Overwrite,
  PutMode' #-}
```

This should be source compatible.  We match values as before, but now get warnings if we don't handle unknown cases:

```haskell
pm :: PutMode -> Either Text Text
pm Merge = Right "M"
pm Overwrite = Right "O"
```

```
warning: [-Wincomplete-patterns]
    Pattern match(es) are non-exhaustive
    In an equation for ‘pm’: Patterns not matched: (PutMode' _)
   |
83 | pm Merge = "M"
   | ^^^^^^^^^^^^^^...
```

To make the code robust to new values, simply add the missing case:

```haskell
pm :: PutMode -> Either Text Text
pm Merge = Right "M"
pm Overwrite = Right "O"
pm (PutMode' m) = Left $ "Uh-oh, AWS added one: " <> (original m)
```

All instances are preserved for compatibility, though two of them become morally dubious in the presence of unknown values:

1. `fromEnum` becomes partial:

```haskell
instance Enum PutMode where
    toEnum i = case i of
        0 -> Merge
        1 -> Overwrite
        _ -> (error . showText) $ "Unknown index for PutMode: " <> toText i
    fromEnum x = case x of
        Merge -> 0
        Overwrite -> 1
        PutMode' name -> (error . showText) $ "Unknown PutMode: " <> original name
```

2. Our `Bounded` instance becomes a lie in the presence of unknown values:

```haskell
instance Bounded PutMode where
    minBound = Merge
    maxBound = Overwrite
```

```
> PutMode' (CI.mk "Aardvark") < minBound             
True
```

We've been running happily on this encoding for a couple months now.  I wanted to present it for consideration for 1.7, or maybe 2.0 if we want to think about pruning the problematic instances.

### DIFF
--- a/gen/src/Gen/AST/Data.hs
+++ b/gen/src/Gen/AST/Data.hs
@@ -147,11 +147,15 @@ sumData p s i vs = Sum s <$> mk <*> (Map.keys <$> insts)
   where
     mk = Sum' (typeId n) (i ^. infoDocumentation)
         <$> pp Print decl
+        <*> pure ctor
         <*> pure bs
 
-    decl = dataD n (map f . sort $ Map.keys bs) (derivingOf s)
+    decl = dataD n [newt] (derivingOf s)
       where
-        f x = conD (Exts.ConDecl () (ident x) [])
+        newt = conD (Exts.ConDecl () (ident ctor) [tyapp (tycon "CI") (tycon "Text")])
+
+    -- Sometimes the values share a name with a type, so we prime the data constructor to avoid clashes.
+    ctor = ((<> "'") . typeId) n
 
     insts = renderInsts p n $ shapeInsts p (s ^. relMode) []
 
--- a/gen/src/Gen/Import.hs
+++ b/gen/src/Gen/Import.hs
@@ -42,7 +42,8 @@ typeImports l = sort $
 
 sumImports :: Library -> [NS]
 sumImports l = sort $
-      "Network.AWS.Prelude"
+      "Data.CaseInsensitive"
+    : "Network.AWS.Prelude"
     : l ^. typeModules
 
 productImports :: Library -> Prod -> [NS]
--- a/gen/src/Gen/Tree.hs
+++ b/gen/src/Gen/Tree.hs
@@ -178,7 +178,6 @@ operation' l t o = module' n is t $ do
 
     is = operationImports l o
 
-
 shape' :: Library
        -> Template
        -> SData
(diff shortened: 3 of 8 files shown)


brendanhay:
I'm more in favour of pruning the `Bounded` and `Enum` instances completely, honestly, they just seem like a complete misfeature.

What's the idea behind using `CI Text` specifically, why not just `Text`?

rossabaker:
> I'm more in favour of pruning the Bounded and Enum instances completely, honestly, they just seem like a complete misfeature.

I only kept them for compatibility.  I struggled to think of a use case for them.  Dropdowns in some UI?  I'd be more than happy to drop them.

> I'm more in favour of pruning the Bounded and Enum instances completely, honestly, they just seem like a complete misfeature.

The parser treats them case insensitively.  It felt right to me to treat them case-insensitively across the board.

brendanhay:
Another misfeature 😃  I honestly cannot remember the original motivation for the case-insensitivity at this point. I suspect it was from a time when the service definitions/models were even more broken than they are now and I had no mechanisms for overriding the erroneous enumeration values.

rossabaker:
I can't find anything conclusive on whether they're supposed to be, and have never seen a contrasting pair that demands case sensitivity.  It makes parsing more generous, and encoding keeps the original.  We'll regret this if a case-sensitive pair ever does appear, but it seems to be a safe representation now.  I could go either way.

I'll rip out the `Bounded` and `Enum` instances of ill repute.

brendanhay:
Ace! 👌🏽 I'm going to merge this into `develop` - there's actually a in-flight transition from `develop` + `master` -> a singular `main` branch occuring for all my repos. For amazonka/gogol this means the introduction of nix/haskell.nix/etc. - so I'll hold of regenerating for the time being, until I can get these changes (+ others) merged into `main`.

PR #598, opened 2020-08-31
Add amazonka-eks module configuration

rossabaker:


rossabaker:
I can push the generated code and regenerated yaml in a separate commit here, or a separate PR.

Issue #604, opened 2020-11-03
Parameters necessary to compile dependency `text-icu` with Cabal on Mac


peterbecich:
I believe a library and parameters are necessary to compile `amazonka` and its dependency https://github.com/bos/text-icu with Cabal on a Mac.

The library `icu4c` is on Homebrew: https://github.com/haskell/haskell-ide-engine/issues/275

However, the solution inferred from https://github.com/haskell/haskell-ide-engine/issues/275 doesn't work:

```
cabal build all \
  --extra-lib-dirs=/usr/local/opt/icu4c/lib \
  --extra-include-dirs=/usr/local/opt/icu4c/include
```

The reason is: https://github.com/haskell/cabal/issues/2997

I have found this (from https://github.com/haskell/cabal/issues/2997#issuecomment-609608367) to work:
`cabal.project.local`
```
package *
  extra-include-dirs: /usr/local/opt/icu4c/include
  extra-lib-dirs: /usr/local/opt/icu4c/lib
```

endgame:
@peterbecich I don't have a mac. Can you check for me whether the following `cabal.project.local` works?

```
package text-icu
  extra-include-dirs: /usr/local/opt/icu4c/include
  extra-lib-dirs: /usr/local/opt/icu4c/lib
```

The hope is that if this fix works, then this becomes only `text-icu`'s problem.

@lrworth maybe if you have a chance, you could point your mac at this too? I suspect we haven't seen this problem because nix fixes the paths up for us. Do you see this if error if you grab GHC from `ghcup` and use `cabal` without Nix?

brendanhay:
These errors are specifically for `amazonka-gen` - the version in `develop` no longer has a direct dependency on `text-icu`.

PR #605, opened 2020-11-04
Fix compilation with Cabal

peterbecich:
I believe this PR fixes two Cabal compilation errors.

Updating the `ede` dependency appears to fix this build error: https://travis-ci.org/github/brendanhay/amazonka/jobs/716818638#L4516-L4541  Is it acceptable to point to an untagged commit for the `ede` dependency?

The second error was caused by the absence of `ClassA` https://github.com/peterbecich/haskell-src-exts/blob/d24b20323a4fe20ce476c4914e19c6ebb2c46cb3/src/Language/Haskell/Exts/ExactPrint.hs#L1174 in `haskell-src-exts` >= 1.22

Tested with GHC 8.6.5

https://github.com/brendanhay/amazonka/issues/604 may be necessary

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -1,1 +1,6 @@
 packages: */
+
+source-repository-package
+  type: git
+  location: https://github.com/brendanhay/ede.git
+  tag: 5a8756c1b3105af55b055f78ea273d00620a6731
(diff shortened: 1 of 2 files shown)


brendanhay:
Thanks @peterbecich - these changes are now incorporated into `develop` and `main`.

PR #612, opened 2021-03-08
added: load region data from aws file when use AWS_PROFILE

ncaq:
I was having trouble with the following error when I was programming.

```
message    = Just The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint.
```

The necessary permissions are given to the user.
For some reason,
it is working in the production environment (ECS).

After investigating,
I found out that this problem occurs only when I pass the authentication data using the environment variable `AWS_PROFILE`.

So I read amazonka's code and found out that `Region` is not set via environment variable.
So I referred to
[Configuration and credential file settings - AWS Command Line Interface](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure -files.html)
and configured it so that the setting is done automatically.

Note: This commit fails to build in my environment,
but that's a problem with the text-icu package.
[Build breaks with icu 68 - Issue #49 - haskell/text-icu](https://github.com/haskell/text-icu/issues/49)
It worked when I configured it to load the modified package.


endgame:
This has merge conflicts, but a 5-minute look at `amazonka/src/Network/AWS/Auth.hs` makes me think it's still relevant. Is it easy to recreate this PR against latest `develop`?

ncaq:
Maybe, I pull request read `~/.aws/credentials` not only config.

ncaq:
I often get stuck on this issue when developing applications and it confuses me.
I didn't know there was another implementation before I sent the PR, but I'd like to see it merged as soon as possible, whichever PR it is.

brendanhay:
This code reads `region` from `~/.aws/credentials` first, then tries `~/.aws/config`. I cannot find any [documentation](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html) that suggests this is intended - only the keys are stored in the credentials file. Even if it is intended behaviour, which value should be preferred is also ambiguous.

I've used this PR and the code from #480 as the basis for the commit https://github.com/brendanhay/amazonka/commit/bae6e40ad3ee5658fbee35f8d93373e4498e5f10. Thanks!

ncaq:
If the problem is solved, I welcome it even if my code is not incorporated.

LeifW:
AWS tooling seems to give precedence to the region set in the `~/.aws/credentials` file over that stored in `~/.aws/config`. I've since deleted my `~/.aws/config` file to avoid ambiguity.

PR #611, opened 2021-03-03
First steps toward Nix/CI

endgame:
#574 make it look like the lack of good build caching is a blocker for CI, which is a blocker for a release. This PR aims to take the first step in unblocking that.

It adds IOHK's `haskell.nix`, with pinning managed by `niv`. `ci.nix` sets up `haskell.nix` projects for GHC 8.6.5, 8.8.4, and 8.10.4, testing both the corresponding `stack.yaml` and the `cabal.project`.

I've tried to avoid making code changes or version bumps, but I needed new `unexceptionalio` to get `MonadFail` support, a tighter bound on `haskell-src-exts` to get `gen` to compile, and I had to tweak a test case so that a `HashMap`'s keys were listed in the same order when using both `hashable-1.3.0.0` and `hashable-1.3.1.0` (which changed the hashing algorithm).

The nix builds work but take a long time - I've tested all three compilers, both `cabal` and `stack` variants, by building the six derivations in the result set of `ci.nix`. They do take a while to start, possibly due to evaluating the giant nix expressions generated by `haskell.nix`. If you'd like, I can try to generate individual `haskell.nix` projects for each `amazonka-*` package.

Happy to rework any or all of this if it helps - anything to get the release cadence back up.

Issue #574, opened 2020-04-24
New hackage release


jchia:
It has been a year and many commits since the last release. It would be good for a new release version to be uploaded to Hackage especially now that there is GHC 8.10 and issues like #546 have been fixed.

domenkozar:
cc @brendanhay @MichaelXavier

brendanhay:
@domenkozar Somewhat ironically setting up cachix for amazonka/gogol now in an effort to unfuck CI and move from Travis to GH Actions.

domenkozar:
That's great to hear - happy to help if needed!

endgame:
@brendanhay This is exciting news. Anything an internet rando can help with? I speak pretty fluent nix.

gwils:
Hi @brendanhay - This is great news! How are you going with this?

RobertFischer:
I'm very willing to help -- I write/integrate/manage GitHub Actions professionally, so I'm happy to bang through those things in order to get a new release.  Even just bumping dependencies would be great.

gwils:
@brendanhay Maybe Robert and/or endgame could help you with this? A new hackage release would be great.

awgn:
Since stack TLS 17 is based on ghc-8.10.2 (to date) and given the MonadUnliftIO issue (fixed in 43ddd87b1ebd6af755b166e16336259ec025b337), it's not trivial to use amazonka as the base package does not compile (and it is not yet available under stack).

*A new release would be very appreciated!*

mbj:
me2 issues do not help @brendanhay as much. Instead I'd happily sponsor this library. @brendanhay Is there an option to support your work, financially? I can only send small amounts, but if we'd all do this the situation can improve significantly for you.

RobertFischer:
@brendanhay As I said before, I'm happy to help: just let me know what you need.  Like @mbj, I'd pitch you some money for a new release, too.

RobertFischer:
See #617

dciug:
Can we get an approximate date for a Hackage release? It's been another year.

emilypi:
Just fork it already, he clearly doesn't have the time to maintain this package.

mbj:
@emilypi there has been lots of activity in the `develop` branch. Effectively amazonka-2, improving a lot of the previous state.

RobertFischer:
But if it's not getting released, who cares how much work is done?

mbj:
[...]

PR #613, opened 2021-03-09
Ci take two

endgame:
This branch cherry-picks the build fixes from #611 and backports the nix/CI framework found on `main-develop`. I've tested that it successfully builds using the following commands:

* `nix-build -A workflows.core --argstr ghc884`
* `nix-build -A workflows.libs --argstr ghc884`
* `nix-build -A workflows.core --argstr ghc8104`
* `nix-build -A workflows.libs --argstr ghc8104`

I had to make some fixes to the nix framework:

* The `nixpkgs` passed to `haskell.nix` is provided by the `haskell.nix` framework. This increases the cache hit rate if you have IOHK's cache set up, and avoids building at least one copy of GHC.
* The `predicate` function passed into the nix local libs wasn't doing anything. Now `workflows.core` and `workflows.libs` are split correctly.
* I updated all the nix pins to get GHC 8.10.4.

I didn't bother testing all of {stack,cabal} with GHC {8.6.5,8.8.4,8.10.4}. I figured that if you're happy with the build matrix on `main-develop`, then so am i.

I'll cherry pick these fixed onto another branch so they can go back into `main-develop`.

cc @domenkozar @robertfischer in case they want to weigh in.

Closes #611 .


brendanhay:
Looks good! The cache is now updated for `linux-x86_64`, I'm still waiting on the `darwin-x86_64` builder to finish but once that's done I'll trigger the workflow run via merge.

RobertFischer:
SO HAPPY

RobertFischer:
This is a (the?) major step towards #574, too, so I'm pinging it.

endgame:
Hi @brendanhay , how'd you go with the darwin builder? Is there any assistance I can provide here?

brendanhay:
@endgame The Darwin build is fine, somewhat ironically, the Linux builds are [failing](https://github.com/brendanhay/amazonka/runs/2269974911) due to errors like the following:

```
# Logs
2021-04-05T13:28:48.6584471Z preallocating file of 27122 bytes: No space left on device

# Dashboard
You are running out of disk space. The runner will stop working when the machine runs out of disk space.
Free space left: 0 MB
```

The Darwin runner is in GitHub's own datacenter, but the Linux runners are just smol Azure DS2v2 instances with 14GiB (total) disk space:

image

The `ubuntu-latest` image which contains things like .NET core only gives you roughly 10GiB of free space to work with.

I haven't investigated further but it seems reasonable that Ubuntu + Nix + Cachix + realising the entire set of build derivations is over well over 10GiB. Unless someone has another idea, to try and work around this I'll split the job into 2 separate sequential workflows initially, doing a set of core libraries + cachix push, then the remaining set of actual API libraries in the second step to see if that's sufficient to get under the limit.

domenkozar:
That can happen when building lots of Haskell packages.

I suggest adding `    - run: sudo rm -rf /opt&` step to free up some space.

endgame:
I don't have a good feeling about that "split the build" idea, because nix doesn't even download GHC before it gets `ENOSPC`'d. Can you stick a `df -h` at the start and end of the mac run, and at the start of the linux run? That'll give us a good sense of how much space in which we have to work.

I found https://github.com/easimon/maximize-build-space , which does some ugly things with LVM to scrounge up more space. That action also has optional parameters to remove stuff like the .NET runtime, which you could crib before adopting such extreme measures.

endgame:
@brendanhay if you can generate a temporary read-only token for amazonka cachix, I will try and get CI working in my fork this weekend, and update the PR.

endgame:
@brendanhay belay that. I'm priming my own cachix to muck around with.

endgame:
`- run: sudo rm -rf /opt` allows the build to succeed, see https://github.com/endgame/amazonka/runs/2310419762 . However, there's only about 2.5G free on `/` after the build finishes, and that's with every derivation a cache hit. I will experiment some with https://github.com/easimon/maximize-build-space and report back.

endgame:
@brendanhay These are the changes I think should happen on your CI branch:

1. `- run: sudo rm -rf /usr/local/lib/android /usr/share/dotnet /opt&` runs in the background and removes ~42GB of crap. That should be more than enough, and avoids crazy LVM shenanigans.
2. Use this version of the `install-nix-action`, so you can pull from IOHK hydra:
   ```
   - uses: cachix/install-nix-action@v12
     with:
       extra_nix_config: |
         trusted-public-keys = hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ= cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=
         substituters = https://hydra.iohk.io https://cache.nixos.org/
   ```
3. If you get OOMs building `amazonka-ec2` or whatever, consider creating another/larger swapfile on `/mnt`.

You can use the `maximize-build-space` action to create a ~60GB `/nix`, but that seems brittle if you need space anywhere else in the system.

endgame:
I am now seeing timeouts from `hydra.iohk.io` on some of my test actions. Given that you've fully primed the caches, all the stuff you need from there should be in your cachix. Maybe it's better not to add their cache to the `extra_nix_config`?

endgame:
@brendanhay Thinking about closing this PR, since branch `bh/ci` has diverged significantly. Any reason I should keep it open? Also: anything else I can help with?

PR #617, opened 2021-04-30
Release Preparation

brendanhay:
Preparation for the next version and subsequent release. Note that this is not the release proper, it's intended to be a merge into `develop` prior to bumping versions, writing the changelog, and updating various documentation.

* Move to Nix for building and GitHub Actions for CI.  (Thanks @endgame)
* Numerous generator and library fixes with regards to dependencies.
* CPP dealing with < GHC 8.8 versions has been removed.
* Every `amazonka-*` package now reexports the `Network.AWS.Prelude` module.
* Enums (sum types) have been replaced with pattern synonyms. (Thanks @rossabaker)
* Record field labels match the service specification as closely as possible and `DuplicateRecordFields` is enabled by default. 
* Record smart constructors (previously `describeInstances`, `getObject`, etc.) are now strictly prefixed with `new`, such as `newDescribeInstances`.
* All type constructors (Such as record constructors) are now exported by default.
* Generated lenses are no longer exported from the top-level `Network.AWS.<name>` module - instead a `Network.AWS.<name>.Lens` module is provided.
* Generated lenses no longer use mnemonic or heuristically assigned prefixes such as `dirsrsInstances` and instead strictly prefix using the type name `describeInstances_instances` - following the form `<type>_<field>`.

Given the original non-opensource amazonka is close to 9 or 10 years old at this point, the use of record prefixing has not aged  well - particularly given the popularity of `generic-lens`, `optics` etc. Problems arising from the use of `unordered-containers` and subsequent prefix changes when the hashing algorithm in `Hashable` change{d,s} or the ordering keys in the service definition changes. There's plenty of ways this could be mitigated, but honestly I'd rather just have field names that map directly to AWS documentation or are guessable without having to refer to Haddock documentation that crashes my browser.

Another follow on effect from the original design was an attempt to have internal vs external types to control serialisation - via the `List1`, `Map`, `Nat`, etc. types from `Network.AWS.Data.*` and then the generated lenses were responsible for translating between these types and a more palatable representation such as `NonEmpty`, `HashMap`, `Natural`, etc. I've made a small inroad to the removal of these types due to the lens changes above, but I have fairly strong opinions nowadays that serialisation should no longer be type directed (via custom newtypes) and instead should be the purview of the generator to construct the necessary de/serialisers via function calls, other inroads into this were made in the `main-develop` branch.

The net effect of all these changes is hopefully a much more maintainable generator and core libraries. In particular, using CPP to support wider and wider compiler version ranges or very old dependency bounds, or maintaining a growing number in-tree `stack.yaml`s is a headache given how much time/effort it takes to build the project. Moving forward I'd like to support much more restrictive compiler versions, say, at most 2 major versions.

For remaining changes not part of this branch:

1.  I'm undecided on the version number but am leaning towards `2.0`given how many breaking changes are included.
2. There's quite a lot of documentation to be updated and examples to be fixed to work with the new lenses - as well as integration testing.
3. Merge any outstanding PRs that are either trivial, must haves for the release, or require only a minimal amount of work to bring them in line with these changes. 

Changes that I wanted in amazonka >= 2.0, but probably aren't going to happen:

3. Migrate entirely from `Network.AWS.*` style namespaces to `Amazonka.*`.
4. Supersede (via deprecation) `Network.AWS` and `Control.Monad.Trans.AWS` in favour of something similar and completely effect agnostic of the form `send :: Env -> ...`. You can then use whatever effects system floats your boat and I don't have to maintain mtl nonsense.
5. Removal of `http-conduit`, `xml-conduit`, and `resourcet` in favour of a `sendWith :: ... -> (a -> IO b) -> b` continuation style.
6. Replace `Network.AWS.{Text,ByteString,Log} with `http-api-data`.
7. Removal of non-essential dependencies (despite 5.)

Once this PR is merged I plan to create a new `main` branch and set that as the new default branch. This matches git's new default and is an opportunity to do away with the `master` vs `develop` split. Releases when then be tags against `main` and all work on `master` and `develop` would cease, with both branches being preserved to avoid breaking anyone's builds.

Closes #613


endgame:
`itshappening.gif`

Great work. There's way too much diff to think about reviewing (and most of it's autogenerated), so I'm just going to assume that it's sound. I'm a big +1 for the 2.0.0.0 version number - changes like this definitely feel like a good use of that second major version number that PVP allows for.

Re: outstanding PRs/Issues: I think that since things have been stalled for so long, an aggressive triage pass might be a good idea. Something like the following, perhaps?

1. Tag all current open PRs/issues with "needs triage"
2. Work through all "needs triage" issues, closing/merging/asking for clarification as appropriate. Anything that needs clarification loses the "needs triage" label, and gains a "awaiting response" label.
3. Put a time limit on "awaiting response" issues, close if they're not responded to promptly, but invite contributors to reopen/rebase PRs off the new `main` branch.

I'm happy to help out with such a triage process.

Re: Integration testing: I think once this PR lands, you might want to think about loudly announcing it, and invite people to set up `source-repository-package` stanzas in their cabal files (or `stack.yaml` equivalent) to test before 2.0.0.0 makes it to Hackage.

Re: Documentation: is there a tracking issue for this? I'm not sure where to begin looking for broken documentation.

RobertFischer:
I believe this closes #574, too, and supersedes #613.

axman6:
This is super exciting, thanks @brendanhay and everyone else involved, it’s a gargantuan effort to get this far. I definitely agree this looks like a 2.0 version, just based off the summary. I’ll try and find some time tonight to review what I can, because any problem found before release will save everyone time.

I think it might be worth considering doing a beta release and giving people, say, a month to see what effects migrating will have on their projects - for a lot of user who use this in production, this would be a much bigger upgrade to deal with than a GHC release, so getting it on hackage, and making a beta announcement would help a lot in making sure most of the wrinkles get ironed out. Sadly I’m not doing anything on AWS at the moment so wouldn’t be a good beta tester (but maybe it’s give me the impetus to rewrite `amazonka-s3-streaming` [and maybe rename it after all these years - still looking for a better name if anyone has one])

endgame:
> but maybe it’s give me the impetus to rewrite amazonka-s3-streaming [and maybe rename it after all these years - still looking for a better name if anyone has one]

@axman6 : what if @brendanhay were to officially recommend the `amazonka-contrib-*` namespace for non-generated amazonka-related libraries?

endgame:
> Does this supersede #613?

Yes. The OP has "Closes #613" in it.

brendanhay:
@endgame I'll start adding labels over the weekend for triage - there are a couple of PRs such as #598 that I'll go ahead and get merged without any outreach, as they're pretty straight forward and the timing is suitable.

There are also no documentation issues yet - I'll need to identify the various docs and examples that have been outdated by this PR and make note in a new issue.

@axman6 I wouldn't worry too much about the naming of `amazonka-s3-streaming`, honestly - just keep it the way it is.

Ideally Cabal's [multiple public libraries](https://cabal.readthedocs.io/en/3.4/cabal-package.html#pkg-field-library-visibility) would actually be usable for this project and we could have a single `amazonka` library with public sublibraries for each service, but alas, we end up with people downloading over 3 million lines of code in a source archive of which they will only use a tiny fraction.

As for package naming - It hasn't (in ten years) become the problem I was originally concerned about so I sort of hesitate to try and dictate how people name their packages, other than use common sense. We can certainly recommend a contrib style in the README, or at least update the wording.

I think the integration testing and beta release / preview window comments from above are on point - I'll create a `2.0.*-beta` tag and offer example stack/source-package-repository stanzas in the README and possibly on reddit, as well as push a Hackage [candidate](https://hackage.haskell.org/upload#candidates).

My immediate plan is:

* Merge this PR.
* Work through specific outstanding PRs I want as part of any potential release (ie. EKS).
* Get the triage process going for the remaining outstanding PRs.
* Briefly revisit some of the "nice to have" changes (3. and 4.) from the PR description.

endgame:
Sounds good. Let me know if you want a hand with triage, and if I need any additional permissions to do that.

RobertFischer:
Happy to help however I can. Just let me know.

PR #626, opened 2021-05-01
Drop Some MonadThrow and MonadCatch Constraints

brendanhay:
Closes #482


brendanhay:
Nothing like being faced with many years of my own cumulative poor choices - why wasn't `throwIO` used? 🤷🏽 

image

PR #633, opened 2021-06-17
Add Amazon SES V2 support, fixes #632

### LINKED ISSUE #632: Support Amazon SES v2
Amazon SES has a [V2 API](https://docs.aws.amazon.com/ses/latest/APIReference-V2/Welcome.html) which includes some new functionality (e.g. list management). There seems to be a definition for it in [the botocore repo](https://github.com/boto/botocore/tree/develop/botocore/data/sesv2/2019-09-27) so if I understand it correctly, it should not be hard to generate a `amazonka-sesv2` library from that. And it would be really useful!

jbransen:
This adds the config for Amazon SES v2, which already exists in the botocore lib and seems to work out of the box (I've not tested all methods, but it compiles and the api calls I tried succeed).

This fixes #632

### DIFF
--- a/config/annexes/sesv2.json
+++ b/config/annexes/sesv2.json
@@ -1,0 +1,5 @@
+{
+    "metadata": {
+        "serviceAbbreviation": "SESv2"
+    }
+}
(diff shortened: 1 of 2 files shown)


brendanhay:
Great, thanks :+1: - I'll merge and regenerate the SES v2 support in a new branch for ci to build.

PR #639, opened 2021-09-28
Replace generate.nix with in-tree generator script

brendanhay:
The current iteration of `generate.nix` somewhat moronically copies a bunch of unnecessary paths (like, the whole repo) in/out of `/nix/store` causing it to eat a bunch of disk space.

This PR replaces most of the logic from `generate.nix` with an equivalent in `script/generate`, using a `nix-shell` shebang and GNU parallel to marginally speed up the formatting phase. The audit step which was used to diff what services were un/implemented has been skipped for now.


default.nix
```diff
@@ -32,17 +37,24 @@ in project // {
...
 
     tools = {
-      cabal = "3.2.0.0";
-      cabal-fmt = "0.1.5.1";
-      shellcheck = "0.7.1";
-      ormolu = "0.3.0.1";
+      cabal-fmt = {
+        inherit (pkgs.cabalProject) index-state;
+        version = "0.1.5.1";
+      };
```
endgame, line 43: Can we use `pkgs.haskellPackages.cabal-fmt` in `buildInputs`, or is that version too old?

endgame (approved):
Seems kinder to people's disks, I agree. Is there a way to work around the lack of audit functionality, or do you have a plan to get that in soon?

default.nix
brendanhay: If we use `pkgs.haskellPackages.cabal-fmt` we need to potentially download another GHC from the cache - by using `tools` it ensures it uses the same (patched) GHC from haskell.nix.

brendanhay:
I'll try and do the audit functionality in a separate pass - probably by exposing the previous diffs (available services in botocore vs services in configs/services) as attributes from `default.nix`, then just munge those in the bash script.

PR #640, opened 2021-09-28
Minimum GHC version supported is 8.6

ysangkok:
This allows GHC 8.6 which is still widely supported in the ecosystem and there seems no reason to not allow it, since it builds and passes the test suite. My employer is currently using GHC 8.6 and it would be nice if we could upgrade amazonka before we switch compilers.

Tested by doing `cabal run -w ghc-8.6.5` in the `amazonka/amazonka` directory which ran the test suite, which passed.

Note that this is not only changing base versions but also adjusting [the note](https://github.com/brendanhay/amazonka/compare/develop...ysangkok:allow-ghc-8.6?expand=1#diff-8065007e4fa64061c955805c98352c64e5cb8544cf78ea45b241785ef17ce6eaL27) that specifies the minimum supported GHC version.

### DIFF
--- a/amazonka-s3-encryption/amazonka-s3-encryption.cabal
+++ b/amazonka-s3-encryption/amazonka-s3-encryption.cabal
@@ -74,7 +74,7 @@ common base
     TypeFamilies
     ViewPatterns
 
-  build-depends:      base >=4.13 && <5
+  build-depends:      base >=4.12 && <5
 
 library
   import:          base
--- a/amazonka/amazonka.cabal
+++ b/amazonka/amazonka.cabal
@@ -69,7 +69,7 @@ common base
     TypeFamilies
     ViewPatterns
 
-  build-depends:      base >=4.13 && <5
+  build-depends:      base >=4.12 && <5
 
 library
   import:          base
--- a/config/templates/cabal.ede
+++ b/config/templates/cabal.ede
@@ -58,7 +58,7 @@ library
 
     build-depends:
           amazonka == {{ clientVersion }}.*
-        , base     >= 4.13 && < 5
+        , base     >= 4.12 && < 5
       {% for dep in extraDependencies %}
         , {{ dep.value }}
       {% endfor %}


endgame:
This appears to have changed only the generated cabal files, which will be clobbered next time services are regenerated. Can you prepare a PR that only changes the cabal file template? (Generate and test locally, though, please.)

brendanhay:
I'm hesitant to do this. Dropping support for older compilers was intentional and allowed us to remove a lot of `CPP` code and reduce the amount of time spent building the same code over and over locally and in CI. 

So while it may build on GHC `8.6.5` (`base-4.12.0.0` specifically) - I won't add it to the list of supported (read: encouraged) compilers in the documentation because we still won't build it on that version due to resource constraints. (143 packages * 2 compilers * 2 operating systems, currently.)

If you do want to make a PR for the lower bound bump then please make the changes in:

* [config/templates/cabal.ede](https://github.com/brendanhay/amazonka/blob/0433ee599feabd747a164bbc4445a08ff5364603/config/templates/cabal.ede#L61)
* [amazonka/amazonka.cabal](https://github.com/brendanhay/amazonka/blob/0433ee599feabd747a164bbc4445a08ff5364603/amazonka/amazonka.cabal#L72)
* [amazonka-s3-encryption/amazonka-s3-encryption.cabal](https://github.com/brendanhay/amazonka/blob/develop/amazonka-s3-encryption/amazonka-s3-encryption.cabal#L77)

ysangkok:
@brendanhay What version would you like to put in the README? Currently it says 8.2.1.

I understand that you can't run CI on every configuration (BTW your list didn't consider the different architectures supported by GHC). It would seem a bit arbitrary to support an obscure architecture while not supporting an obscure GHC version, though.

So I'll re-do the PR in the suggested way once I know which version to put in the README.

brendanhay:
> What version would you like to put in the README? Currently it says 8.2.1.

That will be updated to `8.8.4` or whatever the oldest version that we're building on CI is come 2.0 release. To be clear - I just don't want to signal support for something we don't test on - nor do I want Hackage trustees to have to manually edit metadata to restrict bounds to get things building on older compilers. 

> It would seem a bit arbitrary to support an obscure architecture while not supporting an obscure GHC version, though.

The current CI build matrix is:

```
linux-x86_64: ghc884
linux-x86_64: ghc8107
darwin-x86_64: ghc8107
```

ysangkok:
@brendanhay Ok I have force pushed the branch, only changing the three files you listed.

endgame:
@ysangkok Thanks very much.

I cherry-picked this commit onto `develop`, regenerated all services, and ran `nix build -f . ci --argstr ghcVersion ghc865`. I had to adjust the bounds on `gen/amazonka-gen.cabal` and `test/amazonka-test.cabal` so that it would find a coherent build plan, but after that everything built successfully. Since we're not planning on using CI for GHC 8.6.5, I think that's fine.

@brendanhay I'm not clicking merge on this because I don't know if you have some process around where and when you regenerate services, but I think the change is sound.

brendanhay:
I'll let CI build-and-cache things, then merge. After that I'll regenerate all the services in a separate branch. It's easier if we (collaborators) do the regeneration and localise it to a new branch to keep things reviewable.

Issue #509, opened 2019-01-11
Amazon Textract support


alexanderkjeldaas:
https://aws.amazon.com/textract/

endgame:
Crossing fingers that this is just "get service description, generate and go".

ketzacoatl:
@endgame Given the position we're in, I would advocate for for new services being a separate 2.1 release or whatever. Eg, make the milestone for 2.0 about dire issues, release blockers, and stuff that must be done for the release. Let the rest fall into a follow-up release we make soon after. The sooner we get 2.0 out, the sooner we'll get a 2.1 and on. Just my 2c, feel free to ignore :)

brendanhay:
Now part of `develop`.

PR #644, opened 2021-09-29
Add script to audit and generate missing service configurations

brendanhay:
Adds simplistic audit functionality from `generate.nix` that was removed in #639. Additionally, it now writes example skeleton configurations and annexes to the output directory specified by the singular script argument.

Example:

```bash
% generate-configs tmp
...
Missing wafv2
Missing wellarchitected
Missing wisdom
Missing worklink
Missing workmailmessageflow
Missing 141 service configurations.
Wrote examples to tmp/configs and tmp/annexes.
Found 281 models in /nix/store/bvn31qxjpffkaa5misrb8y0kxphhy0gw-source/botocore/data.
Found 140 service configurations in config/services.
Done.
```

The reason for also generating an empty annex file is illustrated by testing with the currently unsupported `textract` service (#509) - it requires `metadata.serviceAbbreviation` to be specified since it's missing from the definition and this is used to infer the root module namespace. At some point in the future I'll get around to merging the configurations to remove the need for annex at all.

### DIFF
--- a/scripts/generate-configs
+++ b/scripts/generate-configs
@@ -1,0 +1,67 @@
+#!/usr/bin/env nix-shell
+#!nix-shell -i bash ../shell.nix
+
+# Usage: generate-configs OUTPUT
+#
+# Generates configuration for all services found in BOTOCORE_PATH that
+# are missing corresponding configurations in ./config/services.
+#
+# The example configurations will be written to individual files under
+# OUTPUT/configs and OUTPUT/annexes.
+
+set -euo pipefail
+
+cd "$(dirname "${BASH_SOURCE[0]}")/.."
+
+botocore="${BOTOCORE_PATH?}"
+services="config/services"
+output="${1?An OUTPUT argument is required}"
+configs="$output/configs"
+annexes="$output/annexes"
+
+available=()
+configured=()
+unconfigured=0
+
+# Output a message to stderr.
+say() {
+  echo >&2 "$@"
+}
+
+# Write an example configuration and annex to name.json.
+write_config() {
+  local -r name="$1"
+  local -r config="$configs/$name.json"
+  local -r annex="$annexes/$name.json"
+
+  say "Missing $name"
+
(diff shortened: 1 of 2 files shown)


scripts/generate-configs
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env nix-shell
+#!nix-shell -i bash ../shell.nix
```
endgame, line 2: Is `nix-shell` smart enough to resolve this relative to the script, and not to the working directory?

scripts/generate-configs
```diff
@@ -0,0 +1,67 @@
...
+# Write an example configuration and annex to name.json.
+write_config() {
+  local -r name="$1"
+  local -r config="$configs/$name.json"
+  local -r annex="$annexes/$name.json"
+
+  say "Missing $name"
+
+  jq -n --arg name "$name" '{ libraryName: $name }' >"$config"
+  jq -n --arg name "$name" '{}' >"$annex"
```
endgame, line 40: What's this for? Won't it always return `{}`?

endgame (approved):

scripts/generate-configs
brendanhay: Yeah, it will create an empty annex file per:

> The reason for also generating an empty annex file ..

It's just more convenient to pre-create the file since the majority of new service setups needs `serviceAbbreviation` added there.

scripts/generate-configs
brendanhay: It resolves relative to the script location - testing that out did reveal that `cabal` needs to be added to the `buildInputs`, though.

scripts/generate-configs
endgame: I meant the use of `jq` over `echo {} > "$annex"` (using `jq` for symmetry, perhaps?)

brendanhay: Ah, yeah, no other reason.

Issue #585, opened 2020-06-16
Support new instance types


saschat:
Personally, I am interested in T2 and R6g instances but there have been a lot of new instance types since the release of 1.6.1.

endgame:
Current version of `develop` has these instance types:

https://github.com/brendanhay/amazonka/blob/a266fc686cdddb48ac791261b73798c2ed8a6097/amazonka-ec2/gen/Network/AWS/EC2/Types/InstanceType.hs#L345-L370

Since @brendanhay is hoping to release soon, I think this is closeable.

endgame:
AWS just released [vt1 instances](https://aws.amazon.com/blogs/aws/new-amazon-ec2-vt1-instances-for-live-multi-stream-video-transcoding/) so we really should regenerate ec2 before 2.0.

brendanhay:
I've regenerated EC2 as part of #645, so the missing instance types including VT1 are now present.

It's worth reiterating that all enumerations on `develop` (including `InstanceType`) are now defined as `newtype` over `Text` and the constructors (such as `InstanceType_R6g_metal`) are just pattern synonyms. In future, if instance types, regions, or any other enumerations are missing you can just use the `IsString` instance and write it directly as `"vt1.24xlarge"`, for example.

PR #645, opened 2021-09-29
Regenerate all services using latest definitions

brendanhay:


brendanhay:
Fixes #585

brendanhay:
Closes #509 (accidentally.)

PR #646, opened 2021-09-29
ci: Add -build-flags "-L --keep-going" to nix-build-uncached

endgame:
This causes `nix-build-uncached` to emit logs, and to build (and therefore cache) as much as possible on each CI run.

The theory is that on a hot cache, it should be easier to find out what went wrong because you should only see build logs of broken things.


endgame:
Ah, never mind. The system I'm cribbing from must be getting a different nix version.

PR #647, opened 2021-09-30
gen: Correct generator invocation example

JackKelly-Bellroy:
I know `generate.nix` went away, but is this the correct invocation?


brendanhay:
Yeah it is, thank you!

PR #649, opened 2021-09-30
S3 long ints

### LINKED ISSUE #320: S3 gorsContentLength should be Int64 instead of Int
In 32 bit versions of GHC, the "Int" type is 32 bits. This means that `gorsContentLength :: Lens' GetObjectResponse (Maybe Int)` will overflow on S3 files that are larger than 2GB.

I believe that the correct type should be "Int64" (just like how lazy ByteString uses Int64 for length: https://hackage.haskell.org/package/bytestring-0.10.8.1/docs/Data-ByteString-Lazy.html#v:length)

I believe the same also applies to `horsContentLength`, `poContentLength`, `upContentLength`

endgame:
Closes #320


endgame:
Reported upstream: https://github.com/boto/botocore/issues/2509

PR #654, opened 2021-09-30
[#651] Allow 'profile ' in `~/.aws/config` file sections

### LINKED ISSUE #651: amazonka terminates when there is a `profile ` prefix in `~/.aws/config`'s sections
in `~/.aws/config` file, labels for profiles other than `default` are [allowed](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html) to have a `profile `  prefix. 

> The credentials file uses a different naming format than the AWS CLI config file for named profiles. Include the prefix word "profile" only when configuring a named profile in the config file. Do not use the word profile when creating an entry in the credentials file. 

With a [recent change that allows us to read `region` from config file](https://github.com/brendanhay/amazonka/commit/bae6e40ad3ee5658fbee35f8d93373e4498e5f10) `amazonka` assumes that section labels do not have that `profile ` prefix, which causes amazonka to terminate the whole process with an error:

```
<executable>: user error (mzero)
```

which is caused by [this exact line](https://github.com/brendanhay/amazonka/blob/bae6e40ad3ee5658fbee35f8d93373e4498e5f10/amazonka/src/Network/AWS/Auth.hs#L518).

Removing that prefix from `~/.aws/config` labels mitigates the issue. But the better solution is to:
1. not make amazonka throw this error that does not help
2. also scan for `profile ` prefixed version of the profile during lookup.

korayal:
closes #651

### DIFF
--- a/amazonka/src/Network/AWS/Auth.hs
+++ b/amazonka/src/Network/AWS/Auth.hs
@@ -514,11 +514,12 @@ fromFilePath profile cred conf =
         else do
           ini <- INI.readIniFile path >>= either (throwInvalid path Nothing) pure
 
-          let regionValueE =
-                INI.lookupValue profile confRegion ini
-                <> INI.lookupValue ("profile " <> profile) confRegion ini
+          let configProfile =
+                if profile == "default"
+                then profile
+                else "profile " <> profile
 
-          case regionValueE of
+          case INI.lookupValue configProfile confRegion ini of
             Left _ -> pure Nothing
             Right regionValue ->
               case fromText regionValue of
--- a/amazonka/src/Network/AWS/Auth.hs
+++ b/amazonka/src/Network/AWS/Auth.hs
@@ -514,10 +514,14 @@ fromFilePath profile cred conf =
         else do
           ini <- INI.readIniFile path >>= either (throwInvalid path Nothing) pure
 
-          case INI.lookupValue profile confRegion ini of
-            Left _ -> empty
-            Right value ->
-              case fromText value of
+          let regionValueE =
+                INI.lookupValue profile confRegion ini
+                <> INI.lookupValue ("profile " <> profile) confRegion ini
+
+          case regionValueE of
+            Left _ -> pure Nothing
+            Right regionValue ->
+              case fromText regionValue of
                 Left err -> liftIO (throwInvalid path (Just confRegion) err)
                 Right ok -> pure (Just ok)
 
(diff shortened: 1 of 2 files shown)


endgame:
The code does not compile but the intent seems to be pointing in the right direction. I'll have to read up on the details of the config format AWS uses before coming back to this.

endgame:
Alternatively, if you can fix the compile errors and then ping me before then, I'll be happy to take another look. Thanks for the PR.

korayal:
Yeah, sorry about that, I'll fix it in a few hours.

amazonka/src/Network/AWS/Auth.hs
```diff
@@ -514,8 +514,15 @@ fromFilePath profile cred conf =
...
-          case INI.lookupValue profile confRegion ini of
-            Left _ -> empty
+          let configProfile =
+                case INI.sections ini of
+                  sections@(_:_)
+                    | profile `elem` sections -> pure profile
+                    | ("profile " <> profile) `elem` sections -> "profile " <> profile
+                  _ -> throwInvalid path Nothing ("profile " <> profile <> " does not exist in config file")
+
+          case INI.lookupValue configProfile confRegion ini of
```
jhrcek, line 524: Optional suggestion: you might consider achieving this more simply by Alternative instance of Either:
```haskell
case INI.lookupValue configProfile configRegion ini 
  <|> INI.lookupValue ("profile " <> configProfile) configRegion ini of
  ...
```

korayal: Thanks for the suggestion, but I would rather use [`hush`](https://hackage.haskell.org/package/errors-2.3.0/docs/Control-Error-Util.html#v:hush) than using `<>` for this purpose.

We also lose our warning in case we're missing the necessary profile section with this change, but maybe that's not the problem of `amazonka` anyway.

korayal: on the other hand, the package doesn't have `errors` library in its dependencies. so hush be available for me anyway. changing it to your suggestion. (still waiting for my nix environment to come up)

endgame:
I'll get to it as soon as I can. Meanwhile, can you please rebase against latest `develop` and add an entry to `amazonka/CHANGELOG.md`?

korayal:
done.

brendanhay (approved):

endgame:
@korayal thank you for this.

PR #655, opened 2021-09-30
Prefer newer files on exact duplicates

MichaelXavier:
The same issue from #637 has recurred where files with the same name but
different case cause problems on Mac due to the case-insensitive
filesystem. Previously, my script would prefer newer files where there
was a conflict *and* differing file content, but it would prefer *older*
files when the files were the exact same. In retrospect that was a
mistake because if the conflicting files are being automatically
generated, that script is just going to regenerate the files that were
deleted every time. Below is the updated script I used to resolve this
issue. I've spot checked the files in the diff and the ones being
removed are indeed the older ones.

```python

import sys
import hashlib
import subprocess

def md5(fname):
    hash_md5 = hashlib.md5()
    with open(fname, "rb") as f:
        for chunk in iter(lambda: f.read(4096), b""):
            hash_md5.update(chunk)
    return hash_md5.hexdigest()

def file_git_timestamp(fname):
    return int(
        subprocess.check_output(["git", "log", "-1", "--pretty=%ct", fname])
        .decode("utf8")
        .strip()
    )

files = {}

for file in map(str.rstrip, sys.stdin):
    normalized = file.lower()
    if normalized not in files:
        files[normalized] = []
    files[normalized].append(file)

exacts = []
diffs = []

for k, v in files.items():
    if len(v) >= 2:
        hashes = list(map(lambda f: md5(f), v))
        if all(hashes[0] == x for x in hashes):
            exacts.append(v)
        else:
            diffs.append(v)

for exact in exacts:
    newest_to_oldest = sorted(exact, key=lambda x: file_git_timestamp(x), reverse=True)
    print(f"Keep {newest_to_oldest[0]}")
    for kill in newest_to_oldest[1:]:
        print(f"git rm {kill}")
        subprocess.check_call(["git", "rm", kill])

for diff in diffs:
    newest_to_oldest = sorted(diff, key=lambda x: file_git_timestamp(x), reverse=True)
    print(f"Keep {newest_to_oldest[0]}")
    for kill in newest_to_oldest[1:]:
        print(f"git rm {kill}")
        subprocess.check_call(["git", "rm", kill])

print(f"{len(exacts)} exact matches")
print(f"{len(diffs)} differing matches")

```


endgame:
Makes sense. Any chance you can add this PR to the changelog entry?

https://github.com/brendanhay/amazonka/blob/00b157f512b0a36e0e71e341a3ed29a7900f8a1e/amazonka/CHANGELOG.md#L127

endgame:
Actually I'll just fix the changelog straight on develop.

PR #659, opened 2021-09-30
Update version number

endgame:
Figured we have to do this at some point. Unfortunately the nix shell fails to start if it can't find a coherent build plan, so the easiest thing to do is probably to delete `amazonka-*` and then generate. But that nukes all the `Setup.hs` files, as well as fixtures|yamls|protos. So then you have to resurrect all that after. Maybe that's fine because it's only a problem on major breaking releases, or maybe @brendanhay has a better idea how to handle this?


cabal.project
```diff
@@ -17,4 +17,6 @@ packages:
    gen/
    test/
    amazonka/
+
+optional-packages:
```
endgame, line 21: `cabal` will error if a glob fails to match, which it does if you delete every generated package.

brendanhay: Makes sense 👍🏽

brendanhay:
I'm sketching something out in another branch now to split the use of `haskell-nix` and it's subsequent requirement to have a valid plan from the shell the scripts/tools actually care about.

endgame:
Makes sense. I had an aborted sketch that used the `subDir` option to haskell.nix's `cabalProject` function, but while that got me a shell, it still fell over when the `cabal new-run` in `scripts/generate` read the full `cabal.project` function. So I threw it out.

Issue #379, opened 2017-03-31
presigned url does not include needed query params


shlevy:
```haskell
sourceRequest :: CreateDBInstanceReadReplica
sourceRequest =
  createDBInstanceReadReplica "" "arn:aws:rds:us-west-2:fake:db:fakey" & cdirrKMSKeyId ?~ "arn:aws:kms:us-west-1:fake:key/faker"

main :: IO ()
main = do
  lgr  <- newLogger Debug stdout
  env <- credFile >>= newEnv . FromFile "somatix-dev"
  now <- getCurrentTime
  presigned <- runResourceT  $ runAWS (env & envLogger .~ lgr) $
    presignURL now 10 sourceRequest
  putStrLn $ show presigned

```
Gives
`"https://rds.amazonaws.com/?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=CREDREDACTED%2F20170331%2Fus-east-1%2Frds%2Faws4_request&X-Amz-Date=20170331T153116Z&X-Amz-Expires=10&X-Amz-SignedHeaders=content-type%3Bhost&X-Amz-Signature=SIGREDACTED"`

We need other query params there, e.g. `KmsKeyId`

shlevy:
@brendanhay Any thoughts on this?

endgame:
OK, did some digging. I assume this is for the `PreSignedUrl` fields that cross-region RDS requests need (#378)? `amazonka-rds` uses the default V4 signer:

https://github.com/brendanhay/amazonka/blob/53bc5a196acebc6bb7c40da71223c8817fcb6f90/amazonka-rds/gen/Network/AWS/RDS/Types.hs#L1314

However, the default signer appears to have a hardcoded list of headers that it includes:

https://github.com/brendanhay/amazonka/blob/53bc5a196acebc6bb7c40da71223c8817fcb6f90/amazonka/src/Network/AWS/Sign/V4.hs#L38-L43

I suspect that we'll need to override the signer for `amazonka-rds` somewhere if the `PreSignedUrl` field is present on any of the five request types mentioned in #661 (which has a link to the botocore hackery that does similar things). Would be neat to have but definitely 2.0 material unless someone else makes a PR.

PR #661, opened 2021-10-02
amazonka-rds: Support DestinationRegion pseudo-parameter

### LINKED ISSUE #378: RDS.CreateDBInstanceReadReplica missing psuedo "DestinationRegion" parameter
From http://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateDBInstanceReadReplica.html:

> PreSignedUrl
> The URL that contains a Signature Version 4 signed request for the CreateDBInstanceReadReplica API action in the AWS region that contains the source DB instance. The PreSignedUrl parameter must be used when encrypting a Read Replica from another AWS region.
> 
> The presigned URL must be a valid request for the CreateDBInstanceReadReplica API action that can be executed in the source region that contains the encrypted DB instance. The presigned URL request must contain the following parameter values:
> 
> DestinationRegion - The AWS Region that the Read Replica is created in. This region is the same one where the CreateDBInstanceReadReplica action is called that contains this presigned URL.
> For example, if you create an encrypted Read Replica in the us-east-1 region, and the source DB instance is in the west-2 region, then you call the CreateDBInstanceReadReplica action in the us-east-1 region and provide a presigned URL that contains a call to the CreateDBInstanceReadReplica action in the us-west-2 region. For this example, the DestinationRegion in the presigned URL must be set to the us-east-1 region.
> {snip}

`DestinationRegion` is never a parameter for a real request, but we nevertheless need to be able to create a `CreateDBInstanceReadReplica` with `DestinationRegion` set to generate the `PresignedUrl`

endgame:
Cross-region replication on RDS is performed by making a request
against the _destination_ region. In that request, you need to pass a
`PreSignedUrl` for the _source_ region, with the `DestionationRegion`
parameter set to the _destination_ region.

There are five such requests at the moment, specified by

https://github.com/boto/botocore/blob/77359646878e261eaa9debad888fc8f3b20832ef/botocore/data/rds/2014-10-31/service-2.sdk-extras.json

These specify a `SourceRegion` parameter because botocore rearranges
the rds requests internally, stripping `SourceRegion` from the request
itself and injecting a `DestinationRegion` parameter into the
`PreSignedUrl` field:

https://github.com/boto/botocore/blob/ec0eb680d6e6e753c7d7ddf0242b8ae1c3b38db5/botocore/handlers.py#L1080-L1089

Fixes #378

### DIFF
--- a/amazonka/CHANGELOG.md
+++ b/amazonka/CHANGELOG.md
@@ -58,6 +58,8 @@ Released: **unreleased**, Compare: 2.0 RC1 (TODO: Linkify)
 [\#648](https://github.com/brendanhay/amazonka/pull/648)
 - S3 object sizes are now `Integer` instead of `Int`
 [\#649](https://github.com/brendanhay/amazonka/pull/649)
+- amazonka-rds now supports the `DestinationRegion` pseudo-parameter for cross-region requests
+[\#661](https://github.com/brendanhay/amazonka/pull/661)
 
 ### Other Changes
 
--- a/config/annexes/rds.json
+++ b/config/annexes/rds.json
@@ -1,5 +1,47 @@
 {
     "metadata": {
         "serviceAbbreviation": "RDS"
+    },
+    "shapes": {
+        "CopyDBClusterSnapshotMessage": {
+            "members": {
+                "DestinationRegion": {
+                    "shape": "String",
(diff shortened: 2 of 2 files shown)


endgame:
@shlevy If you're still around, would be great to see if this works for you. Run `scripts/generate rds` from this branch to regenerate the `amazonka-rds` library. Otherwise I think we just merge it.

endgame:
Given #379 , this extra field is probably completely useless unless you sub in a different signer by hand, so it's probably worth merging now and fixing the rest later.

brendanhay:
@endgame Makes sense, :shipit:

endgame:
@shlevy no worries, thanks for stopping by.

PR #662, opened 2021-10-02
cognito-idp: optional UserPoolAddOnsType.AdvancedSecurityMode

endgame:
If a `CreateUserPool` call is sent to Cognito with no `UserPoolAddOns`
set, Cognito will return `"UserPoolAddOns": {}`, which causes
deserialisation failure.

Closes #473.

PR #648, opened 2021-09-30
amazonka-autoscaling: AutoScalingGroupName should be optional

endgame:
Boto is lying, and the sample response does not include this field.
http://docs.aws.amazon.com/AutoScaling/latest/APIReference/API_TerminateInstanceInAutoScalingGroup.html#API_TerminateInstanceInAutoScalingGroup_Examples

Closes #342

PR #663, opened 2021-10-04
 Regenerate all services using latest definitions

brendanhay:
Contains the result of generator updates from #640, #648 ,#649, #661, and #662.


endgame (approved):
I see nothing alarming in this.

PR #666, opened 2021-10-06
Revert "cognito-idp: optional UserPoolAddOnsType.AdvancedSecurityMode"

endgame:
Reverts brendanhay/amazonka#662

In https://github.com/boto/botocore/issues/2514 , @stobrien89 asked me to double-check that AWS was generating bogus `UserPoolAddOns` in the `CreateUserResponse`. It is not, so this "fix" shouldn't exist.


endgame:
(Tested in ap-southeast-2 and us-east-1.)

PR #683, opened 2021-10-13
Remove Haskell.nix and build everything with Bazel

brendanhay:
## Highlights 

* Replaces Haskell.nix and all custom Nix code with Bazel and `rules_haskell`. System dependencies are still pulled in via `rules_nixpkgs` to keep everything hermetic.
* `nix-shell && cabal build all` is still supported/promoted for building the repo - it just no longer requires Haskell.nix to prepare the shell and is more representative (from Cabal's POV) of how building will behave for end users who install directly from Hackage at some point in time. 
*  The `index-state` has been removed from the top-level `cabal.project` to better reflect the previous point - pinning your applications or downstream projects to a particular stackage snapshot or `index-state` is a sensible thing to do, but in libraries intended to be used as widely as possible in many downstream projects, which cannot be guaranteed to pin to the same snapshot or state, it hides issues such as overly relaxed dependency bounds until an upload or install occurs - at which point a Hackage trustee is forced to make manual modifications. By omitting the `index-state` (and possibly running `cabal update + build` on CI) we can ensure these issues are surfaced.
* A Bazel remote cache in Google Storage is enabled/available to everyone by default. (Update: this has been replaced with a gRPC cache.) This is cheaper and offers speed and sharing improvements, as cached build products are considerably more granular/incremental than what we get with Nix, which describes packages - see [Remote Caching](https://docs.bazel.build/versions/main/remote-caching.html) for more information. The Cachix cache is now extraneous - unless custom overlays/expressions were to exist.
* There is now a hugo static site under `./docs`, deployed via CI, and served at https://amazonka.brendanhay.nz. Probably I'll change this to Hakyll at some point, just because Haskell - but for now, `rules_hugo` already existed.
* The `README.md` has been overhauled and a new getting started guide has been added for contributors.
* [gazelle_cabal](https://github.com/tweag/gazelle_cabal) is run as part of code generation as a post-processing step to convert each packages `.cabal` file into a corresponding `BUILD.bazel` specification - the `.cabal` file continues to be the source of truth.
* Another Haskell binary for diffing supported services vs botocore has been added, which will allow periodically regenerating all existing + new/missing services as a scheduled CI job.

## Background

I exclusively use Bazel (at work, at the grocery store, etc.) and a lot of this PR content is actually cribbed from a closed source build pipeline for amazonka, gogol, and other openapi-based code generators (which ironically also used Haskell.nix in past lives.) While Haskell.nix is an improvement by certain metrics over the previous Stack builds for amazonka, I don't enjoy separately building everything via `nix-build ` and `cabal build` repeatedly, maintaining materialised sources, tip-toeing around nix source changes, needing to use `nix-build-uncached`, all just to try and keep Nix evaluation speedy and not accidentally modify Nix inputs and be forced to rebuild the world. Every time `haskell-nix`, `hackage-nix`, or `nixpkgs` pins are updated one must endure multi-hour repeated builds both locally and on CI. There's other issues like the `shell.nix` and `cabal.project` conflating releasable packages, such as `amazonka-*` libraries, and repository-only tooling, such as the code generator binary - resulting in problems when bumping versions and the like.

Per [Incremental Rebuilds](https://www.tweag.io/blog/2018-03-15-bazel-nix):

> Nixpkgs is designed to describe packages. Any time any of the inputs to the package change, the entire package must be rebuilt, because one target = one package. In the case of Haskell packages, this could mean rebuilding as much as all of lens, all of Agda, all of any of your ... Cabal packages, at any change anywhere. Clearly, the granularity of Nixpkgs is too large for an iterative development workflow.

To be clear - everything is still Cabal first/focused - the basic documentation should be very clear that entering a `nix-shell` and typing `cabal build ...` is all that is necessary. Bazel is only required for interacting with the code generator and other tooling that isn't released to Hackage, but can be used as a complete Cabal replacement if you're working in the repository and repeatedly building lots of packages.

## Build Times

When nothing has changed in the source tree, the Nix/Cachix happy path can be a couple of minutes faster, but when any Haskell code has changed the Bazel build has to do significantly less work overall. For complete rebuilds  of all services (assuming no cache), which previously took 2-3 hours, the Bazel build is around 20-30~ minutes faster. There are still some issues with the remote cache configuration and GCP timeouts, but decent tooling exists for optimising the cache and debugging cache misses which I'll look at in future. There is also an option to farm out builds for contributors or CI using my own [Remote Execution](https://docs.bazel.build/versions/main/remote-execution.html) setup, but I have some security aspects to consider before going that route.

## TL;DR

Rather than use Nix for hermeticity/reproducibility and then use Cabal separately to retain incremental rebuilds/development - this PR introduces Bazel which merges the `nix-build` and `cabal build` distinction into a single `bazel build` workflow. I don't expect this change to affect users of the repository - only direct contributors.

### DIFF
--- a/docs/BUILD.bazel
+++ b/docs/BUILD.bazel
@@ -35,7 +35,6 @@ haskell_doc(
     # if they don't have the expected directory prefix. This could be made more
     # robust by using `rules_pkg/mappings.bzl:pkg_files` to assemble the directory.
     name = "static",
-    tags = ["ci"],
     visibility = ["//visibility:public"],
     # deps is auto-populated by the buildozer command in scripts/generate.
     deps = [


brendanhay:
Also to note, there's 142~ new services to generate and add to the existing 140 - so total (re)build times are going to double shortly.

brendanhay:
The `docs` and `generate` CI workflow triggers need to be refined.

PR #686, opened 2021-10-16
Restructure directory layout

### LINKED ISSUE #685: Feedback on directory layout restructuring proposal
Currently, we have the following directory layout:

```
amazonka
amazonka-<service1>
amazonka-<service2>
...
docs
examples
test
gen
```

With another 140~ libraries to be added soon ™️, just viewing the top-level directory is already overwhelming, let alone with 280+ `amazonka-*` libraries present. As one potential proposal, we could introduce a new top-level `lib` (or similar) directory to nest the "release" libraries under, such as:

```
lib/amazonka
lib/amazonka-<service1>
lib/amazonka-<service2>
...
docs
examples
test
gen
```

This would also remove the need to use Bazel tags (see the `tags = ["ci"]` attributes in build files) to filter the top-level, and instead have a `lib/BUILD.bazel` which would allow expressing "release" release targets/automation via `bazel build //lib` or `//lib:automate`, for example.

One reason to not do this is the effect on downstream project's `cabal.project` `source-repository-package:subdirs`:

```
# cabal.project
source-repository-package:
    type: git
    location: https://github.com/brendanhay/amazonka
    tag: 3d274c14ca3077c3a081ba7ad57c5182da65c8c1
    subdir: lib/amazonka
...
```

But I suppose the fact you need to update that anyway to point to 2.0, makes it a fairly weak argument.

At any rate - interested to hear other's opinions.

brendanhay:
Overview: https://github.com/brendanhay/amazonka/tree/bh/directory-layout

Notes:

* Haskell libraries all under `lib`.
* Haskell library directories now match Cabal name, ie. `test` -> `lib/amazonka-test`.
* Generated libraries under `lib/services` (naming consistent with `configs/services`.)
* `stackage-snapshot.json` -> `third_party`. (Generated via `bin/pin-stackage`.)

Closes #685


brendanhay:
Hmm, I already have restructuring remorse at moving `examples` specifically under `lib`, seems quite a bit less obvious.

endgame:
In README:

> Basic examples for using the servicy libraries.

Typo: servicy



> Hmm, I already have restructuring remorse at moving `examples` specifically under `lib`, seems quite a bit less obvious.

How much manual build config is required if you put it at the repo root?

brendanhay:
> Typo: servicy

My Slav-ness coming through.

> How much manual build config is required if you put it at the repo root?

None, for examples - moving it now.

PR #688, opened 2021-10-18
2.0 version number

endgame:
Bump version numbers on things. Will need to regenerate services after this is merged, but you were planning on doing that anyway.

`scripts/generate` fails for a few things:

* Fetching botocore failed because the hash that `scripts/update-botocore` put into the `WORKSPACE` doesn't match what bazel computes at build time.
* There were a few places where `scripts/generate` needed to be told about the new layout - I fixed the obvious ones.
* After running `scripts/generate s3`, `gazelle` failed with a `cabalscan` error: 

```
gazelle: cabalscan: MissingModuleFile {modulePath = "Network/AWS/S3/Internal", cabalFile = "/home/endgame/code/amazonka/tmp.uCGonKx7Tx/amazonka-s3/amazonka-s3.cabal", componentName = "amazonka-s3"}
```

### DIFF
--- a/scripts/generate
+++ b/scripts/generate
@@ -34,10 +34,10 @@ format_and_replace() {
       --ghc-opt='-XTypeApplications'
 
   # Delete corresponding amazonka-*/gen dirs from the source tree.
-  rm -rf "./$(basename "$dir")/gen"
+  rm -rf "./lib/services/$(basename "$dir")/gen"
 
   # Copy formatted sources from tmp to the source tree.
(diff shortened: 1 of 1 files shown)


brendanhay:
Does `/home/endgame/code/amazonka/tmp.uCGonKx7Tx` exist? It looks like `cabalscan` is descending into an extraneous directory

brendanhay:
Going to merge so I can run through some gazelle/cabalscan scenarios and potentially regenerate.

PR #699, opened 2021-10-22
gen: handle suffixes with numbers in them again

### LINKED ISSUE #698: No instance for (AWSPager ListObjectsV)
I think I saw that ListObjectsV is being renamed to ListObjectsV2, but I also noticed it doesn't support pagination.

endgame:
This is probably the best compromise - we can't `takeWhile isAlphaNum`
because that captures many operations that have datestamps in them.

Fixes #698 - Once ListObjectsV2 gets the correct name again, it gets its `AWSPager` instance back.


endgame:
There were other operations with silly names so I fixed them for EC2 and S3 also. @brendanhay we'll need a regenerate.

@dustin If you want to test before services get regenerated, you'll need to pull this branch, enter a nix shell, and run `scripts/generate s3`.

PR #701, opened 2021-10-31
Generate new services

brendanhay:
Adds 135 new services and a sprinkling of generator/amazonka fixes to get the new libraries compiling. There are a handful of services that were omitted, for the noted reasons.

Requires generator (and possibly amazonka) support for `s3v4` signer:

- s3control

Generator mysteriously fails to add any sort of lens prefix to fields:

- kendra

Generator fails to correctly deal with recursive types:

- lexv2-models
- lexv2-runtime
- timestream-query
- timestream-write
- wafv2

These are all fixable but in the interests of keeping things moving I'll work on them separately as to not block the release.


endgame:
I read the commits that looked manual and weren't just service config changes. Everything seemed good, so let's ship it if CI passes.

Did you want one-liner service descriptions for the changelog entries? If so, I'm happy to look up some of them.

brendanhay:
If you feel like adding the descriptions - go ahead, I'm not too concerned personally, given there's so many. Maybe a "see AWS documentation" link to the relevant AWS product page for the service is sufficient.

endgame:
I'll do it tomorrow in a separate PR - this one is unwieldy.

configs/annexes/amplifybackend.json
```diff
@@ -0,0 +1,5 @@
+{
```
endgame, line 1: Should this be called amplifyadminui? https://docs.aws.amazon.com/amplify-admin-ui/latest/APIReference/what-is-admin-ui.html

endgame: Nah, it's called amplifybackend in the aws cli.

brendanhay: Some of the older services are named deliberately to try and make the AWS product vs documentation vs actual service name more sensible, but It doesn't scale - I think just rolling with the botocore name (ie. what ends up in the CLI) is "the one true way", even if a little nonsensical sometimes.

endgame: Yep, I agree with that.

endgame:
> I'll do it tomorrow in a separate PR - this one is unwieldy.

Update: I have about half of this done on a separate branch, and I'll keep grinding through them tonight.

endgame:
What's the difference between amazonka-sms-voice and amazonka-pinpoint-sms-voice? The generated code appears to be the same but I can't find `sms-voice` in my AWS CLI.

Judging from https://github.com/boto/botocore/commit/4db78579f72216287d99b6c646db8efeb0104aac , we should probably suppress generation of sms-voice?

brendanhay:
I'll drop the config/omit it in another commit.

endgame:
So you've removed the duplicated service, CI is passing and we have all the service descriptions on the branch. I know we have to update the release date in changelog, etc., but are there any non-admin tasks left to do?

brendanhay:
Nope, kicked off CI and went to bed last night before it had completed. Will merge now.

PR #702, opened 2021-10-31
Implement FromWebIdentity auth method.

### LINKED ISSUE #603: amazonka should use AWS_DEFAULT_REGION and not AWS_REGION
AWS_REGION used to potentially be a thing before but now the AWS CLI and SDKs use AWS_DEFAULT_REGION.

https://github.com/aws/aws-cli/issues/735

K0Te:
Closes #603


lib/amazonka/src/Network/AWS/Auth.hs
```diff
@@ -112,10 +122,28 @@ envProfile ::
 envProfile = "AWS_PROFILE"
 
 -- | Default region environment variable
-envRegion ::
+envRegionVar ::
```
endgame, line 125: Why did this change? I can't see a name clash that it's avoiding, and it may break downstream clients. Also, none of the other constants naming environment variables are using a `Var` suffix. I think it should change back.

scripts/generate
```diff
@@ -95,6 +95,14 @@ generator=(
   --assets="configs/assets"
   --retry="${botocore}/_retry.json"
 )
+say "MODEL"
+say "${models[*]}"
+if [[ " ${models[*]} " =~ " sts " ]]; then
```
endgame, line 100: Does this mean that if someone tries to regenerate only STS, it will update the copy inside amazonka instead?

endgame (commented):
Thank you for this. I am cautious, because duplicated generated code gives me the heebie-jeebies, and it feels like a bit of a hack. Also: we'll need to do similar stuff if we ever want to support `sts:AssumeRole` inside amazonka itself; will we end up pulling in a copy of 80% of the STS bindings? That might not be ideal.

I'd like to see a solution for making unsigned HTTP requests that doesn't duplicate so much code. Is that going to be awkward?

All that said, maybe we want to ship something that gets people unstuck, and fix auth properly in the 2.1 release? Happy to defer to @brendanhay here.

@K0Te Can you please also make a changelog entry for this PR?

lib/amazonka/src/Network/AWS/Auth.hs
K0Te: There was some code that used `envRegion` and lead to a warning, but I've removed it before creating a PR.
Reverted.
Also I've exported new env variables, same as `envRegion`.

scripts/generate
K0Te: In this implementation `sts` updates both embedded STS in amazonka and real `amazonka-sts`.
I've thought about creating extra `sts-embedded` model, but that would be awkward and lead to config copy-paste.

lib/amazonka/amazonka.cabal
```diff
@@ -112,7 +112,15 @@ library
...
     Network.AWS.Error
+    Network.AWS.HTTPUnsigned
     Network.AWS.Pager
+    Network.AWS.STS
+    Network.AWS.STS.AssumeRoleWithWebIdentity
+    Network.AWS.STS.Lens
+    Network.AWS.STS.Types
+    Network.AWS.STS.Types.AssumedRoleUser
+    Network.AWS.STS.Types.PolicyDescriptorType
+    Network.AWS.STS.Waiters
```
endgame, line 123: Perhaps these should go into `other-modules`, so we don't force people using `amazonka-sts` to package-qualify their imports?

scripts/generate
endgame: Could you please add a comment to this section?

lib/amazonka/amazonka.cabal
K0Te: They are already in `other-modules`, I didn't add any new exposed modules.

scripts/generate
K0Te: Sure, done.
Thank you for a prompt review!

lib/amazonka/amazonka.cabal
endgame: OK, thanks for checking.

K0Te:
Hi @endgame ,

Should I rebase this PR over #708 and remove `HTTPUnsigned.hs` from this PR?
I see that #708 is approved, but CI is failing.

PR #704, opened 2021-11-03
Add descriptions for new service bindings

endgame:
Everything new except `amazonka-sms-voice`, which appears to be a copy of `amazonka-pinpoint-sms-voice` and we should probably not gen that service at all.

Links go to product overview pages where possible, or high-level API Reference or Developer Guide pages where they bind some subfeature of a larger AWS product


brendanhay:
Very nice, thanks - I'll remove `amazonka-sms-voice` from the other PR.

brendanhay (approved):

PR #707, opened 2021-11-06
Unsigned requests

endgame:
Add the ability to make unsigned requests. Tested by performing `s3:GetObject` against a bucket whose policy allowed access from `"Principal": "*"`.

Ping @K0Te , since this may be a better way to handle the `sts:AssumeRoleWithWebIdentity` stuff. It also lets us work towards an `amazonka`/`amazonka-auth` split down the line.

### DIFF
--- a/lib/amazonka/src/Amazonka.hs
+++ b/lib/amazonka/src/Amazonka.hs
@@ -528,7 +528,7 @@ presign ::
 presign env time expires rq =
   Presign.presignWith
     (appEndo (getDual (Env.envOverride env)))
-    (Env.envAuth env)
+    (fromMaybe (error "Amazonka.presign: Env with no Auth") $ Env.envAuth env)
     (Env.envRegion env)
     time
     expires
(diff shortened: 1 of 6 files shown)


lib/amazonka/src/Amazonka.hs
```diff
@@ -528,7 +528,7 @@ presign ::
 presign env time expires rq =
   Presign.presignWith
     (appEndo (getDual (Env.envOverride env)))
-    (Env.envAuth env)
+    (fromMaybe (error "Amazonka.presign: Env with no Auth") $ Env.envAuth env)
```
endgame, line 531: Possibly not the best way to do this? But `Env`s with no credentials seem like they'll be incredibly rare...

Open to suggestions, including returning a `Maybe`, using a type-level tracker of whether we have auth info, or something else entirely.

brendanhay:
Oops 🤦🏽‍♂️

brendanhay:
Kinda annoying it closed the branch when I merged, but understandable I guess - apologies, can you reopen on `main`?

endgame:
Coulda sworn that GH would rebase PRs but anyway it's replicated at #708

PR #708, opened 2021-11-06
Add the ability to make unsigned requests

### LINKED ISSUE #602: Automatic handling of AWS_ROLE_ARN and AWS_WEB_IDENTITY_TOKEN_FILE
Some time in 2019, AWS introduced a way for service account in Kubernetes (EKS) to assume IAM roles. The main mechanism that this is achieved by is by changing pod definitions to inject AWS_ROLE_ARN and AWS_WEB_IDENTITY_TOKEN_FILE environment variables into the pod: https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/ . Roughly, this allows to exchange some OpenID Connect token for a set of AWS IAM credentials.

When the pod starts up, SDKs ought to detect these variables if present and issue the `AssumeRoleWithWebIdentity` request. The Java, Java2, Go, Python, Node, Python, Ruby, PHP and .NET SDKs have been updated back in 2019. Here's a simple example of where in the process the Go SDK does it:

https://github.com/aws/aws-sdk-go/blob/87f0aa4be0153f21470e7a9802ebfb3403faa9ce/aws/session/credentials.go#L35

This Haskell SDK should probably be updated too. Without it, Haskell applications deployed into EKS don't Just Work™ with IAM roles for service credentials and we have to add explicit code to do the switch.

endgame:
Add the ability to make unsigned requests. Tested by performing `s3:GetObject` against a bucket whose policy allowed access from `"Principal": "*"`.

Ping @K0Te , since this may be a better way to handle the `sts:AssumeRoleWithWebIdentity` stuff. It also lets us work towards an `amazonka`/`amazonka-auth` split down the line.

Closes #702 
Closes #602


lib/amazonka/src/Amazonka.hs
```diff
@@ -528,7 +528,7 @@ presign ::
 presign env time expires rq =
   Presign.presignWith
     (appEndo (getDual (Env.envOverride env)))
-    (Env.envAuth env)
+    (fromMaybe (error "Amazonka.presign: Env with no Auth") $ Env.envAuth env)
```
endgame, line 531: Possibly not the best way to do this? But `Env`s with no credentials seem like they'll be incredibly rare...

Open to suggestions, including returning a `Maybe`, using a type-level tracker of whether we have auth info, or something else entirely.

brendanhay: I'm not thrilled about making it partial - but I think doing it properly involves rewriting the `Auth` module and/or presigning, so if we really want this in 2.0 lets' just release it (with the appropriate caveats noted in Haddock comments) and plan to revisit.

brendanhay (approved):

endgame:
I don't consider this critical for the 2.0 RC. If there's nothing else we need on that front, we should get it out the door first. Maybe if things are quiet during stabilisation we do this?

The urgent question, I guess: are you happy with the `amazonka-core`/`amazonka-auth` split we talked about previously?

- Have two packages so `amazonka-auth` can depend on `amazonka-sts` without creating a dependency cycle; 
- `amazonka` re-exports from both for convenience and to avoid too much API breakage.

If so, then I think get the pieces lined up to make that possible (which shouldn't be much).

As for avoiding partiality in the API: one way would be to add some type-level tags that track whether or not an `Env` has access to credentials. If I remove `NoCredentials` from the `Credentials` type and instead provide a separate helper function, we can present a safe interface to callers. What do you think of that?

brendanhay:
The current dependency graph is:

```
                         ┌─────► amazonka-<service> ──────┐
                         │                                │
    amazonka-core ───────┤                                ├──────► application
                         │                                │
                         └─────► amazonka ────────────────┘
```

And I understand your proposal as:

```
                       ┌─────► amazonka-<service> ─────────────────────────────────────┐
                       │                                                               │
  amazonka-core ───────┤                                                               ├──────► application
                       │                                                               │
                       └─────► amazonka-sts ────► amazonka-auth ──────► amazonka ──────┘
```

I'm probably misunderstanding, but is the additional library really necessary?

Regarding `Env`, your separate helper function proposal is preferable.

endgame:
> I'm probably misunderstanding, but is the additional library really necessary?

It is not. For some reason I thought generated services depended on `amazonka`, not `amazonka-core`.

After rewriting the `Env` type to have a `withAuth` type parameter, a lot of simplifications fell out. and made `sts:AssumeRoleWithWebIdentity` possible. I did have to break a couple of import loops with `.hs-boot` files but I think that if you're using AWS calls to get AWS creds, that's going to be the price of admission.

@K0Te - Can you please test it when you get a chance?

endgame:
I do wonder if we should be parameterising the `FromWebIdentity` constructor some, so that the token doesn't _have_ to be provided via envvars.

endgame:
> I do wonder if we should be parameterising the `FromWebIdentity` constructor some, so that the token doesn't _have_ to be provided via envvars.

But OTOH people can create their own `EnvNoAuth` and make the call themselves if they don't want to use our approximation to the default credential provider chain.

brendanhay:
> For some reason I thought generated services depended on amazonka, not amazonka-core.

You were correct up until I fat-fingered 524fde6e02b4acf9aa186947de95e9354085600c on `main` recently.

> But OTOH people can create their own EnvNoAuth and make the call themselves if they don't want to use our approximation to the default credential provider chain.

That seems acceptable - It looks good to me, go ahead and merge whenever you're ready!

endgame:
Fixing a couple of documentation nits and giving @fuuzetsu some well-deserved credit also, then will merge as soon as CI passes.

endgame:
CI takes ~3h and I built this locally. Hitting the merge button now.

PR #710, opened 2021-11-08
Various bazel configuration tweaks to improve CI

brendanhay:


endgame:
Not sure there's anything useful I can add here, but let me know if I'm wrong. What happens after this is fixed & merged?

PR #711, opened 2021-11-14
Fully qualify all generator imports

brendanhay:
General tidy up of the generator prior to removing extraneous lens usage and moving from `HashMap` -> `Map` at some point in the near future.


gen/src/Gen/AST.hs
```diff
@@ -98,45 +98,49 @@ renderShapes cfg svc = do
...
     check (Just p) d n f = do
       let k = (p, d, n)
-      m <- uses _2 (Set.member k)
+
+      m <- Lens.uses Lens._2 (HashSet.member k)
+
       if m
         then pure ()
-        else _2 %= Set.insert k >> f
+        else Lens._2 %= HashSet.insert k >> f
```
endgame, line 143: How about:
```haskell
unless m $ Lens._2 %= HashSet.insert k >> f
```

gen/src/Gen/AST/Data/Field.hs
```diff
@@ -125,12 +123,12 @@ mkFields (view metadata -> m) s st =
 -- of the JSON service definition.
 sortFields :: [Id] -> [Field] -> [Field]
 sortFields xs =
-  zipWith (set fieldOrdinal) [1 ..]
+  zipWith (Lens.set fieldOrdinal) [1 ..]
     -- FIXME: optimise
-    . sortBy (on compare isStreaming)
-    . sortBy (on compare idx)
+    . List.sortBy (Function.on compare isStreaming)
+    . List.sortBy (Function.on compare idx)
```
endgame, line 129: Use `Data.Ord.comparing` instead of `Function.on compare`?

gen/src/Gen/AST/Data/Syntax.hs
```diff
@@ -1,31 +1,26 @@
 {-# OPTIONS_GHC -fno-warn-orphans #-}
 
+-- |
 -- Module      : Gen..AST.Data.Syntax
 -- Copyright   : (c) 2013-2021 Brendan Hay
--- License     : This Source Code Form is subject to the terms of
+-- License     : This Source Code Form is subject Lens.to the terms of
```
endgame, line 6: Oops

gen/src/Gen/AST/Data/Syntax.hs
```diff
@@ -430,7 +425,7 @@ responseE p r fs = Exts.app (responseF p r fs) bdy
       | fieldLit f =
         if fieldIsParam f
           then Exts.app (var "Prelude.pure") (var "x")
-          else -- Coerce is inserted here to handle newtypes such as Sensitive.
+          else -- Coerce is inserted here Lens.to handle newtypes such as Sensitive.
```
endgame, line 428: Oops

gen/src/Gen/Tree.hs
```diff
@@ -47,7 +43,7 @@ fold ::
 fold g f (p :/ t) = (p :/) <$> go p t
   where
     go x = \case
-      Failed n e -> fail (show e) >> return (Failed n e)
+      Failed n e -> fail (show e) >> pure (Failed n e)
```
endgame, line 46:
```suggestion
      Failed n e -> fail (show e)
```
Because `fail (show e) >> pure (Failed n e)` = `fail (show e) >>= \_ -> pure (Failed n e)` = `fail (show e)` by the `MonadFail` laws?

gen/src/Gen/Types/Retry.hs
```diff
@@ -1,27 +1,25 @@
 {-# LANGUAGE TemplateHaskell #-}
 
+-- |
 -- Module      : Gen.Types.Retry
 -- Copyright   : (c) 2013-2021 Brendan Hay
 -- License     : This Source Code Form is subject to the terms of
 --               the Mozilla xtPublic License, v. 2.0.
 --               A copy of the MPL can be found in the LICENSE file or
---               you can obtain it at http://mozilla.org/MPL/2.0/.
+--               you can obtain it Lens.at http://mozilla.org/MPL/2.0/.
```
endgame, line 9: Oops

gen/src/Gen/Types/Retry.hs
```diff
@@ -59,63 +57,69 @@ data Delay = Delay
...
       Retry'
-        (fromMaybe (r ^. retryAttempts) m)
-        (fromMaybe (r ^. retryDelay) d)
-        (r ^. retryPolicies <> p)
+        <$> o .: "max_attempts"
+        <*> o .: "delay"
+        <*> pure mempty
 
-parseRetry :: Text -> Object -> Parser Retry
+instance FromJSON (Retry -> Retry) where
```
endgame, line 93: I generally find `FromJSON` instances on functions very confusing, but sometimes there's no better way.

gen/src/Gen/Types/Service.hs
```diff
@@ -53,14 +44,15 @@ data Signature
   deriving (Eq, Show, Generic)
 
 sigToText :: Signature -> Text
-sigToText V2 = "v2"
-sigToText _ = "v4"
+sigToText = \case
+  V2 -> "v2"
+  _ -> "v4"
```
endgame, line 49: Is this correct? Should `V3`, `V3HTTPS` and `S3` all return `"v4"`?

endgame (commented):
So you want to get the gen using `HashMap` and `HashSet` now, so you can mass-rename the whole repo to use `Map` etc in the future?

gen/src/Gen/Types/Service.hs
brendanhay: It's probably not correct but I'll worry it about it separately.

gen/src/Gen/Types/Retry.hs
brendanhay: If this is the most confusing thing you've found in the generator - I applaud you, Sir.

endgame: TBQH I've found the generator reasonably comprehensible - when I've needed to tweak something the lever I needed seemed to be pretty easy to find.

But maybe that's because I used to understand the autotools and it's permanently damaged my baseline.

brendanhay:
> So you want to get the gen using `HashMap` and `HashSet` now, so you can mass-rename the whole repo to use `Map` etc in the future?

Yeah, independent of the wider library's reasoning for supporting `aeson > 2`. By removing the dependency on the `Gen.Types.Map`  `type Map = HashMap` alias we can migrate the generator to use `containers` everywhere, which will be a breaking change of the current hash-based ordering of generated outputs such as smart constructor parameters. 

With `aeson > 2` we can possibly use `Vector (Text, Value)` as the JSON object representation in the generator, specifically - preserving the service definition's key ordering. Or, failing that, something along the lines of `insert-ordered-containers` or `Dhall.Map`. 

Even without JSON objects that preserve the service definition ordering, the lexicographic ordering yielded by `containers` would already be an improvement. This would be a breaking change for generated code, but it will make generated smart constructor/function parameter ordering more sensible/resilient/stable, and with insert-ordered objects it would match the documentation and AWS CLI argument ordering.

gen/src/Gen/Tree.hs
brendanhay: Organically grown grass-fed code.

gen/src/Gen/AST/Data/Syntax.hs
brendanhay: I've nuked all the module headers in the generator - since it's not uploaded to Hackage, it doesn't make a lot of sense - the license is already included in the repository root.

endgame: Makes sense.

endgame:
Yep, that sounds like a pretty good program of works, especially if it stops adjacent `Text` arguments from being swapped around in smart constructors. Can we do it in parallel with the stabilisation period and get RC1 out into the world?

brendanhay:
Doing it after 2.0 makes sense, there's bit of work involved and probably the `aeson-2.*` update should happen first.

endgame:
Does that then mean it's time to regenerate docs/tag release/do the announcement, or is there something else that needs to go in first?

PR #720, opened 2021-11-29
bazel: Add GHC9 toolchains

endgame:
Cargo-culting, but I think I got this right.

We'll need this eventually, but the stackage snapshot doesn't support the base-4.15 versions of some libraries - not sure what @brendanhay wants to do about that. (Merge this later? Re-pin stackage?)

I did a full build of the repo using GHC 9.0.1 and all the tests passed. There is currently no build plan for GHC 9.2.1; at least one of our transitive dependencies ([memory](https://github.com/vincenthz/hs-memory/pull/87)) needs updating for GHC 9.2.1 and a new Hackage release.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -234,6 +234,26 @@ rules_haskell_dependencies()
 register_toolchains("//tools/ghc:toolchain")
 
 haskell_register_ghc_nixpkgs(
+    name = "ghc921",
+    attribute_path = "haskell.compiler.ghc921",
+    haddock_flags = [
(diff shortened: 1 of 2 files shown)


brendanhay (approved):
LGTM - but I'll probably need to reinstate a previous bit of toolchain shenanigans that existed to select the stackage resolver based on what compiler you want to use - ie. nightly for `9.*` and something "older" for `8.*`. (Moving forward, if `9.0.1` works that's good enough for now.)

PR #722, opened 2021-12-01
Regenerate all service NFData and Hashable instances

brendanhay:


lib/services/amazonka-certificatemanager/gen/Amazonka/CertificateManager/AddTagsToCertificate.hs
```diff
@@ -128,9 +128,14 @@ instance Core.AWSRequest AddTagsToCertificate where
   response =
     Response.receiveNull AddTagsToCertificateResponse'
 
-instance Prelude.Hashable AddTagsToCertificate
+instance Prelude.Hashable AddTagsToCertificate where
+  hashWithSalt initialSalt AddTagsToCertificate' {..} =
+    initialSalt `Prelude.hashWithSalt` tags
```
endgame, line 133: Looks like we're not hashing `certificateArn`? Is that right?

endgame (commented):
I wonder if there's an off-by-one in the way the `Hashable` instance is generated?

lib/services/amazonka-certificatemanager/gen/Amazonka/CertificateManager/AddTagsToCertificate.hs
brendanhay: It's a deliberate compilation performance optimisation! (Will fix, thanks for catching.)

endgame: `const 0` is a very fast valid hash function.

endgame:
I suggest that for types with a single constructor, we apply `()` to `hashWithSalt` instead of `0`, to avoid the ambiguity. We might also need annotations for Ints passed into hash functions for sum types.

endgame:
Closing this, since the generator's been fixed in another PR and I'm pretty sure I've got everything useful from here. Reopen and prod me if that's not the case.

PR #723, opened 2021-12-01
amazonka-route53: correct S3 Website Hosted Zone IDs

endgame:
I noticed in my travels that we didn't have all the zone IDs. Perhaps AWS added them after this function was written?


brendanhay (approved):

brendanhay:
Yes, they've been added and I never revisited.

PR #721, opened 2021-11-29
amazonka-gen: Synthesise NFData and Hashable implementations

brendanhay:
Rather than relying on the default `Generic` implementation for the `NFData` and `Hashable` classes, generate each respective function implementation to mitigate some compilation overhead.

Example output:

```haskell
instance Prelude.Hashable CreateBucket where
  hash CreateBucket' {..} =
    Prelude.hash grantReadACP
      `Prelude.hashWithSalt` bucket
      `Prelude.hashWithSalt` acl
      `Prelude.hashWithSalt` grantWrite
      `Prelude.hashWithSalt` createBucketConfiguration
      `Prelude.hashWithSalt` grantFullControl
      `Prelude.hashWithSalt` grantRead
      `Prelude.hashWithSalt` grantWriteACP
      `Prelude.hashWithSalt` objectLockEnabledForBucket

instance Prelude.NFData CreateBucket where
  rnf CreateBucket' {..} =
    Prelude.rnf grantReadACP
      `Prelude.seq` Prelude.rnf bucket
      `Prelude.seq` Prelude.rnf acl
      `Prelude.seq` Prelude.rnf grantWrite
      `Prelude.seq` Prelude.rnf createBucketConfiguration
      `Prelude.seq` Prelude.rnf grantFullControl
      `Prelude.seq` Prelude.rnf grantRead
      `Prelude.seq` Prelude.rnf grantWriteACP
      `Prelude.seq` Prelude.rnf objectLockEnabledForBucket
```

And for a datatype with cardinality 1:

```haskell
instance Prelude.Hashable ContinuationEvent where
  hash _ = 0

instance Prelude.NFData ContinuationEvent where
  rnf _ = ()
```

Closes #719

PR #724, opened 2021-12-01
amazonka-dynamodb: Use a sum type for AttributeValue

endgame:
Note:

* I haven't actually used this in anger yet. Putting it up now for discussion and possibly to test work code against once we get across to `amazonka-2.0` RC1 (next week, most likely).
* Branch is currently based off #723 so I can write a decent changelog entry. Merge that first and then this will look sensible.
* I had to temporarily revert #721 while developing this because the generated instances failed to compile with "Ambiguous Type" errors.
* I have tested building using `--constraints 'aeson < 2.0'` as well as `--constraints 'aeson >= 2.0'`.
* Need to `scripts/generate dynamodb` before testing, obviously.

Open questions:

* Do we actually want this? I feel that it's OK for amazonka to do this because it hews closely to the data formats, but a more advanced/opinionated mapping library would be out of scope for this project.
* Should we use the pattern synonym/`newtype` trick here in case AWS add more data types to dynamo? Feels a bit unlikely.
* I believe @brendanhay has a long-term plan to remove the direct `lens` dependency - I'm prepared to write the prisms manually if necessary.

### DIFF
--- a/configs/services/dynamodb.json
+++ b/configs/services/dynamodb.json
@@ -23,6 +23,11 @@
                 "Responses",
                 "UnprocessedKeys"
             ]
+        },
+        "BatchWriteItemOutput": {
+            "requiredFields": [
+                "UnprocessedItems"
+            ]
         }
     }
 }
--- a/configs/services/dynamodb.json
+++ b/configs/services/dynamodb.json
@@ -4,7 +4,6 @@
         "aeson",
         "containers",
         "hashable >=1.3.4.0 && <1.5",
-        "lens",
         "unordered-containers",
         "vector"
     ],
--- a/configs/services/dynamodb.json
+++ b/configs/services/dynamodb.json
@@ -1,3 +1,22 @@
 {
-    "libraryName": "amazonka-dynamodb"
+    "libraryName": "amazonka-dynamodb",
+    "extraDependencies": [
+        "aeson",
+        "containers",
+        "hashable >=1.3.4.0 && <1.5",
+        "lens",
+        "unordered-containers",
+        "vector"
+    ],
+    "typeModules": [
+        "Amazonka.DynamoDB.Internal"
+    ],
+    "typeOverrides": {
+        "AttributeValue": {
+            "replacedBy": {
+                "name": "AttributeValue",
+                "underive": []
+            }
+        }
+    }
 }
--- a/configs/services/dynamodb.json
+++ b/configs/services/dynamodb.json
@@ -28,6 +28,11 @@
             "requiredFields": [
                 "UnprocessedItems"
             ]
+        },
+        "GetItemOutput": {
+            "requiredFields": [
+                "Item"
+            ]
         }
     }
 }
--- a/configs/services/dynamodb.json
+++ b/configs/services/dynamodb.json
@@ -17,6 +17,12 @@
                 "name": "AttributeValue",
                 "underive": []
             }
+        },
+        "BatchGetItemOutput": {
+            "requiredFields": [
+                "Responses",
+                "UnprocessedKeys"
+            ]
         }
     }
 }
--- a/configs/services/dynamodb.json
+++ b/configs/services/dynamodb.json
@@ -33,6 +33,11 @@
             "requiredFields": [
                 "Item"
             ]
+        },
+        "QueryOutput": {
+            "requiredFields": [
+                "Items"
+            ]
         }
     }
 }
--- a/lib/services/amazonka-dynamodb-streams/src/Amazonka/DynamoDBStreams/Internal.hs
+++ b/lib/services/amazonka-dynamodb-streams/src/Amazonka/DynamoDBStreams/Internal.hs
@@ -5,7 +5,6 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE StrictData #-}
-{-# LANGUAGE TemplateHaskell #-}
 {-# OPTIONS_GHC -Wall -Werror #-}
 
 -- |
@@ -15,27 +14,10 @@
 -- Maintainer  : Brendan Hay <brendan.g.hay@gmail.com>
 -- Stability   : experimental
 -- Portability : non-portable (GHC extensions)
-module Amazonka.DynamoDBStreams.Internal
-  ( -- * Attribute Values
-    AttributeValue (..),
-
-    -- ** Attribute Value Prisms
-    _L,
-    _NS,
-    _M,
-    _NULL,
-    _N,
-    _BS,
-    _B,
-    _SS,
-    _S,
-    _BOOL,
-  )
-where
+module Amazonka.DynamoDBStreams.Internal where
 
 import Amazonka.Core
 import Amazonka.Prelude
-import Control.Lens.TH (makePrisms)
 import Data.Aeson (pairs)
 import Data.Hashable
 import Data.Map (Map)
@@ -107,8 +89,6 @@ data AttributeValue
     BOOL Bool
   deriving stock (Eq, Read, Show, Generic)
   deriving anyclass (NFData)
-
-$(makePrisms ''AttributeValue)
 
 instance Hashable AttributeValue where
   hashWithSalt salt = \case
--- a/lib/services/amazonka-dynamodb-streams/src/Amazonka/DynamoDBStreams/Internal.hs
+++ b/lib/services/amazonka-dynamodb-streams/src/Amazonka/DynamoDBStreams/Internal.hs
@@ -1,0 +1,169 @@
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE StrictData #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# OPTIONS_GHC -Wall -Werror #-}
+
+-- |
+-- Module      : Amazonka.DynamoDBStreams.Internal
+-- Copyright   : (c) 2013-2021 Brendan Hay
+-- License     : Mozilla Public License, v. 2.0.
+-- Maintainer  : Brendan Hay <brendan.g.hay@gmail.com>
+-- Stability   : experimental
+-- Portability : non-portable (GHC extensions)
+module Amazonka.DynamoDBStreams.Internal
+  ( -- * Attribute Values
+    AttributeValue (..),
+
+    -- ** Attribute Value Prisms
+    _L,
+    _NS,
+    _M,
+    _NULL,
+    _N,
+    _BS,
+    _B,
+    _SS,
+    _S,
+    _BOOL,
+  )
+where
+
+import Amazonka.Core
+import Amazonka.Prelude
+import Control.Lens.TH (makePrisms)
+import Data.Aeson (pairs)
+import Data.Hashable
+import Data.Map (Map)
+import Data.Vector (Vector)
+
+#if MIN_VERSION_aeson(2,0,0)
+import qualified  Data.Aeson.KeyMap as KeyMap
+#else
+import qualified  Data.HashMap.Strict as KeyMap
+#endif
+
+-- | Represents the data for an attribute.
+--
+-- DynamoDB sends and receives JSON objects which contain a single
+-- item whose key is a data type and the value is the data itself. We
+-- provide an actual sum type to interact with these.
+--
+-- For more information, see
+-- <https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html#HowItWorks.DataTypes Data Types>
+-- in the /Amazon DynamoDB Developer Guide/.
+data AttributeValue
+  = -- | An attribute of type List. For example:
+    --
+    -- @\"L\": [ {\"S\": \"Cookies\"} , {\"S\": \"Coffee\"}, {\"N\", \"3.14159\"}]@
+    L (Vector AttributeValue)
+  | -- | An attribute of type Number Set. For example:
+    --
+    -- @\"NS\": [\"42.2\", \"-19\", \"7.5\", \"3.14\"]@
+    --
+    -- Numbers are sent across the network to DynamoDB as strings, to maximize
+    -- compatibility across languages and libraries. However, DynamoDB treats
+    -- them as number type attributes for mathematical operations.
+    NS (Vector Text)
+  | -- | An attribute of type Map. For example:
+    --
+    -- @\"M\": {\"Name\": {\"S\": \"Joe\"}, \"Age\": {\"N\": \"35\"}}@
+    M (Map Text AttributeValue)
+  | -- | An attribute of type Null. For example:
+    --
+    -- @\"NULL\": true@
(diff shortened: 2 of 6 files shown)


endgame:
CC @lrworth @unisay @axman6

lib/services/amazonka-dynamodb/src/Amazonka/DynamoDB/Internal.hs
```diff
@@ -0,0 +1,139 @@
...
+-- provide an actual sum type to interact with these.
+--
+-- For more information, see
+-- <https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html#HowItWorks.DataTypes Data Types>
+-- in the /Amazon DynamoDB Developer Guide/.
+data AttributeValue
+  = -- | An attribute of type List. For example:
+    --
+    -- @\"L\": [ {\"S\": \"Cookies\"} , {\"S\": \"Coffee\"}, {\"N\", \"3.14159\"}]@
+    L [AttributeValue]
```
axman6-da, line 59: Is there a good reason for this to be a list and not a vector? I assume it comes out of the aeson `Value` as a `Vector`

lib/services/amazonka-dynamodb/src/Amazonka/DynamoDB/Internal.hs
```diff
@@ -0,0 +1,139 @@
...
+    -- @\"NULL\": true@
+    NULL
+  | -- | An attribute of type Number. For example:
+    --
+    -- @\"N\": \"123.45\"@
+    --
+    -- Numbers are sent across the network to DynamoDB as strings, to maximize
+    -- compatibility across languages and libraries. However, DynamoDB treats
+    -- them as number type attributes for mathematical operations.
+    N Text
```
axman6-da, line 83: I would unpack this, saves a word and an indirection per text.

lib/services/amazonka-dynamodb/src/Amazonka/DynamoDB/Internal.hs
```diff
@@ -0,0 +1,139 @@
...
+  | -- | An attribute of type Binary Set. For example:
+    --
+    -- @\"BS\": [\"U3Vubnk=\", \"UmFpbnk=\", \"U25vd3k=\"]@
+    BS [Base64]
```
axman6-da, line 87: Same as above

lib/services/amazonka-dynamodb/src/Amazonka/DynamoDB/Internal.hs
```diff
@@ -0,0 +1,139 @@
...
+    -- @\"SS\": [\"Giraffe\", \"Hippo\" ,\"Zebra\"]@
+    SS [Text]
+  | -- | An attribute of type String. For example:
+    --
+    -- @\"S\": \"Hello\"@
+    S Text
+  | -- | An attribute of type Boolean. For example:
+    --
+    -- @\"BOOL\": true@
+    BOOL Bool
```
axman6-da, line 103: I wonder if there's any benefit to having just BOOLTrue and BOOLFalse (and possibly a pattern synonym for BOOL Bool). Basically encodes the bool as one bit, which even unpacking won't give us.

lib/services/amazonka-route53/src/Amazonka/Route53/Internal.hs
```diff
@@ -64,8 +64,8 @@ instance FromXML ResourceId where
 getHostedZoneId :: Region -> Maybe ResourceId
 getHostedZoneId = \case
     Ohio            -> Just "Z2O1EMRO9K5GLX"
-    NorthVirginia   -> Just "Z3AQBSTGFYJSTF "
-    NorthCalifornia -> Just "Z2F56UZL2M1ACD "
```
axman6-da, line 68: Is this change related to the above change?

lib/services/amazonka-dynamodb/src/Amazonka/DynamoDB/Internal.hs
```diff
@@ -0,0 +1,139 @@
...
+      [("S", v)] -> S <$> parseJSON v
+      [("BOOL", v)] -> B <$> parseJSON v
+      [] -> fail "No keys"
+      _ -> fail $ "Multiple or unrecognized keys: " ++ show (KeyMap.keys o)
+
+instance ToJSON AttributeValue where
+  toJSON = object . pure . toKeyValue
+  toEncoding = pairs . toKeyValue
+
+toKeyValue :: KeyValue kv => AttributeValue -> kv
```
axman6-da, line 128: Hopefully this will be specialised and/or inlined.

axman6-da (commented):
This looks great, my only comments relate to the in memory representation. I can't see much point in not using the most space efficient representations we can for a package like this. I'm sure most heavy users of the package would appreciate not having a large increase in memory size compared to the underlying JSON. I would be very surprised if laziness in the values is particularly useful here. Happy to discuss, but this is how I would design it for better performance/memory efficiency.

lib/services/amazonka-route53/src/Amazonka/Route53/Internal.hs
endgame: No:

> Branch is currently based off #723 so I can write a decent changelog entry. Merge that first and then this will look sensible.

lib/services/amazonka-dynamodb/src/Amazonka/DynamoDB/Internal.hs
endgame: Got any suggestions to turn that "hopefully" into a "definitely"? I suppose I could just repeat myself.

lib/services/amazonka-dynamodb/src/Amazonka/DynamoDB/Internal.hs
endgame: Not to my knowledge, I just reached for the thing that took the least typing.

lib/services/amazonka-dynamodb/src/Amazonka/DynamoDB/Internal.hs
endgame: My gut is that it doesn't outweigh the costs, and you're going to assemble a `Bool` almost immediately (either through the pattern synonym or a prism).

lib/services/amazonka-dynamodb/src/Amazonka/DynamoDB/Internal.hs
axman6: INLINE should be enough

lib/services/amazonka-dynamodb/src/Amazonka/DynamoDB/Internal.hs
axman6: True.

lib/services/amazonka-dynamodb/src/Amazonka/DynamoDB/Internal.hs
axman6-da: aeson has gone through the hard work of figuring out the length of this and giving us constant time access, I'd hate to lose that or have to re-do that. Anyone can use toList if they want to.

lib/services/amazonka-dynamodb/src/Amazonka/DynamoDB/Internal.hs
```diff
@@ -0,0 +1,139 @@
...
+    -- @\"NS\": [\"42.2\", \"-19\", \"7.5\", \"3.14\"]@
+    --
+    -- Numbers are sent across the network to DynamoDB as strings, to maximize
+    -- compatibility across languages and libraries. However, DynamoDB treats
+    -- them as number type attributes for mathematical operations.
+    NS [Text]
+  | -- | An attribute of type Map. For example:
+    --
+    -- @\"M\": {\"Name\": {\"S\": \"Joe\"}, \"Age\": {\"N\": \"35\"}}@
+    M (HashMap Text AttributeValue)
```
axman6-da, line 71: This might also be a bit controversial (see https://github.com/haskell/aeson/blob/master/changelog.md#2000), though I guess the sorts of issues aeson is attempting to fix around HashMap at the moment aren't such an issue here (unless you expect users of the library to accept arbitrary AttributeValues from users of their service, which aren't coming from DynamoDB)

endgame: The thing with aeson is that it's often used to decode request payloads and stuff like that, which means the attacker has control over what you try and deserialise. This seems much harder to exploit - the dodgy map would (most often) have to be read out of a database and the items size limits on dynamodb bound just how bad things can get.

That said, I've had surprising performance problems with `HashMap` in the past; perhaps we just go with `Map` here?

axman6-da: Yeah, or since we're depending on aeson anyway, use its keycap type perhaps... not quite sure what the implications of that would be but it does shift the work of doing the right thing to another package, that people are familiar with.

On an unrelated note, are there restrictions on what the values in map keys can look like?

endgame: Aeson's keymap type is provided only to smooth the transition from `HashMap` to `Map` as internal representation. We're doing a breaking change anyway and may as well move to `Map`?

Restrictions: the docs say "Attribute names must be at least one character long, but no greater than 64 KB long." -- I expect this to apply to maps, and there's nothing else that claims to constrain them. Values are AttributeValues and can be nested to I think arbitrary depth.

lib/services/amazonka-dynamodb/src/Amazonka/DynamoDB/Internal.hs
endgame: Couldn't get it to inline even with `{-# SPECIALIZE INLINE #-}`, inlining by hand instead.

lib/services/amazonka-dynamodb/src/Amazonka/DynamoDB/Internal.hs
endgame: Thought hard about this, but as soon as you use it anywhere else it'll need to be reboxed, and it's quite likely to get pulled out of the AST and sent elsewhere.

lib/services/amazonka-dynamodb/src/Amazonka/DynamoDB/Internal.hs
endgame: OK

configs/services/dynamodb.json
```diff
@@ -1,3 +1,21 @@
 {
-    "libraryName": "amazonka-dynamodb"
+    "libraryName": "amazonka-dynamodb",
+    "extraDependencies": [
+        "aeson",
+        "containers",
+        "hashable >=1.3.4.0 && <1.5",
```
endgame, line 6: We need `hashable >= 1.3.4.0` to get `instance (Hashable k, Hashable v) => Hashable (Map k v)`. I am happy to raise the bound on `amazonka-core` instead if you prefer.

endgame:
Oh, and we'll need to repin stackage after merge.

lib/services/amazonka-dynamodb/src/Amazonka/DynamoDB/Internal.hs
```diff
@@ -0,0 +1,167 @@
...
+    SS (Vector Text)
+  | -- | An attribute of type String. For example:
+    --
+    -- @\"S\": \"Hello\"@
+    S Text
+  | -- | An attribute of type Boolean. For example:
+    --
+    -- @\"BOOL\": true@
+    BOOL Bool
+  deriving (Eq, Read, Show, Generic, NFData)
```
brendanhay, line 107: Please use `DerivingStrategies` and differentiate the `stock` vs `anyclass` instances

brendanhay (approved):
See https://github.com/brendanhay/amazonka/tree/bh/dynamodb-mapper/amazonka-dynamodb-extras/src/Amazonka/DynamoDB for a blast from the past. It was mostly just random musings around what a layer above `amazonka-dynamodb` might look like.

configs/services/dynamodb.json
brendanhay: I'm fine with the constraint being here

lib/services/amazonka-dynamodb/src/Amazonka/DynamoDB/Internal.hs
endgame: OK, will do. I always think of it as a "new extension" and forget that this has been around since 8.2.1.

endgame:
I had a look at your mapper; we have something a bit similar in an internal codebase. The main problem I have with the typeclass-driven approaches is that I often don't want to read/write an entire item - many of our operations are doing partial `SET`s, or projecting out a couple of fields, or whatever.

For this reason, I prefer building up encoder/decoders by combining values. If you keep an `Encoder a` and `Decoder a` next to each other, you can make a profunctor `P a a` and use the tools in `product-profunctors` to build up a codec for larger structures. Then you can break off the encoder or decoder as-needed. But I need to get around to finishing the big refactor of that package...

*****

Anyway. I'll build up an internal test branch for work (with this, regen `amazonka-dynamodb`, etc) to make sure this type does what we need here. That should happen early-mid next week, after which I can merge this.

brendanhay:
The branch is close to 6 years old - I likely now wouldn't use type classes either, as it currently stands. I just figured it might be of interest in regards to other functionality.

endgame:
Oh BTW @brendanhay do you want hand-written prisms instead of TH-generated ones?

endgame:
Discovered that DynamoDB Streams has the same type. Not sure if there's a better way to do this.

endgame:
> Oh, and we'll need to repin stackage after merge.

And, annoyingly, stackage is on hashable-1.3.0.0 for GHC 8.10.7, which doesn't have `instance (Hashable k, Hashable v) => Hashable (Map k v)`. Should we `hashUsing (Map.toList)`?

axman6 (approved):
🎉

lib/services/amazonka-dynamodb-streams/src/Amazonka/DynamoDBStreams/Internal.hs
```diff
@@ -0,0 +1,169 @@
...
+    -- @\"S\": \"Hello\"@
+    S Text
+  | -- | An attribute of type Boolean. For example:
+    --
+    -- @\"BOOL\": true@
+    BOOL Bool
+  deriving stock (Eq, Read, Show, Generic)
+  deriving anyclass (NFData)
+
+$(makePrisms ''AttributeValue)
```
endgame, line 111: Perhaps we should rely on `generic-lens`/`generic-optics`, as we do for everything else?

endgame: Yes, I think we should do that.

endgame:
Status update: generic-lens has a good story for prisms, and this branch makes using `AttributeValue` so much more pleasant.

endgame:
Gonna merge this now, but I've filed https://github.com/commercialhaskell/stackage/issues/6386 to see how likely it we'll see Stackage LTS with GHC 8.10.x and `hashable >=1.3.4.0`. If it's not gonna happen, I'll PR a compatibility fix.

endgame:
Since amazonka-2.0 (whenever we get it out) won't make it into existing stackage LTS, stack users will have to extra-deps their way to victory. I'm comfortable with forcing them to add `hashable` to their `extra-deps:`.

PR #725, opened 2021-12-02
S3 mandatory multipart upload ids

endgame:
As before, this is forked from #723 so I have a changelog to work in.

### DIFF
--- a/configs/services/glacier.json
+++ b/configs/services/glacier.json
@@ -11,6 +11,11 @@
                 "partSize"
             ]
         },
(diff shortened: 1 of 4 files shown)


endgame:
@axman6 FYI

endgame:
Reported upstream: https://github.com/boto/botocore/issues/2572

> S3's stance [the AWS team maintaining the service] on seemingly required output fields is to mark them as optional, as output is something the user cannot control. I did inquire about changing them since it seemed to cause some confusion, but they declined to do so.

PR #738, opened 2021-12-07
Env lenses

endgame:


endgame:
It's probable that `generics-lens` is smarter than I thought (and the problem I was fighting was actually PEBCAK), but it always felt weird that we had all sorts of useful `Env` combinators _except_ field lenses in `Amazonka`/`Amazonka.Env`. Perhaps we soften the language in the haddocks and keep them anyway?

PR #741, opened 2021-12-21
SSM: Require more fields

### LINKED ISSUE #740: ssm: Is the Value field on Parameter required?
I am a bit unsure about the Value field on the Parameter type of AWS Systems Manager (formerly known as SSM).

It is required according to docs on
    https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ssm-parameter.html
    
These docs don't mention anything about optionality, but they do mention the field and supply it in their examples:
    
- Systems Manager API reference for get_parameter method
      https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_GetParameter.html
- Ruby Parameter docs
      https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/SSM/Types/Parameter.html
- Python get_parameter docs
      https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ssm.html#SSM.Client.get_parameter
    
In the Go bindings, it is a pointer, which suggests that it is optional. But it seems like everything is a pointer.
    https://docs.aws.amazon.com/sdk-for-go/api/service/ssm/#Parameter
    
This piece of Parameter documentation claims it is not required:
    https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_Parameter.html

So what to do? Don't know whether I should make a PR.

_Originally posted by @ysangkok in https://github.com/brendanhay/amazonka/issues/739#issuecomment-998275138_

ysangkok:
These were tested by trying the commands put-parameter, get-parameter
and get-parameters with awscli.

```
 % aws ssm put-parameter --name test-dec-20 --value "abc" --type String
{
    "Version": 1
}
 % aws ssm get-parameter --name test-dec-20                            
{
    "Parameter": {
        "Version": 1,
        "Value": "abc",
        "Type": "String",
        "Name": "test-dec-20"
    }
}
 % aws ssm get-parameters --names test-dec-20
{
    "Parameters": [
        {
            "Name": "test-dec-20",
            "Value": "abc",
            "Type": "String",
            "Version": 1
        }
    ],
    "InvalidParameters": []
}
 % aws ssm get-parameters --names test-dec-21
{
    "InvalidParameters": [
        "test-dec-21"
    ],
    "Parameters": []
}
 % aws ssm delete-parameter --name test-dec-20 
 % aws ssm delete-parameter --name test-dec-20

An error occurred (ParameterNotFound) when calling the DeleteParameter operation:
```

Fixes #740.

### DIFF
--- a/configs/services/ssm.json
+++ b/configs/services/ssm.json
@@ -1,3 +1,29 @@
 {
-    "libraryName": "amazonka-ssm"
+    "libraryName": "amazonka-ssm",
+    "typeOverrides": {
+        "GetParameterResult": {
+            "requiredFields": [
+                "Parameter"
+            ]
+        },
+        "GetParametersResult": {
+            "requiredFields": [
+                "InvalidParameters",
+                "Parameters"
+            ]
+        },
+        "PutParameterResult": {
+            "requiredFields": [
+                "Version"
+            ]
+        },
+        "Parameter": {
+            "requiredFields": [
+                "Name",
+                "Type",
+                "Value",
+                "Version"
+            ]
+        }
+    }
 }
--- a/lib/amazonka/CHANGELOG.md
+++ b/lib/amazonka/CHANGELOG.md
@@ -9,6 +9,8 @@ Released: **?**, Compare: [2.0.0-rc1](https://github.com/brendanhay/amazonka/com
 [\#725](https://github.com/brendanhay/amazonka/pull/725)
 - Hosted Zone IDs for S3 website endpoints are correct for all regions.
 [\#723](https://github.com/brendanhay/amazonka/pull/723)
+- `amazonka-ssm`: Various fields that are always available are no longer Maybe's.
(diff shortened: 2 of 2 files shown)


ysangkok:
I am unsure of the naming, and I don't know how to verify it since I can't get Bazel working. Is it GetParameterResult or GetParameterResponse? `configs/annexes/ssm.json` uses Result but the source code uses Response. Same goes for the other names.

endgame:
You'll want to use the names of the shapes in [`botocore`](https://github.com/boto/botocore/blob/25978afbdfb3052c00dcae65d18c6ca3bbc9dcd9/botocore/data/ssm/2014-11-06/service-2.json) when writing out `typeOverrides`. The generator normalises a bunch of different conventions from different services when generating the names of Haskell types.

Could you please also add a changelog entry?

ysangkok:
Ok, added!

endgame:
Thanks!

PR #743, opened 2022-01-05
Regenerate all services

endgame:
Regenerate after the recent dynamodb fixes and the Hashable/NFData instance ordering/off-by-one fixes.


endgame:
@brendanhay you may prefer a var called `salt'` instead - I'm happy to switch the generator and regenerate if so.

endgame:
I'm going to merge this now, to make it easier for people to test things.

Issue #742, opened 2022-01-03
Performance of request signing


ethercrow:
While profiling a service I've seen Amazonka request signing featured prominently.

From a quick look at https://github.com/brendanhay/amazonka/blob/main/lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs my first suspect would be bytestring operations like `<>` and `BS8.intercalate` which contain `malloc` inside and having too many of those is usually not great.

It would be great to have a benchmark for request signing, I'd imagine it would be useful for every user of Amazonka if the signing was efficient.

endgame:
It would indeed be great, but a release is a higher priority. That said, I'd certainly look at a PR if you had time to write one (even if only adding a benchmark).

endgame:
Thoughts, after discussion with @axman6 :

* Replace uses of chained `(<>)` with `ByteString.concat`,
* `canonicalHeaders` might be better if it used a `Builder`,
* `BS8.intercalate s = concat . intersperse s`, and `concat` is smart (it computes the destination length and allocates a single buffer)

Do you have a benchmark case? It would be good to know if changes actually improve things.

endgame:
And fix the FIXME about using an ordered map.

PR #744, opened 2022-01-06
Signing improvements

axman6:
Based on #742, this might help with the performance of signing requests by avoiding repeated concatenation of bytestrings. It should have the same behaviour as the previous code but I haven't tested that requests work.

@ethercrow can you try out this patch and see if it helps at all? I wouldn't be surprised if it doesn't, but it cleans things up a bit and avoids some obvious less than optimal code.

### DIFF
--- a/lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs
+++ b/lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs
@@ -15,7 +15,9 @@ import Amazonka.Prelude
 import Amazonka.Request
 import Amazonka.Types
 import qualified Data.ByteString as BS
+import qualified Data.ByteString.Builder as BSB
 import qualified Data.ByteString.Char8 as BS8
+import qualified Data.ByteString.Lazy as BSL
 import qualified Data.CaseInsensitive as CI
 import qualified Data.Foldable as Foldable
 import qualified Data.Map.Strict as Map
@@ -272,9 +274,9 @@ canonicalQuery = Tag . toBS
 -- all internal whitespace, replacing with a single space char,
 -- unless quoted with \"...\"
 canonicalHeaders :: NormalisedHeaders -> CanonicalHeaders
-canonicalHeaders = Tag . Foldable.foldMap (uncurry f) . untag
+canonicalHeaders = Tag . BSL.toStrict . BSB.toLazyByteString . Foldable.foldMap (uncurry f) . untag
   where
-    f k v = k <> ":" <> stripBS v <> "\n"
+    f k v = BSB.byteString k <> BSB.char7 ':' <> BSB.byteString (stripBS v) <> BSB.char7 '\n'
 
 signedHeaders :: NormalisedHeaders -> SignedHeaders
 signedHeaders = Tag . BS8.intercalate ";" . map fst . untag
--- a/lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs
+++ b/lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs
@@ -18,8 +18,7 @@ import qualified Data.ByteString as BS
 import qualified Data.ByteString.Char8 as BS8
 import qualified Data.CaseInsensitive as CI
 import qualified Data.Foldable as Foldable
-import qualified Data.Function as Function
-import qualified Data.List as List
+import qualified Data.Map.Strict as Map
 import qualified Network.HTTP.Client as Client
 import qualified Network.HTTP.Types as HTTP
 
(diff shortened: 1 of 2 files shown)


lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs
```diff
@@ -271,19 +274,18 @@ canonicalQuery = Tag . toBS
...
 normaliseHeaders :: [Header] -> NormalisedHeaders
 normaliseHeaders =
-  -- FIXME: convert this to an ordered map.
   Tag
     . map (first CI.foldedCase)
-    . List.nubBy ((==) `Function.on` fst)
-    . List.sortBy (compare `Function.on` fst)
-    . filter ((/= "authorization") . fst)
-    . filter ((/= "content-length") . fst)
+    . Map.toList
```
endgame, line 288: https://hackage.haskell.org/package/containers-0.6.5.1/docs/Data-Map-Strict.html#v:toAscList
```suggestion
    . Map.toAscList
```

ethercrow:
Hi @axman6, thank you for working on it!

I've just tried this branch, the performance difference is small enough that I don't see it in the profile noise.

endgame:
Could you please attach a profile (redacted if necessary)? It is not fun to attempt make performance improvements while nearly blind.

The other thing @axman6 and I discussed last night is the possibility that the signing step is the first point at which the various request objects are forced, and that's why it shows up here.

axman6:
@ethercrow Can you provide any more info? We're trying to help, but as @endgame has said, we're doing this blind. I would recommend you also try the following: if you have code that looks like

```haskell
send (someRequest someParams)
```
try changing it to:
```haskell
import Control.DeepSeq.Generics (genericRnf) -- From package deepseq-generics

let req = someRequest someParams
    !_ = genericRnf req
in send req
```

(Sadly it doesn't look like we derive NFData instances for requests in amazonka... this is probably a good thing for compilation time though.)

and then see if the profile shows the time being spent in signing still, or if it's happening in your module.

A profile report would be very useful for helping you if you can provide one.

endgame:
> (Sadly it doesn't look like we derive NFData instances for requests in amazonka... this is probably a good thing for compilation time though.)

Yeah we do, but this is a recent change that Brendan wrote:

https://github.com/brendanhay/amazonka/blob/c063bc7329b51d83afe4cd454c38d1645f3d4774/lib/services/amazonka-accessanalyzer/gen/Amazonka/AccessAnalyzer/CreateAnalyzer.hs#L155-L161

lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs
```diff
@@ -130,13 +131,15 @@ type Signature = Tag "signature" ByteString
...
 authorisation :: V4 -> ByteString
 authorisation V4 {..} =
-  algorithm
-    <> " Credential="
-    <> toBS metaCredential
-    <> ", SignedHeaders="
-    <> toBS metaSignedHeaders
-    <> ", Signature="
-    <> toBS metaSignature
+  mconcat
```
brendanhay, line 134: *squints*

lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs
```diff
@@ -271,19 +274,18 @@ canonicalQuery = Tag . toBS
...
 normaliseHeaders :: [Header] -> NormalisedHeaders
 normaliseHeaders =
-  -- FIXME: convert this to an ordered map.
   Tag
     . map (first CI.foldedCase)
-    . List.nubBy ((==) `Function.on` fst)
-    . List.sortBy (compare `Function.on` fst)
-    . filter ((/= "authorization") . fst)
-    . filter ((/= "content-length") . fst)
+    . Map.toAscList
```
brendanhay, line 288: The original and somewhat unfounded intuition was that when the list of headers is small, in the range of <10ish items, then you're paying more to construct/allocate the `Map`.  Due to the use of `nubBy` you're getting `O(n^2)` and possibly the moon, but it all depends on the size of the input as to how much of an effect this actually has in reality.

At any rate, the `Map` code is arguably nicer, so even in the absence of performance-related evidence I'd be fine with merging it.

lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs
endgame: We checked the implementation of `mconcat` for `ByteString` before making this change - it first works out exactly how much space is needed, allocates that, and then copies the bits in.

axman6: Repeated application of `<>` leads to O(n^2) complexity - not likely to be a killer but worth getting into the habit in the system of picking the efficient alternative if there is one. It's not a big issue one way or the other, just thought I'd slip it in while trying to work on another performance issue.

lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs
axman6: I can see there's some irony with what I've said above and also suggesting this, but I think the cleaner code is worth it.

Issue #271, opened 2016-02-11
EC2 credentials discovery fails (instance-data host gone?)


screamish:
I'm trying to use Discover for credentials from on an AWS EC2 instance and it's throwing this exception:

```"MissingFileError \"/root/.aws/credentials\"```

However I've had a look at the code that's doing this and this function seems like Amazon might have broken a dependency.

https://github.com/brendanhay/amazonka/blob/develop/amazonka/src/Network/AWS/EC2/Metadata.hs#L282

When I try to run this from an ssh terminal on my EC2 instance I get the following:

```$ curl http://instance-data/latest/meta-data/
curl: (6) Could not resolve host: instance-data```

but if I try the following it works:

```$ curl http://169.254.169.254/latest/meta-data/
ami-id
ami-launch-index
... snip ...```

I couldn't find anything on AWS's docs about them removing that hostname. What do you think?

**Comment by brendanhay** [comment]
That is concerning. The `instance-data` name resolution is used because it's convenient for the EC2 test, if that no longer existed I'd need some alternative way of 'quickly' determining if the host is in EC2 or not.

Which region and instance type are you encountering this issue on?

**Comment by screamish** [comment]
ap-southeast-2, t2.small

Is there a reason you can't just use this:
```
latest :: Text
latest = "http://169.254.169.254/latest/"
```

like you do in all the other calls?

brendanhay:
Because it relies on name resolution failure, instead of a timeout. If `169.254.169.254` is used - circumstances such as an overloaded host could result in the timeout being reached, and the instance not being detected as an EC2 host. I'm open to other suggestions, but I'd prefer to avoid the timeout if possible.

In the interim, if you're deploying to EC2 I'd suggest using `newEnv Sydney `(FromProfile "default")`, or whatever the default IAM profile on that host happens to be. This way the `isEC2` check will be skipped entirely.

screamish:
Ah, thanks for the explanation, that makes good sense. And thanks for the workaround, it covers my use case since I don't need to deploy this multi-region. If I think of any ideas for quick, robust way to implement `isEC2` I'll let you know.

gregwebs:
I just ran into this same issue. I would suggest not worrying about the overloaded host issue so much unless you have good reason to. My reasoning is that generally finding credentials is done at startup. If the host is already overloaded and you start a new process you are probably in for trouble already.

brendanhay:
[...]

PR #746, opened 2022-01-10
Credential chain

### LINKED ISSUE #476: Support for AWS config file
When using the AWS CLI we make use of profiles, defined in `~/.aws/config`, not `~/.aws/credentials`.  The CLI can lookup profiles in there then fetch the credentials from `~/.aws/credentials` ([here](https://docs.aws.amazon.com/cli/latest/userguide/cli-multiple-profiles.html)), and it looks like `boto3` can do this too ([here](https://boto3.readthedocs.io/en/latest/guide/configuration.html#configuring-credentials)).  However the docs for the Java and Javascript SDKs don't appear to mention this.  Are there plans to support making use of the `config` file in amazonka?

### LINKED ISSUE #514: Native AssumeRole credentials support
Poking around the code a bit, the `Credentials` type feels a bit confusing to me. For example, it's unclear why `FromContainer` is an explicit option for getting temporary credentials the ECS way, but there's no equivalent `FromEc2Metadata` constructor for getting them the EC2 way, and you need to use the much more complicated `Discover` constructor for that, or manually retrieve them yourself and feed them into `FromSession`. The reason I bring this up is because factoring these out slightly differently could open up a neat possibility:

There's some nice support for an automatic expiration thread for those temporary credentials that could greatly benefit dynamic `AssumeRole` support, which I use extensively. What I'd like to be able to do, ideally, is `newEnv (Chain Discover [AssumeRole "arn:aws:iam::12345:role/foo" "mysession", AssumeRole "arn:aws:iam::23456:role/bar" "anothersessionid"])` and have amazonka automatically make the STS `AssumeRole` calls for me, as well as handling credential expiry threads for all three sets of temporary credentials.

Any thoughts? I could throw a version of it into a PR if anyone was interested, but I wanted to float the idea before putting effort into it.

### LINKED ISSUE #629: Feature Request: Credentials supports `source_profile` in `.aws/config`
Currently `Credentials` (`FromFile` and `Discover`) can't read `~/.aws/credentials` specified by `source_profile` in `~/.aws/config`.

### local `~/.aws/config`

```
[profile first]
region = ap-northeast-1

[profile second]
role_arn = arn:aws:iam::000000000000:role/Sample
source_profile = first
```

### local `~/.aws/credentials`

```
[first]
aws_access_key_id        = XXXXXXXXXXXXXXXXXXXX
aws_secret_access_key    = XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
aws_session_token        = XXX...
aws_security_token       = XXX...
x_principal_arn          = arn:aws:sts::000000000000:assumed-role/First/sample@example.com
x_security_token_expires = 2021-05-19T13:05:47+09:00
```

### env

```console
$ export AWS_PROFILE=second
$ export AWS_REGION=ap-northeast-1
```

### haskell

```haskell
sample :: IO GetCallerIdentityResponse
sample = do
  env <- newEnv Discover
  runResourceT . runAWS env $ send getCallerIdentity
```

```
*** Exception: InvalidFileError "/home/dan/.aws/credentials, key aws_access_key_id Couldn't find section: second"
```

endgame:
Rewrite the authentication code to support something like an explicit "credential chain", in the vein of the official AWS SDKs. The `Credentials` data type is removed in favour of authentication functions of type `Env' withEnv -> m Env` or `Env -> m Env`; that's a function that takes an input `Env'` (either indifferent to authentication or in the case of `sts:AssumeRole`, requiring it), and produces a new output `Env`. This allows library clients much greater control over how they get the first set of access keys - if AWS releases some new way to acquire credentials, it's possible to plug that in much more easily than before.

This PR also breaks up the authentication into a bunch of separate modules. This has the nice side-effect of removing the `.hs-boot` files which came in when we started supporting `sts:AssumeRoleWithWebIdentity`.

On the library consumer side, the new `Amazonka.Auth.STS.fromAssumedRole` makes it easy to spin off a new `Env` under an assumed role, and the parser for credentials/config files has been rewritten to support many authentication methods supported by the official SDKs.

This also provides a workaround for the VPC issues in #271 - it's possible to write something like `newEnv fromDefaultInstanceProfile` and have amazonka skip the `isEC2` check. (The IMDS client connects by IP; only the `isEC2` check uses `http://instance-data`.)

Migration for most users will be to replace `newEnv Discover` with `newEnv discover`.

Recommend reviewing commit-by-commit.

Remaining work:

- [x] Punt logging format discussion to new issue
- [x] Detect infinite loops in config profiles
- [x] Punt de-underscoring field names to new issue
- [x] Mark support for profiles without `profile` prefix in config as "works, but not supported"
- [x] Fix failure to read credentials when `~/.aws/credentials` exists but `~/.aws/config` doesn't.

Closes #514 
Closes #476
Closes #629


endgame:
CC:

* @axman6 , who helped brainstorm the design (thanks very much!),
* @fumieval @K0Te , who were in discussions when the initial version of the `sts:AssumeRoleWithWebIdentity` code was being hammered out, and may have opinions,
* @lrworth @unisay , who may also have opinions since they'll need to build atop the new auth stuff as soon as this looks solid.

lib/amazonka/CHANGELOG.md
```diff
@@ -3,6 +3,33 @@
...
+    | `FromKeys`        | `fromKeys`                   | Explicit access key id/secret access key.                                                                                                       |
+    | `FromSession`     | `fromSession`                | As above, but with session token.                                                                                                               |
+    |                   | `fromTemporarySession`       | As above, but with token with expiry time. Likely not useful.                                                                                   |
+    |                   | `fromKeysEnv`                | Reads `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and optionally `AWS_SESSION_TOKEN`.                                                         |
+    | `FromEnv`         |                              |                                                                                                                                                 |
```
lrworth, line 19: Should these be in the same row?

lib/amazonka/src/Amazonka/Auth/ConfigFile.hs
```diff
@@ -0,0 +1,230 @@
...
+
+-- | Retrieve credentials from the AWS config/credentials files, as
+-- Amazonka currently understands them:
+--
+-- * AWS recommends credentials do not live in the config file, but
+--   allows it.
+--
+-- * Sections in the config file start should either be named
+--   @[default]@ or @[profile foo]@, but unprefixed @[foo]@ is also
+--   supported.
```
lrworth, line 39: Is unprefixed format really supported? [The documentation](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html) says 
> When naming the profile in a `config` file, include the prefix word "profile".

lib/amazonka/src/Amazonka/Auth/ConfigFile.hs
```diff
@@ -0,0 +1,230 @@
...
+fromFileEnv ::
+  (MonadIO m, Foldable withAuth) => Env' withAuth -> m Env
+fromFileEnv env = liftIO $ do
+  mProfile <- Environment.lookupEnv "AWS_PROFILE"
+  cred <- file "/.aws/credentials"
+  conf <- file "/.aws/config"
+
+  fromFilePath (maybe "default" Text.pack mProfile) cred conf env
+  where
+    file p = catching_ _IOException dir err
```
lrworth, line 227:
```suggestion
    homePath p = catching_ _IOException dir err
```

and perhaps `credPath` and `confPath` above.

endgame:
Question for the floor: would it be worth adding pattern synonyms where there's a compatible new name?

lib/amazonka/src/Amazonka/Auth/ConfigFile.hs
endgame: I have some unprefixed `[foo]` sections in my profile where I hand-edited it, and they work.

lib/amazonka/CHANGELOG.md
endgame: No, they're different. `FromEnv` took in the _names_ of the environment variables to use, whereas `fromKeysEnv`. I will push a clarification.

lib/amazonka/CHANGELOG.md
```diff
@@ -3,6 +3,33 @@
...
+    | `FromKeys`        | `fromKeys`                   | Explicit access key id/secret access key.                                                                                                       |
+    | `FromSession`     | `fromSession`                | As above, but with session token.                                                                                                               |
+    |                   | `fromTemporarySession`       | As above, but with token with expiry time. Likely not useful.                                                                                   |
+    |                   | `fromKeysEnv`                | Reads `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and optionally `AWS_SESSION_TOKEN`.                                                         |
+    | `FromEnv`         |                              | Removed - used to look up named environment variables for access key id/secret access key/session token/expiry time.                            |
```
endgame, line 19: Please sing out if you need this - it would be easy to add a `fromKeysEnvNamed` function, but then I would feel obliged to do it for every `fromFooEnv` function, which feels a bit odd since nobody really needed _those_.

lib/amazonka/src/Amazonka/Auth/ConfigFile.hs
lrworth: "Supported" usually means it has official support from the vendor. I think "happens to work" would be more accurate.

See: https://github.com/boto/botocore/blob/develop/botocore/configloader.py#L216

AlexeyRaga:
I like it. 
Things like `fromAssumedRole`, `fromKeysEnv` are very useful.

axman6 (commented):
Plenty of comments but I don't think there's anything that _needs_ to be changed.

Nice work @endgame, I feel like this change is very much in line with the changes to enums, where users aren't completely stuffed when Amazon make changes and Amazonka isn't updated fast enough for them.

lib/amazonka/src/Amazonka.hs
```diff
@@ -48,7 +47,7 @@ module Amazonka
     AccessKey (..),
     SecretKey (..),
     SessionToken (..),
-    Credentials (..),
+    discover,
```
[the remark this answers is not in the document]
endgame: Everything's pushed people towards discover since ages. The other auth names look strange if you import them straight out of `Amazonka`, but look okay under `import qualified Amazonka.Auth as Auth` (`Auth.fromKeys _ _` etc). I will add more haddocks to remind people that they exist.

lib/amazonka/src/Amazonka/Auth.hs
```diff
@@ -66,786 +48,81 @@ module Amazonka.Auth
...
+    [ fromKeysEnv,
+      fromFileEnv,
+      fromWebIdentityEnv,
+      fromContainerEnv,
+      -- NOTE: This is not 100% consistent with the AWS SDKs, which
+      -- does not attempt to query the ECS service if either
+      -- AWS_CONTAINER_CREDENTIALS_RELATIVE_URI or
+      -- AWS_CONTAINER_CREDENTIALS_FULL_URI are set.
+      --
+      -- See: https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/EC2ContainerCredentialsProviderWrapper.html
```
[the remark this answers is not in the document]
endgame: Yes, will move to haddock.

lib/amazonka/src/Amazonka/Auth/ConfigFile.hs
```diff
@@ -0,0 +1,230 @@
...
+--   supported.
+--
+-- * Sections in the credentials file are always unprefixed -
+--   @[default]@ or @[foo]@.
+--
+-- /See:/ the 'ConfigProfile' type, to understand the methods Amazonka
+-- currently supports.
+fromFilePath ::
+  forall m withAuth.
+  (MonadIO m, Foldable withAuth) =>
```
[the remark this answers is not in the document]
endgame: Here, it is: `retryRequest` needs it, which induces the constraint on `fromWebIdentity`, which we call in this function.

lib/amazonka/src/Amazonka/Auth/Background.hs
```diff
@@ -0,0 +1,68 @@
...
+        w <- IORef.mkWeakIORef r (Concurrent.killThread s)
+
+        loop ma w p x
+
+    loop :: IO AuthEnv -> Weak (IORef AuthEnv) -> ThreadId -> ISO8601 -> IO ()
+    loop ma w !p !x = do
+      diff x <$> Time.getCurrentTime >>= Concurrent.threadDelay
+      env <- Exception.try ma
+      case env of
+        Left e -> Exception.throwTo p (RetrievalError e)
```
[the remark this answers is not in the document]
endgame: File reading does not use the background fetcher - only sources of _temporary_ credentials like the ones you get from STS or instance profile, or from the ECS Container Agent.

lib/amazonka/src/Amazonka/Auth/Background.hs
```diff
@@ -0,0 +1,68 @@
...
+        Right !a -> do
+          mr <- Weak.deRefWeak w
+          case mr of
+            Nothing -> pure ()
+            Just r -> do
+              IORef.atomicWriteIORef r a
+              maybe (pure ()) (loop ma w p) (_authExpiration a)
```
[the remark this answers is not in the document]
axman6-da: Yeah all good, it just got me thinking about semantics - do we want to fetch every n period of time or to wait n period of time between fetches. Not sure it would ever really cause issues.

lib/amazonka/src/Amazonka/Auth/ConfigFile.hs
```diff
@@ -0,0 +1,230 @@
...
+fromFileEnv env = liftIO $ do
+  mProfile <- Environment.lookupEnv "AWS_PROFILE"
+  cred <- homePathRelative "/.aws/credentials"
+  conf <- homePathRelative "/.aws/config"
+
+  fromFilePath (maybe "default" Text.pack mProfile) cred conf env
+  where
+    homePathRelative p = catching_ _IOException dir err
+      where
+        dir = Directory.getHomeDirectory <&> (++ p)
```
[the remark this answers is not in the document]
endgame: I suppose if we're shipping breaking changes, now is the time.

lib/amazonka/src/Amazonka/Auth/Exception.hs
```diff
@@ -0,0 +1,90 @@
...
+
+instance Exception AuthError
+
+instance ToLog AuthError where
+  build = \case
+    RetrievalError e -> build e
+    MissingEnvError e -> "[MissingEnvError]  { message = " <> build e <> "}"
+    MissingFileError f -> "[MissingFileError] { path = " <> build f <> "}"
+    InvalidFileError e -> "[InvalidFileError] { message = " <> build e <> "}"
+    InvalidIAMError e -> "[InvalidIAMError]  { message = " <> build e <> "}"
```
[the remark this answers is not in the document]
endgame: I don't know. They've been like this for ages. @brendanhay ?

lib/amazonka/src/Amazonka/Auth/InstanceProfile.hs
```diff
@@ -0,0 +1,99 @@
...
+    manager = _envManager env
+
+    handleErr f g = \case
+      Left e -> Exception.throwIO (RetrievalError e)
+      Right x -> either (Exception.throwIO . g) pure (f x)
+
+    invalidIAMErr =
+      InvalidIAMError
+        . mappend ("Error parsing IAM profile '" <> name <> "' ")
+        . Text.pack
```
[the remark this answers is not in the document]
endgame: The second will only be a single `mappend` either way but I'll do both for consistency.

lib/amazonka/src/Amazonka/Auth/ConfigFile.hs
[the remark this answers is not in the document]
endgame: Fixed the path on Windows - take a look?

endgame:
Another question: Should we detect infinite loops in the AWS config files? The other SDKs do, and we probably should.

axman6-da (approved):
🤔

lib/amazonka/src/Amazonka/Env.hs
```diff
@@ -42,20 +41,36 @@ module Amazonka.Env
...
+-- information.
+data Env' withAuth = Env
+  { _envRegion :: Region,
+    _envLogger :: Logger,
+    _envRetryCheck :: Int -> Client.HttpException -> Bool,
+    _envOverride :: Dual (Endo Service),
+    _envManager :: Client.Manager,
+    _envAuth :: withAuth Auth
+  }
+  deriving stock (Generic)
```
brendanhay, line 72: If `Generic` is used for `generic-lens` or `optics` support - the `_` prefixes at the least should probably also be removed.

lib/amazonka/src/Amazonka/Auth/Exception.hs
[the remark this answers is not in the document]
brendanhay: Who knows. Logging should be redone at some point in the future - I'd rather emit a simple line-oriented format designed for a reading by an operator by default and then let the library consumer configure/override formats such as JSON, if they wish.

lib/amazonka/src/Amazonka/Auth/Background.hs
```diff
@@ -0,0 +1,70 @@
+{-# LANGUAGE BangPatterns #-}
```
brendanhay, line 1: Not sure if you copied over the original bang patterns or not - but if so, the original reasoning behind them seems to be lost. I'd consider just removing them where they don't clearly beat you about the head with their obviousness, or alternatively mark this entire module `{-# LANGUAGE Strict #-}`.

brendanhay (approved):
Changes look good, I like the new design!

brendanhay:
> Another question: Should we detect infinite loops in the AWS config files? The other SDKs do, and we probably should.

We should (eventually) - but I don't think it should get in the way of getting this released.

lib/amazonka/src/Amazonka/Env.hs
ysangkok: If they should be removed, they should also be removed in e.g. `AuthEnv(_authAccessKeyId)` from Amazonka/Types.hs, no?

brendanhay: Yes, it would be nice to remove them fully from the library - although perhaps this would be better as a separate PR.

endgame:
> > Another question: Should we detect infinite loops in the AWS config files? The other SDKs do, and we probably should.
> 
> We should (eventually) - but I don't think it should get in the way of getting this released.

I think it should be quick (execStateT while traversing the profile stuff), so I'll try to do it on this PR. Most other stuff should be moved into separate issues, which I'll do before merging this.

lib/amazonka/src/Amazonka/Auth.hs
```diff
@@ -17,43 +17,25 @@ module Amazonka.Auth
...
     fromKeys,
     fromSession,
     fromTemporarySession,
-    fromEnv,
-    fromEnvKeys,
-    fromFile,
+    fromKeysEnv,
     fromFilePath,
-    fromProfile,
-    fromProfileName,
```
JackKelly-Bellroy, line 55: export `fromAssumedRole` here too

lib/amazonka/src/Amazonka/Env.hs
endgame: Agreed. I'll do the ones on `Env'` now, and defer a general record cleanup to https://github.com/brendanhay/amazonka/issues/752.

lib/amazonka/src/Amazonka/Auth/Background.hs
endgame: OK.

lib/amazonka/src/Amazonka/Auth/Exception.hs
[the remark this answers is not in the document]
endgame: #753

examples/src/DynamoDB.hs
```diff
@@ -65,7 +67,9 @@ insertItem region secure host port table item = do
   let dynamo = setEndpoint secure host port DynamoDB.defaultService
 
   lgr <- newLogger Debug stdout
-  env <- newEnv discover <&> set envLogger lgr . within region . configure dynamo
+  env <-
+    newEnv discover
+      <&> set (field @"envLogger") lgr . configure dynamo . within region
```
axman6-da, line 72:
```suggestion
    newEnv discover <&> set #envLogger lgr . configure dynamo . within region
```
Seems this is a bit inconsistent intros and the following files.

lib/amazonka/src/Amazonka/Auth/ConfigFile.hs
```diff
@@ -89,10 +115,19 @@ fromFilePath profile credentialsFile configFile env = do
...
               AssumeRoleFromProfile roleArn sourceProfileName -> do
-                sourceEnv <- evalConfig config sourceProfileName
-                fromAssumedRole roleArn "amazonka-assumed-role" sourceEnv
+                seenProfiles <- lift get
+                if sourceProfileName `elem` seenProfiles
+                  then
+                    liftIO
+                      . Exception.throwIO
+                      . SourceProfileLoop
+                      $ reverse seenProfiles
```
axman6-da, line 126: Nice!

axman6-da (commented):
Just some quick things I noticed while skimming. Nice work man.

examples/src/DynamoDB.hs
endgame: I'm specifically doing some of them with `OverloadedLabels` and some with raw `generic-lens` just so people can see both ways.

lib/amazonka/src/Amazonka/Auth/ConfigFile.hs
endgame: I rewrote this part since I don't think it needs a new exception constructor, but yea it wasn't too bad.

endgame:
Going to leave this bake for a few days longer now that I've updated work code to run on it, but I think there's nothing stopping a merge early next week.

endgame:
This has been working well for as at work, so I'm going to merge it.

PR #757, opened 2022-02-09
Implement Auth.SSO

pbrisbin:
- REVERT ME: Configure for Stack

  I had trouble with the nix setup; this worked straight away. I can remove it
  before merge.

- Expose Auth.ConfigFile functions for re-use

  The way ConfigFile works right now, I couldn't draw the perfect seams to
  totally DRY what I'm doing; but this seems like a good 80/20.

- Implement Auth.SSO

  Huzzah!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,6 @@ dist*
 
 # Direnv
 .envrc.local
+
+# Stack
+.stack-work
--- a/lib/amazonka/amazonka.cabal
+++ b/lib/amazonka/amazonka.cabal
@@ -84,6 +84,7 @@ library
     Amazonka.Auth.Exception
     Amazonka.Auth.InstanceProfile
     Amazonka.Auth.Keys
+    Amazonka.Auth.SSO
     Amazonka.Auth.STS
     Amazonka.EC2.Metadata
     Amazonka.Env
@@ -95,7 +96,9 @@ library
     Amazonka.Data, Amazonka.Types, Amazonka.Bytes, Amazonka.Endpoint, Amazonka.Crypto
 
   build-depends:
+    , aeson                 ^>=1.5.0.0 || ^>=2.0.0.0
     , amazonka-core         ^>=2.0
+    , amazonka-sso          ^>=2.0
     , amazonka-sts          ^>=2.0
     , bytestring            >=0.10.8
     , conduit               >=1.3
--- a/lib/amazonka/src/Amazonka/Auth.hs
+++ b/lib/amazonka/src/Amazonka/Auth.hs
@@ -37,6 +37,7 @@ module Amazonka.Auth
     fromWebIdentityEnv,
     fromDefaultInstanceProfile,
     fromNamedInstanceProfile,
+    fromSSO,
 
     -- ** Keys
     AccessKey (..),
@@ -60,6 +61,7 @@ import Amazonka.Auth.Container (fromContainer, fromContainerEnv)
 import Amazonka.Auth.Exception
 import Amazonka.Auth.InstanceProfile (fromDefaultInstanceProfile, fromNamedInstanceProfile)
 import Amazonka.Auth.Keys (fromKeys, fromKeysEnv, fromSession, fromTemporarySession)
+import Amazonka.Auth.SSO (fromSSO)
 import Amazonka.Auth.STS (fromAssumedRole, fromWebIdentity, fromWebIdentityEnv)
 import Amazonka.EC2.Metadata
 import Amazonka.Env (Env, EnvNoAuth, Env' (..))
--- a/lib/amazonka/src/Amazonka/Auth/ConfigFile.hs
+++ b/lib/amazonka/src/Amazonka/Auth/ConfigFile.hs
@@ -249,12 +249,13 @@ fromFileEnv env = liftIO $ do
   conf <- configPathRelative "/.aws/config"
 
   fromFilePath (maybe "default" Text.pack mProfile) cred conf env
+
+configPathRelative :: String -> IO String
+configPathRelative p = handling_ _IOException err dir
   where
-    configPathRelative p = handling_ _IOException err dir
-      where
-        err = Exception.throwIO $ MissingFileError ("$HOME" ++ p)
-        dir = case os of
-          "mingw32" ->
-            Environment.lookupEnv "USERPROFILE"
-              >>= maybe (Exception.throwIO $ MissingFileError "%USERPROFILE%") pure
-          _ -> Directory.getHomeDirectory <&> (++ p)
+    err = Exception.throwIO $ MissingFileError ("$HOME" ++ p)
+    dir = case os of
+      "mingw32" ->
+        Environment.lookupEnv "USERPROFILE"
+          >>= maybe (Exception.throwIO $ MissingFileError "%USERPROFILE%") pure
+      _ -> Directory.getHomeDirectory <&> (++ p)
--- a/lib/amazonka/src/Amazonka/Auth/ConfigFile.hs
+++ b/lib/amazonka/src/Amazonka/Auth/ConfigFile.hs
@@ -13,6 +13,7 @@ import Amazonka.Auth.Container (fromContainerEnv)
 import Amazonka.Auth.Exception
 import Amazonka.Auth.InstanceProfile (fromDefaultInstanceProfile)
 import Amazonka.Auth.Keys (fromKeysEnv)
+import Amazonka.Auth.SSO (fromSSO, relativeCachedTokenFile)
 import Amazonka.Auth.STS (fromAssumedRole, fromWebIdentity)
 import Amazonka.Data
 import Amazonka.Env (Env, Env' (..), lookupRegion)
@@ -138,6 +139,11 @@ fromFilePath profile credentialsFile configFile env = liftIO $ do
                 fromAssumedRole roleArn "amazonka-assumed-role" sourceEnv
               AssumeRoleWithWebIdentity roleArn mRoleSessionName tokenFile ->
                 fromWebIdentity tokenFile roleArn mRoleSessionName env
+              AssumeRoleViaSSO startUrl ssoRegion accountId roleName -> do
+                cachedTokenFile <-
+                  liftIO $
+                    configPathRelative =<< relativeCachedTokenFile startUrl
+                fromSSO cachedTokenFile ssoRegion accountId roleName env
 
             -- Once we have the env from the profile, apply the region
             -- if we parsed one out.
@@ -172,7 +178,8 @@ parseConfigProfile profile = parseProfile <&> \p -> (p, parseRegion)
         [ explicitKey,
           assumeRoleFromProfile,
           assumeRoleFromCredentialSource,
-          assumeRoleWithWebIdentity
+          assumeRoleWithWebIdentity,
+          assumeRoleViaSSO
         ]
 
     parseRegion :: Maybe Region
@@ -213,6 +220,13 @@ parseConfigProfile profile = parseProfile <&> \p -> (p, parseRegion)
         <*> Just (HashMap.lookup "role_session_name" profile)
         <*> (Text.unpack <$> HashMap.lookup "web_identity_token_file" profile)
 
+    assumeRoleViaSSO =
+      AssumeRoleViaSSO
+        <$> HashMap.lookup "sso_start_url" profile
+        <*> (Region' <$> HashMap.lookup "sso_region" profile)
+        <*> HashMap.lookup "sso_account_id" profile
+        <*> HashMap.lookup "sso_role_name" profile
+
 data ConfigProfile
   = -- | Recognizes @aws_access_key_id@, @aws_secret_access_key@, and
     -- optionally @aws_session_token@.
@@ -224,6 +238,9 @@ data ConfigProfile
   | -- | Recognizes @role_arn@, @role_session_name@, and
     -- @web_identity_token_file@.
     AssumeRoleWithWebIdentity Text (Maybe Text) FilePath
+  | -- | Recognizes @sso_start_url@, @sso_region@, @sso_account_id@, and
+    -- @sso_role_name@.
+    AssumeRoleViaSSO Text Region Text Text
   deriving stock (Eq, Show, Generic)
 
 data CredentialSource = Environment | Ec2InstanceMetadata | EcsContainer
--- a/lib/amazonka/src/Amazonka/Auth/SSO.hs
+++ b/lib/amazonka/src/Amazonka/Auth/SSO.hs
@@ -1,0 +1,131 @@
+-- |
+-- Module      : Amazonka.Auth.SSO
+-- Copyright   : (c) 2013-2021 Brendan Hay
+-- License     : Mozilla Public License, v. 2.0.
+-- Maintainer  : Brendan Hay <brendan.g.hay+amazonka@gmail.com>
+-- Stability   : provisional
+-- Portability : non-portable (GHC extensions)
+module Amazonka.Auth.SSO where
+
+import Amazonka.Auth.Background (fetchAuthInBackground)
+import Amazonka.Auth.Exception
+import qualified Amazonka.Crypto as Crypto
+import Amazonka.Data.Sensitive
+import Amazonka.Data.Time (Time (..))
+import Amazonka.Env (Env, Env' (..))
+import Amazonka.HTTP (retryRequest)
+import Amazonka.Lens ((^.))
+import Amazonka.Prelude
+import Amazonka.SSO.GetRoleCredentials as SSO
+import qualified Amazonka.SSO.Types as SSO (RoleCredentials (..))
+import Amazonka.Types
+import qualified Control.Exception as Exception
+import Control.Exception.Lens (handling_, _IOException)
+import Control.Monad.Trans.Resource (runResourceT)
+import Data.Aeson (FromJSON, decodeFileStrict)
+import qualified Data.Text as Text
+import qualified Data.Text.Encoding as Text
+import Data.Time.Clock.POSIX (posixSecondsToUTCTime)
+import qualified Network.HTTP.Client as Client
+
+data CachedAccessToken = CachedAccessToken
+  { startUrl :: Text,
+    region :: Region,
+    accessToken :: Sensitive Text,
+    expiresAt :: UTCTime
+  }
+  deriving stock (Show, Eq, Generic)
+  deriving anyclass (FromJSON)
+
+-- | Assume a role using an SSO Token.
+--
+-- The user must have previously called @aws sso login@, and pass in the path to
+-- the cached token file, along with SSO region, account ID and role name.
+-- ('Amazonka.Auth.ConfigFile.fromFilePath' understands the @sso_@ variables
+-- used by the official AWS CLI and will call 'fromSSO' for you.) This function
+-- uses 'fetchAuthInBackground' to refresh the credentials as long as the token
+-- in the @sso/cache@ file is not expired. When it has, the user will need to
+-- @aws sso login@ again.
+--
+-- <https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sso.html>
+fromSSO ::
+  (MonadIO m, Foldable withAuth) =>
+  FilePath ->
+  Region ->
+  -- | Account ID
+  Text ->
+  -- | Role Name
+  Text ->
+  Env' withAuth ->
+  m Env
+fromSSO cachedTokenFile ssoRegion accountId roleName env = do
+  auth <- liftIO $ fetchAuthInBackground getCredentials
+  pure $ env {envAuth = Identity auth}
+  where
+    getCredentials = do
+      CachedAccessToken {..} <- readCachedAccessToken cachedTokenFile
+
+      -- The Region you SSO through may differ from the Region you intend to
+      -- interact with after. The former is handled here, the latter is taken
+      -- care of later, in ConfigFile.
+      let ssoEnv = env {envRegion = ssoRegion}
+
+      resp <-
+        runResourceT $
+          sendUnsigned ssoEnv $
+            SSO.newGetRoleCredentials roleName accountId (fromSensitive accessToken)
+
+      let mCreds = do
+            rc <- resp ^. SSO.getRoleCredentialsResponse_roleCredentials
+            roleCredentialsToAuthEnv rc
+
+      case mCreds of
+        Nothing -> fail "sso:GetRoleWithCredentials returned no credentials."
+        Just c -> pure c
+
+sendUnsigned ::
+  ( MonadResource m,
+    AWSRequest a,
+    Foldable withAuth
+  ) =>
+  Env' withAuth ->
+  a ->
+  m (AWSResponse a)
+sendUnsigned env req = do
+  eResponse <- retryRequest env req
+  either (liftIO . Exception.throwIO) (pure . Client.responseBody) eResponse
+
+-- | Return the cached token file for a given @sso_start_url@
+--
+-- Matches
+-- [botocore](https://github.com/boto/botocore/blob/c02f3561f56085b8a3f98501d25b9857b916c10e/botocore/utils.py#L2596-L2597),
+-- so that we find tokens produced by @aws sso login@.
+relativeCachedTokenFile :: MonadIO m => Text -> m FilePath
+relativeCachedTokenFile startUrl = do
+  let sha1 = show . Crypto.hashSHA1 $ Text.encodeUtf8 startUrl
+  pure $ "/.aws/sso/cache/" <> sha1 <> ".json"
+
+readCachedAccessToken :: MonadIO m => FilePath -> m CachedAccessToken
+readCachedAccessToken p = liftIO $
+  handling_ _IOException err $ do
+    mCache <- decodeFileStrict p
+    maybe err pure mCache
+  where
+    err =
+      Exception.throwIO $
+        InvalidFileError $
+          mconcat
+            [ "Unable to read SSO cache. ",
+              Text.pack p,
+              " is missing or invalid."
+            ]
+
+roleCredentialsToAuthEnv :: SSO.RoleCredentials -> Maybe AuthEnv
+roleCredentialsToAuthEnv rc =
+  AuthEnv
+    <$> (AccessKey . Text.encodeUtf8 <$> SSO.accessKeyId rc)
+    <*> (fmap (SecretKey . Text.encodeUtf8) <$> SSO.secretAccessKey rc)
+    <*> pure (fmap (SessionToken . Text.encodeUtf8) <$> SSO.sessionToken rc)
+    <*> pure (expirationToExpires <$> SSO.expiration rc)
+  where
+    expirationToExpires = Time . posixSecondsToUTCTime . fromInteger
--- a/lib/amazonka-core/src/Amazonka/Crypto.hs
+++ b/lib/amazonka-core/src/Amazonka/Crypto.hs
@@ -12,6 +12,7 @@ module Amazonka.Crypto
     hmacSHA256,
 
     -- * Hashing
+    hashSHA1,
     hashSHA256,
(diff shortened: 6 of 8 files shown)


lib/amazonka/amazonka.cabal
```diff
@@ -95,7 +96,10 @@ library
     Amazonka.Data, Amazonka.Types, Amazonka.Bytes, Amazonka.Endpoint, Amazonka.Crypto
 
   build-depends:
+    , Glob
```
endgame, line 99: Please add bounds here. From looking at the changelog I think `>= 0.9 && <0.11` should be safe. Then `cabal-fmt --inplace` the cabal file.

.gitignore
```diff
@@ -49,4 +49,7 @@ dist*
 /bazel-*
 
 # Direnv
-.envrc.local
\ No newline at end of file
+.envrc.local
+
+# Stack
+.stack-work
```
endgame, line 55: This is probably worth keeping, even once you remove your stack commit from the PR.

lib/amazonka/src/Amazonka/Auth/SSO.hs
```diff
@@ -0,0 +1,165 @@
...
+  either (liftIO . Exception.throwIO) (pure . Client.responseBody) eResponse
+
+-- | Read @~/.aws/sso/cache/*.json@
+--
+-- This will throw @_MissingEnvError@ if there is no such file or none that
+-- correctly parse; otherwise it returns the first -- though we never expect
+-- there to be more than one (even if you've @sso login@d multiple profiles).
+readCache :: MonadIO m => m Cache
+readCache = liftIO $ do
+  home <- Directory.getHomeDirectory
```
endgame, line 104: This probably needs to use `configPathRelative`.

lib/amazonka/src/Amazonka/Auth/SSO.hs
```diff
@@ -0,0 +1,165 @@
...
+  paths <- globDir1 glob home
```
endgame, line 105: From that botocore snippet I put on the linked issue, we might be able to guess the name of the config file if we can find the right URL to hash? (If we can't, then globbing is fine)

lib/amazonka/src/Amazonka/Auth.hs
```diff
@@ -109,6 +111,7 @@ discover ::
 discover =
   runCredentialChain
     [ fromKeysEnv,
+      fromSSOCache,
```
endgame, line 114: I don't think this will be necessary because we should be doing all the SSO lookup when we handle config files in `fromFileEnv`. The CLI doesn't document that it can do SSO from environment variables, so we can assume that it won't and that we won't need a `fromSSOEnv` either.

lib/amazonka/src/Amazonka/Auth/SSO.hs
```diff
@@ -0,0 +1,165 @@
+-- |
+-- Module      : Amazonka.Auth.SSO
+-- Copyright   : (c) 2013-2021 Brendan Hay
+-- License     : Mozilla Public License, v. 2.0.
+-- Maintainer  : Brendan Hay <brendan.g.hay+amazonka@gmail.com>
+-- Stability   : provisional
+-- Portability : non-portable (GHC extensions)
+--
+-- Authentication using a JWT stored in @~/.aws/sso/cache/{uuid}.json@ as
```
endgame, line 9: Is it actually writing uuids, or the sha of `start_url`?

lib/amazonka/src/Amazonka/Auth/SSO.hs
```diff
@@ -0,0 +1,165 @@
...
+  ( MonadResource m,
+    AWSRequest a,
+    Foldable withAuth
+  ) =>
+  Env' withAuth ->
+  a ->
+  m (AWSResponse a)
+sendWithoutAuth env req = do
+  eResponse <- retryRequest env req
+  either (liftIO . Exception.throwIO) (pure . Client.responseBody) eResponse
```
endgame, line 95: I'd like to see this pulled out somewhere and used to DRY up `Amazonka.Auth.STS`' `assumeRoleWithWebIdentity` support. I don't think it needs to be part of the public interface, but maybe `Amazonka.Auth.Internal.sendWithoutAuth` (new module) would be good?

lib/amazonka/src/Amazonka/Auth/SSO.hs
```diff
@@ -0,0 +1,165 @@
...
+import Amazonka.Prelude
+import Amazonka.SSO.GetRoleCredentials as SSO
+import qualified Amazonka.SSO.Types as SSO (RoleCredentials (..))
+import Amazonka.Types
+import qualified Control.Exception as Exception
+import Control.Monad.Trans.Resource (runResourceT)
+import Data.Aeson (FromJSON, decodeFileStrict)
+import qualified Data.HashMap.Strict as HashMap
+import Data.Text (pack)
+import Data.Text.Encoding (encodeUtf8)
```
endgame, line 31:
```suggestion
import qualified Data.Text as Text
import qualified Data.Text.Encoding as Text
```

endgame (changes requested):
Thanks for PRing this; I think it will be a great help to a lot of people. I've picked a couple of nits, but I think there's a more fundamental restructuring that needs to happen first.

The decision to try SSO as part of looking up a config profile should happen in `ConfigFile.hs`, where it should be treated as "just another auth method". This probably means extending the `ConfigProfile` type (the parsers for the different profiles are in the code immediately above; you will probably want to make a new constructor for SSO):

https://github.com/brendanhay/amazonka/blob/29e4888f847797533c968a855804990ecf8660dd/lib/amazonka/src/Amazonka/Auth/ConfigFile.hs#L216-L227

Then in this `case`-expression, check for your new SSO constructor and in that branch, go looking for cached JSON files. For each one you find, call into the functions you provide in `Amazonka.Auth.SSO`:

https://github.com/brendanhay/amazonka/blob/29e4888f847797533c968a855804990ecf8660dd/lib/amazonka/src/Amazonka/Auth/ConfigFile.hs#L116-L140

When implemented this way, it will get handled in the same way as all other config-file-based methods, which means this hack to set region will apply to it also:

https://github.com/brendanhay/amazonka/blob/29e4888f847797533c968a855804990ecf8660dd/lib/amazonka/src/Amazonka/Auth/ConfigFile.hs#L84-L86

Please also add an entry for this PR (including giving yourself credit!) to `lib/amazonka/CHANGELOG.md`; once we have the names of the new SSO functions nailed down the big table of auth methods will need a new entry too.

*****

> Attempting to use this when an expired `{uuid}.json` is present results in:
> 
> ```
> ServiceError (...snip...)
> ```
> 
> Is that acceptable, or should we catch it and emit something clearer?

I think it's acceptable for now; STS errors seem to be thrown in the same way. If we want to pull them into Auth errors in response to user complaints, that's fine but I don't think we have to hold up this PR to do it.

pbrisbin:
Thank you for the detailed review! I misunderstood some things, but I think I understand better now.

I'll basically be adding support for seeing the `sso_*` fields in `ConfigFile` and, if present, that's where/when to go into the SSO path. This inverts things a little from what I have now, but is a better design (which you know, of course).

The `sso_start_url`, which I'll have handy at that point, then determines which file under `~/.aws/sso/cache` to look for/parse, which solves a number of uncertainties and feels far more reliable. This is a great catch, I was thrown off by the fact that my two Accounts are the same start-url in my use-case.

I'll be sure to format, bounds, and changelog once I get the new approach dialed in. I should be able to push another attempt tonight or tomorrow.

On the topic of formatting, would you be interested in my [Restyled](https://restyled.io) App on this project? It has support for ormolu and I could add cabal-fmt (and anything else really) relatively quickly.

lib/amazonka/src/Amazonka/Auth/ConfigFile.hs
```diff
@@ -138,6 +139,11 @@ fromFilePath profile credentialsFile configFile env = liftIO $ do
                 fromAssumedRole roleArn "amazonka-assumed-role" sourceEnv
               AssumeRoleWithWebIdentity roleArn mRoleSessionName tokenFile ->
                 fromWebIdentity tokenFile roleArn mRoleSessionName env
+              SSO startUrl ssoRegion accountId roleName -> do
+                cache <-
+                  liftIO $
+                    configPathRelative =<< relativeCacheFile startUrl
```
pbrisbin, line 145: Open to ideas here. Passing the `startUrl` to `fromSSO` and having it call `configPathRelative` internally would make much more sense to me, but causes an import loop. This was the best I could come up with.

lib/amazonka/src/Amazonka/Auth/ConfigFile.hs
```diff
@@ -224,6 +238,9 @@ data ConfigProfile
   | -- | Recognizes @role_arn@, @role_session_name@, and
     -- @web_identity_token_file@.
     AssumeRoleWithWebIdentity Text (Maybe Text) FilePath
+  | -- | Recognized @sso_start_url@, @sso_region@, @sso_account_id@, and
```
pbrisbin, line 241:
```suggestion
  | -- | Recognizes @sso_start_url@, @sso_region@, @sso_account_id@, and
```

lib/amazonka/src/Amazonka/Auth/SSO.hs
```diff
@@ -0,0 +1,111 @@
...
+
+sendWithoutAuth ::
+  ( MonadResource m,
+    AWSRequest a,
+    Foldable withAuth
+  ) =>
+  Env' withAuth ->
+  a ->
+  m (AWSResponse a)
+sendWithoutAuth env req = do
```
pbrisbin, line 76: Agreed this should be centralized. Is there is a reason it can't go with `send`/`sendEither`? I'm not sure it needs to be considered "internal" -- I can imagine users wanting to send un-authed requests themselves in some cases.

lib/amazonka/src/Amazonka/Auth/SSO.hs
```diff
@@ -0,0 +1,111 @@
...
+  eResponse <- retryRequest env req
+  either (liftIO . Exception.throwIO) (pure . Client.responseBody) eResponse
+
+relativeCacheFile :: MonadIO m => Text -> m FilePath
+relativeCacheFile startUrl = do
+  sha1 <- undefined startUrl -- TODO
```
pbrisbin, line 82: Need to match the following Python,

```py
hashlib.sha1(start_url.encode('utf-8')).hexdigest()
```

lib/amazonka/src/Amazonka/Auth/SSO.hs
```diff
@@ -0,0 +1,111 @@
...
+
+fromSSO ::
+  (MonadIO m, Foldable withAuth) =>
+  FilePath ->
+  Region ->
+  Text ->
+  Text ->
+  Env' withAuth ->
+  m Env
+fromSSO cache ssoRegion accountId roleName env = do
```
endgame, line 46: As with the comment in `ConfigFile.hs`, maybe rename to `cachedTokenFilePath`?

lib/amazonka/src/Amazonka/Auth/SSO.hs
```diff
@@ -0,0 +1,111 @@
...
+      let ssoEnv = env {envRegion = ssoRegion}
+
+      resp <-
+        runResourceT $
+          sendWithoutAuth ssoEnv $
+            SSO.newGetRoleCredentials roleName accountId accessToken
+
+      let mCreds = do
+            rc <- resp ^. SSO.getRoleCredentialsResponse_roleCredentials
+            roleCredentialsToAuthEnv rc
```
endgame, line 62: Would you like to fix up the service definition, or shall we proceed as-is?

lib/amazonka/src/Amazonka/Auth/SSO.hs
```diff
@@ -0,0 +1,111 @@
...
+  a ->
+  m (AWSResponse a)
+sendWithoutAuth env req = do
+  eResponse <- retryRequest env req
+  either (liftIO . Exception.throwIO) (pure . Client.responseBody) eResponse
+
+relativeCacheFile :: MonadIO m => Text -> m FilePath
+relativeCacheFile startUrl = do
+  let sha1 = show $ Crypto.hashSHA1 $ Text.encodeUtf8 startUrl
+  pure $ "/.aws/sso/cache/" <> sha1 <> ".json"
```
endgame, line 83: Can you please add a comment with that link to the `botocore` repo, and an explanation why we're doing this? Because so much of this behaviour is determined by reading source, it's better if we can make that easier for the devs who come after us. Example:

https://github.com/brendanhay/amazonka/blob/29e4888f847797533c968a855804990ecf8660dd/lib/amazonka-core/src/Amazonka/Request.hs#L182-L183

lib/amazonka/src/Amazonka/Auth/SSO.hs
```diff
@@ -0,0 +1,111 @@
...
+    <*> (overSensitive (SecretKey . Text.encodeUtf8) <$> SSO.secretAccessKey rc)
+    <*> pure (overSensitive (SessionToken . Text.encodeUtf8) <$> SSO.sessionToken rc)
+    <*> pure (expirationToExpires <$> SSO.expiration rc)
+
+expirationToExpires :: Integer -> ISO8601
+expirationToExpires = Time . posixSecondsToUTCTime . fromInteger
+
+-- Should we derive Functor?
+overSensitive :: (a -> b) -> Sensitive a -> Sensitive b
+overSensitive f = Sensitive . f . fromSensitive
```
endgame, line 111: I think deriving Functor is reasonable.

lib/amazonka/src/Amazonka/Auth/SSO.hs
```diff
@@ -0,0 +1,111 @@
...
+roleCredentialsToAuthEnv :: SSO.RoleCredentials -> Maybe AuthEnv
+roleCredentialsToAuthEnv rc =
+  AuthEnv
+    <$> (AccessKey . Text.encodeUtf8 <$> SSO.accessKeyId rc)
+    <*> (overSensitive (SecretKey . Text.encodeUtf8) <$> SSO.secretAccessKey rc)
+    <*> pure (overSensitive (SessionToken . Text.encodeUtf8) <$> SSO.sessionToken rc)
+    <*> pure (expirationToExpires <$> SSO.expiration rc)
+
+expirationToExpires :: Integer -> ISO8601
+expirationToExpires = Time . posixSecondsToUTCTime . fromInteger
```
endgame, line 107: Can you please hang this under a `where` clause in `roleCredentialsToAuthEnv`?

lib/amazonka/src/Amazonka/Auth/SSO.hs
```diff
@@ -0,0 +1,111 @@
...
+  handling_ _IOException err $ do
+    mCache <- decodeFileStrict p
+    maybe err pure mCache
+  where
+    err =
+      Exception.throwIO $
+        InvalidFileError $
+          "Unable to read SSO cache. "
+            <> Text.pack p
+            <> " is missing or invalid."
```
endgame, line 96: Please use `mconcat` when accumulating text like this - it works out how long the string needs to be, and only has to allocate once.

lib/amazonka/src/Amazonka/Auth/SSO.hs
```diff
@@ -0,0 +1,111 @@
...
+  Env' withAuth ->
+  a ->
+  m (AWSResponse a)
+sendWithoutAuth env req = do
+  eResponse <- retryRequest env req
+  either (liftIO . Exception.throwIO) (pure . Client.responseBody) eResponse
+
+relativeCacheFile :: MonadIO m => Text -> m FilePath
+relativeCacheFile startUrl = do
+  let sha1 = show $ Crypto.hashSHA1 $ Text.encodeUtf8 startUrl
```
endgame, line 82:
```suggestion
  let sha1 = show . Crypto.hashSHA1 $ Text.encodeUtf8 startUrl
```

lib/amazonka/src/Amazonka/Auth/SSO.hs
```diff
@@ -0,0 +1,111 @@
...
+  (MonadIO m, Foldable withAuth) =>
+  FilePath ->
+  Region ->
+  Text ->
+  Text ->
+  Env' withAuth ->
+  m Env
+fromSSO cache ssoRegion accountId roleName env = do
+  auth <- liftIO $ fetchAuthInBackground getCredentials
+  pure $ env {envAuth = Identity auth}
```
endgame, line 48: Question: It's correct to not update region on the returned environment, right? This is because the SSO region and the region the library user wants might be different. Even though we respect `AWS_REGION` somewhat, we might also have the region we want in the input `Env' withAuth`.

If I've got things right, would you mind adding a comment here? This seems a bit subtle.

lib/amazonka/src/Amazonka/Auth/SSO.hs
```diff
@@ -0,0 +1,111 @@
...
+import qualified Network.HTTP.Client as Client
+
+data Cache = Cache
+  { accessToken :: Text,
+    region :: Region
+  }
+  deriving stock (Generic)
+  deriving anyclass (FromJSON)
+
+fromSSO ::
```
endgame, line 38: Please re-export this function from `Amazonka.Auth`, and add a haddock.

endgame (changes requested):
All right, this is looking much better, thanks for your efforts. There are a lot of comments in this review, but most of them should be easy to address.

lib/amazonka/src/Amazonka/Auth/ConfigFile.hs
```diff
@@ -224,6 +238,9 @@ data ConfigProfile
...
+    -- @sso_role_name@.
+    SSO Text Region Text Text
```
endgame, line 243: Maybe we call this `AssumeRoleViaSSO` for consistency with the other constructors?

lib/amazonka/src/Amazonka/Auth/SSO.hs
```diff
@@ -0,0 +1,111 @@
...
+import qualified Control.Exception as Exception
+import Control.Exception.Lens (handling_, _IOException)
+import Control.Monad.Trans.Resource (runResourceT)
+import Data.Aeson (FromJSON, decodeFileStrict)
+import qualified Data.Text as Text
+import qualified Data.Text.Encoding as Text
+import Data.Time.Clock.POSIX (posixSecondsToUTCTime)
+import qualified Network.HTTP.Client as Client
+
+data Cache = Cache
```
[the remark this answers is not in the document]
pbrisbin: Yeah, originally I was OK with that for an internal type, it's actually "SSO.Cache" when you think about it, but the convention of open exports in this module damages that approach. I'll do `CachedAccessToken` though it is more than that: the fields are `{startUrl, region, accessToken, expires, expiresAt}`.

lib/amazonka/src/Amazonka/Auth/SSO.hs
pbrisbin: Sure, care to elaborate why?

At $work, we have a convention to always make something that _can_ be top-level top-level -- we find it helps with thinking about parts in isolation (vs lumping the complexity of this function into the other one as a single "thing"), results in more type signatures (being now non-optional), and will very often end up being re-used (my need to move `configFileRelative` to accomplish this PR muddies its diff when it wouldn't need to had this been followed).

Not pushing back at all, just curious to hear the other side.

lib/amazonka/src/Amazonka/Auth/SSO.hs
pbrisbin: TIL! Apparently, I have **a lot** of my own code to go and update now :grimacing:

lib/amazonka/src/Amazonka/Auth/ConfigFile.hs
pbrisbin: I see by your second comment you mean `cache`. Done.

lib/amazonka/src/Amazonka/Auth/SSO.hs
pbrisbin: It's up to you. I'd need some hand-holding, but I'm happy to do it if you'd like.

lib/amazonka/src/Amazonka/Auth/SSO.hs
pbrisbin: Renamed to `sendUnsigned` going to save the rest for later.

lib/amazonka/src/Amazonka/Auth/SSO.hs
pbrisbin: I added a comment on the use of `ssoRegion` that I think addresses this. Going to optimistically Resolve conversation.

pbrisbin:
OK, I think everything is resolved with the exception of the service definition.

For the CHANGELOG, can you offer some guidance? I'm not sure if this is a "Major Change" or "Other Changes".

I'd also like to cleanup the git History before merge, to isolate the feature from other pre/re-factors -- unless you happen to have a "Squash" policy (or some other preference). Do you?

lib/amazonka/src/Amazonka/Auth/SSO.hs
endgame: Agreed that it can be done separately. Created #758 - do you want to own this? If not, it's small enough that I can get it done.

lib/amazonka/src/Amazonka/Auth/SSO.hs
endgame: OK, if you're willing to wait for this, then let's do it properly.

endgame: #759 - I'll merge in a day or so in case anyone has comments, then regenerate.

lib/amazonka/src/Amazonka/Auth/ConfigFile.hs
endgame: Yeah I think it makes sense conceptually, even though it's not an `sts:AssumeRole` or `sts:AssumeRoleWithWebIdentity` call.

lib/amazonka/src/Amazonka/Auth/ConfigFile.hs
```diff
@@ -172,7 +178,8 @@ parseConfigProfile profile = parseProfile <&> \p -> (p, parseRegion)
         [ explicitKey,
           assumeRoleFromProfile,
           assumeRoleFromCredentialSource,
-          assumeRoleWithWebIdentity
+          assumeRoleWithWebIdentity,
+          sso
```
endgame, line 182: Nit: please rename this variable to `assumeRoleViaSSO` for consistency.

lib/amazonka/src/Amazonka/Auth/ConfigFile.hs
```diff
@@ -224,6 +238,9 @@ data ConfigProfile
   | -- | Recognizes @role_arn@, @role_session_name@, and
     -- @web_identity_token_file@.
     AssumeRoleWithWebIdentity Text (Maybe Text) FilePath
+  | -- | Recognizes @sso_start_url@, @sso_region@, @sso_account_id@, and
+    -- @sso_role_name@.
```
endgame, line 242: I appreciate your diligence maintaining haddock conventions.

lib/amazonka/src/Amazonka/Auth/SSO.hs
```diff
@@ -0,0 +1,126 @@
...
+  a ->
+  m (AWSResponse a)
+sendUnsigned env req = do
+  eResponse <- retryRequest env req
+  either (liftIO . Exception.throwIO) (pure . Client.responseBody) eResponse
+
+-- | Return the cached token file for a given @sso_start_url@
+--
+-- Matches
+-- <https://github.com/boto/botocore/blob/c02f3561f56085b8a3f98501d25b9857b916c10e/botocore/utils.py#L2596-L2597>
```
endgame, line 97:
```suggestion
-- [botocore](https://github.com/boto/botocore/blob/c02f3561f56085b8a3f98501d25b9857b916c10e/botocore/utils.py#L2596-L2597),
-- so that we find tokens produced by @aws sso login@.
```

endgame (changes requested):
Thanks very much. One last round of nitpicking, otherwise I'm pretty happy with how this looks. I'm also happy with your CHANGELOG entries, so thanks for updating the big table.

Here's how I see the next few days going:

1. Leave #759 open for a day or two in case someone objects.
2. (Me) Merge #759 and regenerate service bindings.
3. (You) While waiting for #759, rebase/rearrange commits until you're happy. I usually merge rather than squash in this repository, so if you can make the history look nice that'd be cool. But I also wouldn't lose sleep over it.
4. (You) Rebase on top of `main` once the regenerated `amazonka-sso` is available, so you're not coding against an errant `Maybe`; also remove the temporary stack files.
5. (Me) Merge this PR.

lib/amazonka/src/Amazonka/Auth/SSO.hs
pbrisbin: Whoa, when did Haddock start supporting markdown for links? I had to look it up to confirm but you're totally right.

pbrisbin:
OK! I think this is done from my end. I will,

- Wait for #759 
- Rebase after that's merged

Then you can Approve/Merge.

lib/amazonka/src/Amazonka/Auth/SSO.hs
```diff
@@ -0,0 +1,127 @@
...
+data CachedAccessToken = CachedAccessToken
+  { startUrl :: Text,
+    region :: Region,
+    accessToken :: Sensitive Text,
+    expiresAt :: UTCTime
+  }
+  deriving stock (Show, Eq, Generic)
+  deriving anyclass (FromJSON)
+
+-- | Assume a role using an SSO Token
```
endgame, line 40:
```suggestion
-- | Assume a role using an SSO Token.
```

lib/amazonka/src/Amazonka/Auth/SSO.hs
```diff
@@ -0,0 +1,127 @@
...
+-- the token in the @sso/cache@ file is not expired. When it has, the user will
+-- need to @aws sso login@ again.
+--
+-- <https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sso.html>
+fromSSO ::
+  (MonadIO m, Foldable withAuth) =>
+  FilePath ->
+  Region ->
+  Text ->
+  Text ->
```
endgame, line 54:
```suggestion
  -- | Account ID
  Text ->
  -- | Role Name
  Text ->
```

lib/amazonka/src/Amazonka/Auth/SSO.hs
```diff
@@ -0,0 +1,127 @@
...
+    accessToken :: Sensitive Text,
+    expiresAt :: UTCTime
+  }
+  deriving stock (Show, Eq, Generic)
+  deriving anyclass (FromJSON)
+
+-- | Assume a role using an SSO Token
+--
+-- The user must have called @aws sso login@ and have appropriate @sso_@ values
+-- in their active profile, which are the arguments to this function. This
```
endgame, line 43: This haddock is actually describing the behaviour of `fromConfigFile`. Would you mind rewriting to say something like:

```
The user must have previously called @aws sso login@, and pass in the path to the cached token file, along with SSO region, account ID and role name. ('Amazonka.Auth.ConfigFile.fromFilePath' understands the @sso_@ variables used by the official AWS CLI and will call 'fromSSO' for you.) This function uses 'fetchAuthInBackground'...
```

endgame (approved):
Great, now we wait for the new version of service definitions. I don't expect anyone to object so I'll do them over the next couple of days.

endgame:
The situation with service definitions and the generator is more complex than I first anticipated. I'm going to merge this now, as removing the `Maybe` is an easy fix when I do a generator pass.

endgame:
Thanks again for your work and patience.

pbrisbin:
Thank you for shepherding this through, this'll make using our amazonka-based projects so much easier.

Issue #736, opened 2021-12-06
amazonka-wafv2: generate


endgame:
I'm capturing the services that we failed to generate in #701 as standalone issues because it was hard for me to find them. If you care about this service, please leave a comment or react so I can prioritise.

> wafv2: generator fails to correctly deal with recursive types.

Issue #730, opened 2021-12-06
amazonka-kendra: generate


endgame:
I'm capturing the services that we failed to generate in #701 as standalone issues because it was hard for me to find them. If you care about this service, please leave a comment or react so I can prioritise.

> The generator for kendra mysteriously fails to add any sort of lens prefix to fields.

PR #761, opened 2022-03-11
Gen hackery

endgame:
~~I had hoped to teach `gen` how to handle recursive shape definitions in general, but ran out of time and mainline project work re-starts from next week. This at least lets us regenerate everything, as previously `rds-data` failed to generate due to a three-deep recursive cycle.~~

This PR teaches `gen` to handle recursive shape definitions.

### DIFF
--- a/gen/src/Gen/Types/Config.hs
+++ b/gen/src/Gen/Types/Config.hs
@@ -6,7 +6,7 @@ import qualified Control.Lens as Lens
 import Data.Aeson ((.!=), (.:), (.:?), (.=))
 import qualified Data.Aeson as Aeson
 import qualified Data.List as List
-import Data.Ord (Down (..), comparing)
+import Data.Ord (Down (..))
 import qualified Data.Text as Text
 import qualified Data.Time as Time
 import Gen.Prelude
@@ -166,7 +166,7 @@ instance ToJSON Library where
             "otherModules" .= List.sort (l ^. otherModules),
             "extraDependencies" .= List.sort (l ^. extraDependencies),
             "operations"
-              .= List.sortBy (comparing _opName) (l ^.. operations . Lens.each),
+              .= List.sortOn _opName (l ^.. operations . Lens.each),
             "shapes" .= List.sort (l ^.. shapes . Lens.each),
             "waiters" .= (l ^.. waiters . Lens.each)
           ]
--- a/gen/src/Gen/Types/Config.hs
+++ b/gen/src/Gen/Types/Config.hs
@@ -6,7 +6,7 @@ import qualified Control.Lens as Lens
 import Data.Aeson ((.!=), (.:), (.:?), (.=))
 import qualified Data.Aeson as Aeson
 import qualified Data.List as List
-import Data.Ord (Down (..))
+import Data.Ord (Down (..), comparing)
 import qualified Data.Text as Text
 import qualified Data.Time as Time
 import Gen.Prelude
@@ -165,7 +165,8 @@ instance ToJSON Library where
             "exposedModules" .= List.sort (l ^. exposedModules),
             "otherModules" .= List.sort (l ^. otherModules),
             "extraDependencies" .= List.sort (l ^. extraDependencies),
-            "operations" .= (l ^.. operations . Lens.each),
+            "operations"
+              .= List.sortBy (comparing _opName) (l ^.. operations . Lens.each),
             "shapes" .= List.sort (l ^.. shapes . Lens.each),
             "waiters" .= (l ^.. waiters . Lens.each)
           ]
--- a/gen/src/Gen/Types/Service.hs
+++ b/gen/src/Gen/Types/Service.hs
@@ -31,6 +31,7 @@ data Signature
   | V3HTTPS
   | V4
   | S3
+  | S3V4
   deriving (Eq, Show, Generic)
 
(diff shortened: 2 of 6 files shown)


endgame:
@brendanhay I tried generating a `Ptr` every time we see a type that we've seen before, basically by generalising the condition on the below line to <code>n \`elem\` seen</code>:

https://github.com/brendanhay/amazonka/blob/f73a957d05f64863e867cf39d0db260718f0fadd/gen/src/Gen/AST/Cofree.hs#L65

When I tried to generate bindings for `dynamodb` (it's a service I know reasonably well), it failed to generate lens prefixes on labels, like what happened when you tried to generate `kendra` bindings (reported at #730). This makes me think that there's some rewriting pass or something that doesn't add the prefixes correctly if it sees a `Ptr` instead of a proper recursive shape; do you have any insights?

Otherwise, I hope we can at least turn the handle on some of the new services, and get a big regenerate in sometime in the next few weeks.

RobertFischer:
We are going to get rid of the lens stuff, right?  So maybe just sit on
this until then?

endgame:
If you mean the PR as a whole, no: we need to be able to regen services.

If you mean the `Ptr` stuff, then I still think that's unwise: I think there's probably something else afoot that we ignore at our peril. AIUI, the eventual plan for lens is to avoid the `lens` dependency (so callers can use either `generic-lens` or `generic-optics`, but still provide VL lenses for data types under unambiguous names.

brendanhay:
Need to run this and poke around to rebuild my memory index as nothing concrete comes to mind off the top of my head.

I suspect your intuition is correct, ie. elaborate from the flat JSON parse result (shape) -> recursive cofree, then probably only rename the top-level without descending into the recursive shape to rename all references, or something similar stupid.

endgame:
That would be excellent. I wasn't able to grok that part of the generator, so if you get a chance to test my theory, I'd be very grateful.

endgame:
@brendanhay Starting next week, I should have a couple of work weeks where I can give amazonka some more attention. If you can get a chance to look at the generator before then, I'd really appreciate it. If not, then I intend to merge this to unblock things (we can't gen right now for some services we previously could; I suspect the change to hash functions in `hashable ^>=1.4`) and see what i can get done.

endgame:
@brendanhay Some friends and I found what we believe to be a subtle bug in the way the `seen` list was being tracked inside `elaborate`: when recursing into refs, the refs would add _themselves_ to the `seen` list. If the root object adds itself to `seen`, then it looks like we can unconditionally generate `Ptr` when we hit a shape that we've seen before. Because `seen` is cleared each time we explore a value in the input `HashMap`, we guarantee that we never memoise a `Ptr` and that the top level structures are always `StructF`, so generation still proceeds.

It would be great if you could review b3985cd347877f4dbaf06c65bf4049a5ea1eacc8 in particular, but I'm happy to merge and get back to fixing services if you're busy.

endgame:
> When I tried to generate bindings for `dynamodb` (it's a service I know reasonably well), it failed to generate lens prefixes on labels, like what happened when you tried to generate `kendra` bindings (reported at #730)...

This appears to be fixed, too.

gen/src/Gen/Types/Config.hs
```diff
@@ -165,7 +165,8 @@ instance ToJSON Library where
             "exposedModules" .= List.sort (l ^. exposedModules),
             "otherModules" .= List.sort (l ^. otherModules),
             "extraDependencies" .= List.sort (l ^. extraDependencies),
-            "operations" .= (l ^.. operations . Lens.each),
+            "operations"
+              .= List.sortBy (comparing _opName) (l ^.. operations . Lens.each),
```
endgame, line 169: @axman6 says `List.sortOn _opName`

endgame:
For some libraries with recursive definitions (e.g., `wafv2`; #736), this version of `gen` successfully generates code, but the generated library will not build thanks to a module import cycle. Should we generate `.hs-boot` files for any shapes where we had to create a `Ptr`?

endgame:
Notes for myself, for when I next get a chance to hack on this:

I think also it might not be too bad:
* when writing out a type file, if the shape you need to import is a `Ptr`, do an `import {-# SOURCE #-}`
* when building the tree of files to write out, generate `.hs-boot` for files which are referenced at least once by `Ptr`

gen/src/Gen/Types/Service.hs
```diff
@@ -31,6 +31,7 @@ data Signature
   | V3HTTPS
   | V4
   | S3
+  | S3V4
```
endgame, line 34: `s3control`, which is in a future botocore, will need this.

endgame (commented):
@brendanhay The branch in its current form can re-generate all configured services in the current botocore pin, and they all compiled successfully on my machine. Do you think we should merge this?

Future work: when we advance the botocore pin and configure new services, we may have problems with import cycles - wafv2 is the only one I'm aware of. If it's the only one, I think a manual fixup is probably the way to go: suppress generation of its `Statement` shape, and replace it with a proper sum type a la #724 .

I hit other problems on an updated botocore pin, but I think we do them one at a time:

1. merge this
2. ship config fixups that have current issues/PRs
3. regen all services
4. update botocore, regen all services, fixing up as necessary
5. start configuring new services.

Thoughts? If you're busy, I'll probably merge in the back half of next week.

PR #769, opened 2022-04-27
Send unsigned

### LINKED ISSUE #758: Expose `sendUnsigned`; move sending functions to `Amazonka.Send` and re-export
In the current layout, a function like `sendUnsigned` (if you can forgive the bikeshedding/renaming) would live in the root `Amazonka` module, but if you do this you will create an import loop. Maybe create an `Amazonka.Send` module which can hold `send`/`sendEither`/`paginate`/`paginateEither`/`sendUnsigned`/`sendUnsignedEither` and ensure these functions are re-exported by the `Amazonka` module?

_Originally posted by @endgame in https://github.com/brendanhay/amazonka/pull/757#discussion_r803356026_

endgame:
This came up when @pbrisbin was adding SSO support - it would be nice to have an easy interface to send unsigned requests, in case AWS stand up more exotic auth methods.

I also refactored STS and SSO to use it.

- [x] Use `EnvNoAuth` in public type signatures
- [x] Remove `Foldable withAuth` from `fromSSO`
- [x] Remove `Foldable withAuth` from `fromWebIdentity`

Closes #758


endgame:
@pbrisbin We talked about doing this when you wrote up the SSO stuff - do you have time to take a look?

lib/amazonka/src/Amazonka/Send.hs
```diff
@@ -0,0 +1,156 @@
...
+-- See 'sendUnsigned'.
+sendUnsignedEither ::
+  ( MonadResource m,
+    AWSRequest a
+  ) =>
+  Env' Proxy ->
+  a ->
+  m (Either Error (AWSResponse a))
+sendUnsignedEither env =
+  fmap (second Client.responseBody) . HTTP.retryRequest env
```
pbrisbin, line 69: I didn't realize these were the same concrete implementation and it's just a type-level restriction. Would it be valuable to DRY?

```hs
sendEither ::
  ( MonadResource m,
    AWSRequest a
  ) =>
  Env ->
  a ->
  m (Either Error (AWSResponse a))
sendEither = sendSomethingEither

sendUnsignedEither ::
  ( MonadResource m,
    AWSRequest a
  ) =>
  Env' Proxy ->
  a ->
  m (Either Error (AWSResponse a))
sendUnsignedEither = sendSomethingEither

sendSomethingEither ::
  ( MonadResource m,
    AWSRequest a
  ) =>
  Env' withAuth ->
  a ->
  m (Either Error (AWSResponse a))
sendSomethingEither env =
  fmap (second Client.responseBody) . HTTP.retryRequest env
```

The expressions pretty simple, so I don't see a huge risk in fixing a bug in one and forgetting the other.

lib/amazonka/src/Amazonka/Auth/SSO.hs
```diff
@@ -68,33 +67,21 @@ fromSSO cachedTokenFile ssoRegion accountId roleName env = do
       -- The Region you SSO through may differ from the Region you intend to
       -- interact with after. The former is handled here, the latter is taken
       -- care of later, in ConfigFile.
-      let ssoEnv = env {envRegion = ssoRegion}
-
-      resp <-
-        runResourceT $
-          sendUnsigned ssoEnv $
-            SSO.newGetRoleCredentials roleName accountId (fromSensitive accessToken)
+      let ssoEnv = env {envRegion = ssoRegion, envAuth = Proxy}
```
pbrisbin, line 70: So I see here we are accepting any `Env' withAuth` and so we `envAuth = Proxy` to be able to use `sendUnsigned` at `Env' Proxy`. Could that clobbering be risky in that someone might mis-use `fromSSO` with an authorized `Env` unintentionally? Or is that a feature?

I'm wondering if moving the type-restriction up as necessary, e.g.

```hs
fromSSO ::
  (MonadIO m, Foldable withAuth) =>
  FilePath ->
  Region ->
  -- | Account ID
  Text ->
  -- | Role Name
  Text ->
  Env' Proxy ->
  m Env
```

Would be better.

lib/amazonka/BUILD.bazel
```diff
@@ -12,12 +12,14 @@ haskell_library(
         "src/Amazonka/Auth/Exception.hs",
         "src/Amazonka/Auth/InstanceProfile.hs",
         "src/Amazonka/Auth/Keys.hs",
+        "src/Amazonka/Auth/SSO.hs",
```
pbrisbin, line 15: Oops, did I forget some Bazel-related bookkeeping when doing the SSO stuff?

pbrisbin (approved):

lib/amazonka/src/Amazonka/Send.hs
endgame: I didn't notice either until I put them all next to each other. I think that the giant type signature makes DRYing up the expression pretty unappealing. I don't think these functions have changed or will change that much, so I'm happy to leave as-is.

lib/amazonka/BUILD.bazel
endgame: Yeah, and I didn't notice either until I re-ran `scripts/update-gazelle`, which generates `BUILD.bazel` from cabal files.

lib/amazonka/src/Amazonka/Auth/SSO.hs
endgame: I thought about it after reading your comment, but I don't think so. When we provide `Env' withAuth` in a type signature, we're saying to the caller "I don't care whether or not this env has auth information", which I think is a pretty clear signal that it's not going to be used. But `Amazonka.HTTP.httpRequest` uses the `Foldable` instance to get auth information if it is present, so I have to put `Env' Proxy` into the type signature of `sendUnsigned`, otherwise it would sign requests if auth information was present - not really what we want to do.

So that forces `sendUnsigned` to take an `Env' Proxy` argument, but why not push that explicit `Proxy` into the type signature of functions like `fromSSO`? My main fear is that it makes it needlessly harder to build credential chains: consider some code which might try to `sts:AssumeRoleWithWebIdentity` (on ECS) or `sts:AssumeRole` (when a dev is working locally). The first method would need an `Env' Proxy`, and the second would need an `Env`, and they'd no longer unify. I think it's better to just ignore the auth information on any `Env'` we are given.

It does remind me of a few tidy-ups that should happen as part of this PR: we have an `EnvNoAuth` type alias which should be used, and `fromSSO` and `fromWebIdentity` should be able to drop the `Foldable withAuth` constraints.

lib/amazonka/src/Amazonka/Send.hs
```diff
@@ -0,0 +1,156 @@
...
+  ) =>
+  Env ->
+  a ->
+  m (AWSResponse a)
+send env =
+  sendEither env >=> hoistEither
+
+-- | Make a request without signing it. You will almost never need to
+-- do this, but some authentication methods
+-- (e.g. @sts:AssumeRoleWithWebIentity@ and @sso:GetRoleCredentials@)
```
Tristano8, line 54: Normally this wouldn't matter but it might affect searchability

```suggestion
-- (e.g. @sts:AssumeRoleWithWebIdentity@ and @sso:GetRoleCredentials@)
```

endgame: Good catch.

endgame:
@pbrisbin What if, instead of requiring `EnvNoAuth` in `sendUnsigned`, I do the `env {envAuth = Proxy}` override inside `sendUnsigned` instead of the SSO/STS functions? This would give a more consistent interface and involve less type magic.

I think I will do this tomorrow.

pbrisbin:
That does seem reasonable to me. I think seeing `sendUnsigned :: Env' withAuth` better indicates "I'm going to ignore whatever auth there is", but you really have to understand how this library works to hear what it's saying, so :shrug:

endgame:
I'm much happier with how this turned out. Thanks for your input.

PR #772, opened 2022-05-09
provide license file for `examples.cabal`

peterbecich:
The license file referred to here
https://github.com/brendanhay/amazonka/blob/4fd7f3866342ba92b8deef719409ef4eba3270dd/examples/examples.cabal#L7
was missing.

This PR copies it from `lib/amazonka/`.


endgame (approved):
Matches the MPL-2.0 SPDX identifier, so I'm happy with it.

Issue #752, opened 2022-02-03
Remove underscores from amazonka/amazonka-core record fields


endgame:
In general we are moving to supporting generic-lens/generic-optics and allowing the users to choose the optic lib that they prefer. So make the hand-written records in the two main amazonka libraries have unprefixed field names.

Should go in before 2.0 just because it's another minor breaking change.

bsaul:
I see underscores in record field names in:

* `IdentityDocument` in `Amazonka.EC2.Metadata` 
*  several records in `Amazonka.Types`
* `Waiter` in `Amazonka.Waiter`

Are these the ones to address?

endgame:
Yes, that's right. There shouldn't be any in the generated service bindings, so you'll only be looking for ones in `lib/amazonka` and `lib/amazonka-core`. You should hopefully also be able to drop any prefixes from them, like the `_serializeError` prefixes in https://github.com/brendanhay/amazonka/blob/4be74644af107d67bc19230826b64595095384b8/lib/amazonka-core/src/Amazonka/Types.hs#L241-L246

I hope this means that the manual lenses like `serializeAbbrev` can also go away too, but this may get a bit gnarly. Not sure about the `AsError` typeclass either.

bsaul:
Opened up #778 to start work towards this.

PR #778, opened 2022-05-27
Remove underscores from record fields

bsaul:
Currently, this is just a start towards #752. I've focused mainly on `amazonka-core` for now.

I'm just opening this up so that others can jump on if you have time, as I'll be out for a week.

I know the following functions are broken as of b50c24800d7bb6b130e93fef89716d4f8449a7b3:

* `s3vhost` in Amazonka.Request
* `setEndpoint` in Amazonka.Endpoint

in both cases, my Friday afternoon lens-fu is just not happening.

NOTE: I added `generic-lens` as a dependency.

### DIFF
--- a/lib/amazonka-core/amazonka-core.cabal
+++ b/lib/amazonka-core/amazonka-core.cabal
@@ -108,6 +108,7 @@ library
     , containers            >=0.5     && < 0.7
     , cryptonite            >=0.25
     , deepseq               >=1.4
+    , generic-lens          >=2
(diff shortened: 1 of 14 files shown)


bsaul:
@endgame - so far I have [not removed prefixes](https://github.com/brendanhay/amazonka/issues/752#issuecomment-1138268190). I ran into duplicated field names and I felt it better for someone who knows what they're doing to address this.

PR #779, opened 2022-05-31
mtl-2.3 compatibility

ysangkok:
Before this PR, compilation on mtl-2.3 fails with

```
src/Amazonka/Error.hs:98:51: error:
    • No instance for (Alternative (Either String))
        arising from a use of ‘<|>’
    • In the third argument of ‘either’, namely
        ‘(h .# hAMZRequestId <|> h .# hAMZNRequestId)’
      In the expression:
        either
          (const Nothing) Just (h .# hAMZRequestId <|> h .# hAMZNRequestId)
      In an equation for ‘getRequestId’:
          getRequestId h
            = either
                (const Nothing) Just (h .# hAMZRequestId <|> h .# hAMZNRequestId)
   |
98 |   either (const Nothing) Just (h .# hAMZRequestId <|> h .# hAMZNRequestId)
```

The instance `Alternative (Either String)` makes little sense to me since I don't know how the instance can manufacture an `empty` value for the Alternative instance.

So in this PR, I try to avoid this instance.

This is untested, I'd appreciate advice on how best to test this.

### DIFF
--- a/lib/amazonka-core/src/Amazonka/Error.hs
+++ b/lib/amazonka-core/src/Amazonka/Error.hs
@@ -14,6 +14,7 @@ import Amazonka.Types
 import qualified Data.Aeson as Aeson
 import qualified Data.Aeson.Types as Aeson.Types
 import qualified Data.ByteString.Lazy as LBS
+import Data.Either (fromRight)
 import qualified Network.HTTP.Client as Client
 import Network.HTTP.Types.Status (Status (..))
 
--- a/lib/amazonka-core/src/Amazonka/Error.hs
+++ b/lib/amazonka-core/src/Amazonka/Error.hs
@@ -141,19 +141,15 @@ parseXMLError a s h bs = decodeError a s h bs (go <$> decodeXML bs)
   where
     go x =
       serviceError a s h
-      (code x)
-      (may' (firstElement "Message" x))
-      (may' (firstElement "RequestId" x) <|> may' (firstElement "RequestID" x))
+        (code x)
+        (may' (firstElement "Message" x))
+        (may' (firstElement "RequestId" x) <|> may' (firstElement "RequestID" x))
 
-    code x
-      | Right y <- firstElement "Code" x >>= parseXML = y
-      | otherwise = root
+    code x = fromRight root $ parseXML =<< firstElement "Code" x
 
     root = newErrorCode <$> rootElementName bs
 
-    may' x
-      | Right y <- x >>= parseXML = Just y
-      | otherwise = Nothing
+    may' x = either (const Nothing) Just $ parseXML =<< x
 
 parseRESTError ::
   Abbrev ->
@@ -174,7 +170,7 @@ decodeError ::
 decodeError a s h bs e
   | LBS.null bs = parseRESTError a s h bs
   | otherwise =
-    either
-      (SerializeError . SerializeError' a s (Just bs))
-      ServiceError
-      e
+      either
+        (SerializeError . SerializeError' a s (Just bs))
+        ServiceError
+        e
--- a/lib/amazonka-core/src/Amazonka/Error.hs
+++ b/lib/amazonka-core/src/Amazonka/Error.hs
@@ -94,8 +94,10 @@ serviceError a s h c m r =
   ServiceError' a s h (fromMaybe (getErrorCode s h) c) m (r <|> getRequestId h)
 
 getRequestId :: [Header] -> Maybe RequestId
-getRequestId h =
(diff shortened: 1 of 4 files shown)


lib/amazonka-core/src/Amazonka/Error.hs
```diff
@@ -135,22 +137,23 @@ parseXMLError ::
...
+parseXMLError a s h bs = decodeError a s h bs (go <$> decodeXML bs)
   where
     go x =
       serviceError a s h
-        <$> code x
-        <*> may' (firstElement "Message" x)
-        <*> may' (firstElement "RequestId" x <|> firstElement "RequestID" x)
+      (code x)
+      (may' (firstElement "Message" x))
+      (may' (firstElement "RequestId" x) <|> may' (firstElement "RequestID" x))
```
endgame, line 146: If you re-run `ormolu` over this file, does this indent remain? I suspect it shifts across.

lib/amazonka-core/src/Amazonka/Error.hs
```diff
@@ -135,22 +137,23 @@ parseXMLError ::
...
 
-    code x =
-      Just <$> (firstElement "Code" x >>= parseXML)
-        <|> return root
+    code x
+      | Right y <- firstElement "Code" x >>= parseXML = y
+      | otherwise = root
```
endgame, line 150: I find this style a little hard to read, I think because there's left-to-right data flow in the `firstElement "Code" x >>= parseXML` part, and also in the `= y` part, but it's interrupted by the pattern-guard, which is right-to-left. How about this?
```suggestion
    code x = fromRight root $ parseXML =<< firstElement "Code"
```
(`fromRight` is in `Data.Either` since `base-4.10.0.0`.)

lib/amazonka-core/src/Amazonka/Error.hs
```diff
@@ -135,22 +137,23 @@ parseXMLError ::
...
 
     root = newErrorCode <$> rootElementName bs
 
-    may' (Left _) = pure Nothing
-    may' (Right x) = Just <$> parseXML x
+    may' x
+      | Right y <- x >>= parseXML = Just y
+      | otherwise = Nothing
```
endgame, line 156: Similarly:
```suggestion
    may' x = either Nothing Just $ parseXML =<< x
```
(Unless there's a better tool in `Amazonka.Prelude` or something.)

endgame (commented):
Thanks for this. I have some requests, beyond the inline comments:

* Please add `mtl >=2.2 && <2.4` to the bounds on `amazonka-core.cabal`.
* Please add an entry to `lib/amazonka/CHANGELOG.md`, which represents the project as a whole.

As for testing, the best thing I can think of is to build some packages (or maybe some of the example programs) with `--constraint 'mtl >=2.3'` and see if it looks good.

lib/amazonka-core/src/Amazonka/Error.hs
ysangkok: It does indeed, as shown in [e71414a](https://github.com/brendanhay/amazonka/pull/779/commits/e71414a8d267e8e40d3056096c0457fce8866d1f)

lib/amazonka-core/src/Amazonka/Prelude.hs
```diff
@@ -47,6 +47,7 @@ import qualified Data.ByteString.Builder as ByteString.Builder
 import qualified Data.ByteString.Lazy as ByteString.Lazy
 import Data.CaseInsensitive as Export (CI)
 import Data.Coerce as Export (Coercible, coerce)
+import Data.Either as Export (fromRight)
```
endgame, line 50: I think import this only in the file that needs it. Feels weird to have only `fromRight` in the prelude.

endgame (commented):
I have another request. How goes your plans to test this?

ysangkok:
Testing is going well, `cabal test --constraint='mtl>=2.3'` is passing with no issues. And I can build the examples.

endgame (approved):
I am happy with this. Ping me when you're satisfied with your testing and I will merge. Thanks very much.

endgame:
@ysangkok is there anything else you want to test before I merge?

ysangkok:
@endgame I was hoping that somebody would get the chance to test in a 'staging' first. Sadly at my work we don't infrastructure for the project we use Amazonka for, so I would have to test it in production, which is kinda dangerous. I think it looks fairly safe since the worst case seems to be that it passes a Nothing to `serviceError`, in place of making all of `go` a Nothing, which doesn't sound too dangerous, but I haven't looked into it.

PR #783, opened 2022-06-10
Change threadDelay for credential refreshing

### LINKED ISSUE #747: Timing of background refreshing of credentials
@Axman6 flags a potential issue with the way we refresh credentials - "delay until expiry, then fetch new crews, then delay for that time again", which risks a small window of time where we are holding expired credentials and the refresh thread hasn't yet got us new ones.

I suspect nobody has noticed this, as if it's a problem it will manifest in the form of very occasional AWS failures, only in process that run long enough for their tokens to expire.

It may be worth sleeping for halfway until credential expiry, then refreshing.

_Originally posted by @axman6 in https://github.com/brendanhay/amazonka/pull/746#discussion_r781717976_

Tristano8:
Closes #747


lib/amazonka/src/Amazonka/Auth/Background.hs
```diff
@@ -53,7 +53,9 @@ fetchAuthInBackground menv =
     loop :: IO AuthEnv -> Weak (IORef AuthEnv) -> ThreadId -> ISO8601 -> IO ()
     loop ma w p x = do
       next <- diff x <$> Time.getCurrentTime
```
endgame, line 55: Possibly better name:
```suggestion
      untilExpiry <- diff x <$> Time.getCurrentTime
```

lib/amazonka/src/Amazonka/Auth/Background.hs
```diff
@@ -53,7 +53,9 @@ fetchAuthInBackground menv =
...
-      Concurrent.threadDelay next
+      -- | Refresh the token conservatively -- within half its lifespan --
+      -- | to account for execution time of the IO AuthEnv action
```
endgame, line 57: Please don't use haddock comments here.
```suggestion
      -- Refresh the token conservatively -- within half its lifespan --
      -- to account for execution time of the refresh action
```

lib/amazonka/src/Amazonka/Auth/Background.hs
```diff
@@ -53,7 +53,9 @@ fetchAuthInBackground menv =
...
+      Concurrent.threadDelay (next `div` 2)
```
endgame, line 58:
```suggestion
      Concurrent.threadDelay $ untilExpiry `div` 2
```

endgame (changes requested):
Thanks for doing this. Please also add an entry in `lib/amazonka/CHANGELOG.md`.

lib/amazonka/CHANGELOG.md
```diff
@@ -51,6 +51,11 @@ Released: **?**, Compare: [2.0.0-rc1](https://github.com/brendanhay/amazonka/com
 
 ### Fixed
 
+- Background credential refreshing now uses half the token's lifespan
+as a thread delay to avoid potential concurrency issues.
+[\#747](https://github.com/brendanhay/amazonka/pull/783)
+- Lenses for the `Env'` data type have been manually provided, as `generic-lens` cannot always derive them when required.
+[\#727](https://github.com/brendanhay/amazonka/pull/727)
```
Tristano8, line 58:
```suggestion
```

PR #785, opened 2022-06-14
Secure token expiry: Wait 80% of expiry time before refresh

smatting:


smatting:
Sorry, I accidentally created this PR

PR #788, opened 2022-06-16
encode Canonical URI paths twice

### LINKED ISSUE #529: SignV4 Canonical URI paths are only encoded once. Should be twice
Per [AWS SignV4 docs](https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html)

> Normalize URI paths according to RFC 3986. Remove redundant and relative path components. Each path segment must be URI-encoded twice (except for Amazon S3 which only gets URI-encoded once). 

Most URIs have nothing to encode, which is why it usually works. However, when there is, the signature will be invalid.

As an example from [API Gateway @connections](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-how-to-call-websocket-api-connections.html)
The Endpoint is
```
https://{prefix}.execute-api.us-east-1.amazonaws.com/{stage}/@connections/{connection_id}
https://aabbccddee.execute-api.us-east-1.amazonaws.com/production/@connections/R0oXAdfD0kwCH6w=
```
The URI needs to be encode once for the request
```
https://aabbccddee.execute-api.us-east-1.amazonaws.com/prod/%40connections/R0oXAdfD0kwCH6w%3D
```
And then again for the Canonical URI
```
https://aabbccddee.execute-api.us-east-1.amazonaws.com/prod/%2540connections/R0oXAdfD0kwCH6w%253D
```

stevemao:
https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html

fixes #529

### DIFF
--- a/lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs
+++ b/lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs
@@ -241,6 +241,9 @@ credentialScope s e t =
       "aws4_request"
     ]
 
+encode :: ByteString -> ByteString
+encode = toBS . escapePath . rawPath
+
 canonicalRequest ::
   Method ->
   Path ->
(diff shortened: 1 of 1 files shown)


stevemao:
Thanks for the review. I'll get back to this when I have time.

PR #793, opened 2022-06-24
Regenerate services 2022-06-24

### LINKED ISSUE #765: Pinpoint toJSON methods are being generated incorrectly
Hi, we've had our own internal fork of the old Amazonka. It had a bug where the toJSON for various calls were incorrect - it's adding an extra key when it should be flat). It looks like 2.0 has the same issue.

For example:
* https://github.com/brendanhay/amazonka/blob/29e4888f847797533c968a855804990ecf8660dd/lib/services/amazonka-pinpoint/gen/Amazonka/Pinpoint/UpdateEndpoint.hs#L147
* https://github.com/brendanhay/amazonka/blob/29e4888f847797533c968a855804990ecf8660dd/lib/services/amazonka-pinpoint/gen/Amazonka/Pinpoint/UpdateSegment.hs#L145
* https://github.com/brendanhay/amazonka/blob/29e4888f847797533c968a855804990ecf8660dd/lib/services/amazonka-pinpoint/gen/Amazonka/Pinpoint/SendMessages.hs#L124

```instance Core.ToJSON UpdateEndpoint where
  toJSON UpdateEndpoint' {..} =
    Core.object
      ( Prelude.catMaybes
          [ Prelude.Just
              ("EndpointRequest" Core..= endpointRequest)
          ]
      )
```

The "EndpointRequest" should be removed (something like):

```
instance Core.ToJSON UpdateEndpoint where
  toJSON UpdateEndpoint' {..} = Core.toJSON endpointRequest
```

The error you get back from AWS (the pertinent bit is "Unrecognized field "EndpointRequest"):
```
Left (ServiceError (ServiceError' {_serviceErrorAbbrev = Abbrev {fromAbbrev = "Pinpoint"}, _serviceErrorStatus = Status {statusCode = 400, statusMessage = "Bad Request"}, _serviceErrorHeaders = [("Content-Type","application/json"),("Content-Length","260"),("Connection","keep-alive"),("Date","Tue, 05 Apr 2022 04:36:40 GMT"),("x-amzn-RequestId","21d60202-29e8-4df2-8f0f-2ed5f7504e95"),("Access-Control-Allow-Origin","*"),("x-amzn-ErrorType","BadRequestException"),("x-amz-apigw-id","QFwV0G-3SwMFXNg="),("Cache-Control","no-store"),("X-Amzn-Trace-Id","Root=1-624bc758-05e2608a66a7952a043beb65"),("X-Cache","Error from cloudfront"),("Via","1.1 4cc1cef5b7e8e0c2a1020c88ce286fbc.cloudfront.net (CloudFront)"),("X-Amz-Cf-Pop","SYD4-C1"),("X-Amz-Cf-Id","OjLJ0IXAhU_DZywSM7-7V1FaAlzVWCzNrfNSDGJdHnDN3ELG7TTwkQ==")], _serviceErrorCode = ErrorCode "BadRequest", _serviceErrorMessage = Just (ErrorMessage {fromErrorMessage = "Unrecognized field \"EndpointRequest\" (11 known properties: \"Metrics\", \"Address\", \"Location\", \"EndpointStatus\", \"RequestId\", \"OptOut\", \"ChannelType\", \"Demographic\", \"Attributes\", \"User\", \"EffectiveDate\"])\nLine 1, column 21"}), _serviceErrorRequestId = Just (RequestId {fromRequestId = "21d60202-29e8-4df2-8f0f-2ed5f7504e95"})}))
```

You can see how they're supposed to be structured:

* [Endpoint Schemas](https://docs.aws.amazon.com/pinpoint/latest/apireference/apps-application-id-endpoints-endpoint-id.html#apps-application-id-endpoints-endpoint-id-schemas)
* [Segment Schemas](https://docs.aws.amazon.com/pinpoint/latest/apireference/apps-application-id-segments-segment-id.html#apps-application-id-segments-segment-id-schemas)
* [Messages Schema](https://docs.aws.amazon.com/pinpoint/latest/apireference/apps-application-id-messages.html#apps-application-id-messages-schemas)

endgame:
It's about time we did one of these, as part of the plan to catch up to recent AWS service definitions. This pulls in various recent generator/service-definition changes:

* More things are now sorted in the service definitions so we shouldn't have as much churn in the generated output.
* `amazonka-pinpoint` should work properly; @newmana please check.
  - `amazonka-glacier` changed in similar ways, which I think means it now works but I don't know if we have any users.
* `amazonka-sso` copies `amazonka-sts`'s style of `typeOverrides` in the service configs to splice in `amazonka-core` types for Access Key ID, Secret Access Key, and Session Token fields. @pbrisbin please check that I haven't broken SSO for you.
* `amazonka-sts` declares the `Credentials` field `required` in all `AssumeRole*` responses.

The only interesting commits are:
- https://github.com/brendanhay/amazonka/commit/e4bed95ab1ac9e8be2b2b7881550006403228859 : let the generator script proceed to committing
- https://github.com/brendanhay/amazonka/commit/1566947a7eae0eacb1afbfbe047557eee13eca96 : support the changed STS/SSO bindings.

Once this is done, the next steps will be to advance to the latest botocore pin, generate again, and then start configuring missing service definitions.

Fixes #765


cdaloisio (commented):
The changes called out in those specific commits look OK to me @JackKelly-Bellroy. 

I'm not familiar with the Github actions in this project. What needs to be updated to make sure the `build and test` step is passing?

endgame:
There's a missing API key in the workflow runs, possibly because the PR branch is coming from a fork. I don't have enough permissions to play around with the config here, so I think it's a @brendanhay thing to fix that.

endgame:
Thanks @pbrisbin. It would also be good if @K0Te could report that `sts:AssumeRoleWithWebIdentity` works. But we have stuff deployed at work that's running off this branch using other STS functions so I'm happy to merge later this week.

K0Te:
Hi @endgame ,

I've just migrated my project to this PR branch, will test it tomorrow and report any issues. Project uses the following services: `amazonka-lambda`, `amazonka-sts`, `amazonka-s3`, `amazonka-cloudfront`, `amazonka-route53`, `amazonka-sqs`, `amazonka-ssm`.

I've used a patched 1.6.1 version before, so there were a lot of code changes.
Migration was smooth, except for a few cases where a generic lens for a field is different from a regular lens. For  example: `Amazonka.Lambda.Lens.functionCode_zipFile`, `Amazonka.SSM.Lens.parameter_value`.
This is probably a design feature, not a bug, right?

endgame:
Thanks for the good news @K0Te. I hadn't noticed that difference before - looks like the generator is smart enough to unwrap through things like `Sensitive` and `Base64`, and if you use a generic lens you'll have to use the `_Sensitive` and `_Base64` isos yourself (they should be exported by `Amazonka`, hopefully.

endgame:
@brendanhay that might be a good reason not to deprecate generated lenses?

brendanhay:
Ideally we want to continue to improve the codegen enough that it inserts the equivalent of `Sensitive` and `Base64` newtypes at de/serialisation points - ie. `o .: "field" <&> fromSensitive`, so the user never has to deal with un/wrapping said newtypes.

K0Te:
@endgame
I've tested what I could - everything seems to work correctly, including `sts:AssumeRoleWithWebIdentity`.

PR #794, opened 2022-06-29
amazonka-core: fix body length calculation in hashedFileRange

### LINKED ISSUE #548: `hashedFileRange` results in `WrongRequestBodyStreamSize`
While using `amazonka-s3-streaming` I started getting an error:
```
TransportError - WrongRequestBodyStreamSize 43626199 8388608
```
I could be wrong, but I suspect the error happens in `amazonka` itself cause `HashedStream` keeps the full file size rather than the stream size, causing `http-client` to explode:

```haskell
hashedFileRange :: MonadIO m
                => FilePath -- ^ The file path to read.
                -> Integer  -- ^ The byte offset at which to start reading.
                -> Integer  -- ^ The maximum number of bytes to read.
                -> m HashedBody
hashedFileRange path (Just -> offset) (Just -> len) =
    liftIO $ HashedStream
        <$> runResourceT (Conduit.sourceFileRange path offset len `connect` sinkSHA256)
        <*> getFileSize path -- < the full file size
        <*> pure (Conduit.sourceFileRange path offset len)
```
And here is the request creation:
```haskell
toRequestBody :: RqBody -> RequestBody
toRequestBody = \case
    Chunked x -> requestBodySourceChunked (_chunkedBody x)
    Hashed  x -> case x of
         HashedStream _ n f -> requestBodySource (fromIntegral n) f -- < problem here? n is the size of the whole file
         HashedBytes  _ b   -> RequestBodyBS b
```

endgame:
@axman6 

Fixes #548

### DIFF
--- a/lib/amazonka-core/src/Amazonka/Data/Body.hs
+++ b/lib/amazonka-core/src/Amazonka/Data/Body.hs
@@ -288,12 +288,14 @@ hashedFileRange ::
   -- | The maximum number of bytes to read.
   Integer ->
   m HashedBody
-hashedFileRange path (Just -> offset) (Just -> len) =
-  liftIO $
-    HashedStream
-      <$> runResourceT (Conduit.Binary.sourceFileRange path offset len `Conduit.connect` Crypto.sinkSHA256)
-      <*> getFileSize path
-      <*> pure (Conduit.Binary.sourceFileRange path offset len)
+hashedFileRange path offset len = do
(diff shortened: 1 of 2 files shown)


endgame:
@axman6 and I spent an evening shuffling large files between EC2 and S3 while fixing up amazonka-s3-streaming so we're pretty sure that this is good.

PR #799, opened 2022-07-27
Dynamodb writeitem

### LINKED ISSUE #798: amazonka-dynamodb: WriteRequest should be a sum type
Like #724, but for https://github.com/brendanhay/amazonka/blob/e4fdf4a33e357942a6bc5bb6bfc133f18b56fa0c/lib/services/amazonka-dynamodb/gen/Amazonka/DynamoDB/Types/WriteRequest.hs#L35-L41

endgame:
Closes #798 

@divarvel can you see if this is better for you?

- [ ] Changelog entry


Tristano8 (approved):

divarvel:
Thanks for the change! this makes things easier indeed.
The only thing i'm not sure of is the use of `Data.Map` (instead of `Data.Map.Strict` or a `HashMap`, which seem to be used a bit more in amazonka-adjacent code)

endgame:
Data.Map.Strict is probably a good idea both here and with AttributeValue, but I think that moving towards Map of some kind is a good long-term move because a) you avoid accidentally-quadratic behaviour when merging many maps, and b) aeson's internal type for KeyMap switched over to Map.

endgame:
Actually, it doesn't matter whether we import `Data.Map.Lazy` or `Data.Map.Strict` as we're only importing the type name.

PR #803, opened 2022-08-14
Allow aeson-2.1

ysangkok:
Tested using

    cabal test --constraint='aeson^>=2.1' -w ghc-9.2.4

### DIFF
--- a/lib/amazonka-core/amazonka-core.cabal
+++ b/lib/amazonka-core/amazonka-core.cabal
@@ -99,7 +99,7 @@ library
     Amazonka.Waiter
 
   build-depends:
-    , aeson                 ^>=1.5.0.0 || ^>= 2.0.0.0
+    , aeson                 ^>=1.5.0.0 || >=2.0 && <2.2
     , attoparsec            >=0.11.3
     , bytestring            >=0.10.8
     , case-insensitive      >=1.2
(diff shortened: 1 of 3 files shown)


endgame:
Is there a reason for writing `^>= 2.0.0.0 || ^>= 2.1.0.0` instead of `>= 2.0 && <2.2`? My reading of the [package description](https://cabal.readthedocs.io/en/3.4/cabal-package.html) page says they should be the same?

ysangkok:
Yes, the reason was that we'd wanna exclude a hypothetical version 1.6 of aeson. I was following the final suggestion of @tfausak on https://github.com/haskell/aeson/issues/881#issuecomment-944920664 . But it sounds like you'd prefer assuming that won't be released, so I have adjusted it.

endgame:
That's why it was `^>=1.5.0.0 || ^>=2.0.0.0`, but you wrote `^>=1.5.0.0 || ^>=2.0.0.0 || ^>=2.1.0.0`, which should be equivalent to `^>=1.5.0.0 || >=2.0 && <2.2`, by my understanding of how Cabal works. Am I wrong here?

ysangkok:
Oooh, now I see. I had misread your comment, sorry. I have now revised once again to use your suggested equation `^>=1.5.0.0 || >=2.0 && <2.2`.

endgame:
Thank you for that.

PR #809, opened 2022-08-25
New services 2022 08

endgame:
And fix the generator to copy the license into any new packages


endgame:
The only service I'm particularly concerned about is `amazonka-rbin`, because I had to change its abbrev from `RbIn` to `RBin`. But I can make test requests against it, so I think these new bindings are good enough to merge.

PR #811, opened 2022-09-02
amazonka: fixing sigv4 path encoding so that it happens once for s3 and twice for other services

mwu:
should address issue #529.


mwu:
targeted the wrong branch :/

PR #812, opened 2022-09-02
amazonka: fixing sigv4 path encoding so that it happens once for s3 and twice for other services

### LINKED ISSUE #529: SignV4 Canonical URI paths are only encoded once. Should be twice

mwu:
Should fix #529.

I did some manual smoke testing:
* EC2 describe instances calls
* lambda invokeFunction using all name forms (arn, partial arn, partial arn + alias, function name, name + alias)
* S3 getObject on keys including and excluding uri reserved chars


lib/amazonka-core/src/Amazonka/Data/Path.hs
```diff
@@ -74,6 +79,13 @@ escapePath :: Path a -> EscapedPath
 escapePath (Raw xs) = Encoded (map (URI.urlEncode True) xs)
 escapePath (Encoded xs) = Encoded xs
 
+-- | Escape a path twice for rare occasions like computing the sigv4 canonical path.
```
endgame, line 82:
```suggestion
-- | Escape a path twice. Used when computing the SigV4 canonical path.
```
Reason: it's nearly every AWS request, so better not to call it rare.

lib/amazonka-core/test/Test/Amazonka/Sign/V4/Base.hs
```diff
@@ -0,0 +1,113 @@
...
+        aReq
+          & requestPath .~ aPath
+          & requestService .~ svc
+          & requestBody .~ Hashed aBody
+
+  (meta, _) <- base (Tag (sha256Base16 aBody)) req <$> arbitrary <*> arbitrary <*> arbitrary
+  pure (aPath, metaPath meta, metaCanonicalPath meta)
+
+mkNormalizedPath :: Gen RawPath
+mkNormalizedPath = suchThat arbitrary noDots
```
endgame, line 111:
```suggestion
mkNormalizedPath = arbitrary `suchThat` noDots
```

lib/amazonka-core/test/Test/Amazonka/Sign/V4/Base.hs
```diff
@@ -0,0 +1,113 @@
...
+  (_, reqPath, canonicalPath) <- generate $ mkV4Paths (/= "S3") (pure key)
+  toBS reqPath @?= "/bar/baz"
+  untag canonicalPath @?= untag reqPath
+
+mkV4Paths :: (Abbrev -> Bool) -> Gen RawPath -> Gen (RawPath, Base.Path, CanonicalPath)
+mkV4Paths p genPath = do
+  aReq <- arbitrary :: Gen (Request ())
+  aPath <- genPath
+  aService <- arbitrary
+  anAbbrev <- suchThat arbitrary p
```
endgame, line 93: Can you modify this function to take a `Gen Abbrev`? Otherwise you risk blowing out test suite runtime by re-rolling the abbrev until it returns `S3`.

lib/amazonka/CHANGELOG.md
```diff
@@ -97,6 +97,8 @@ Released: **?**, Compare: [2.0.0-rc1](https://github.com/brendanhay/amazonka/com
 
 ### Fixed
 
+- `amazonka-core`: Fixed sigv4 canonical uri/path encoding issue
```
endgame, line 100:
```suggestion
- `amazonka-core`: Correctly double-url-encode request paths when computing V4 signatures
```

endgame (changes requested):
This is excellent work, thank you - I particularly appreciate the extensive test suite. I have a few small nits to pick but I look forward to merging this soon.

lib/amazonka-core/src/Amazonka/Data/Path.hs
```diff
@@ -59,6 +61,9 @@ type RawPath = Path 'NoEncoding
 
 type EscapedPath = Path 'Percent
 
+newtype TwiceEscapedPath = TwiceEscapedPath (Path 'Percent)
```
endgame, line 64: I think we want to add an additional `TwicePercent` constructor to `data Encoding`, so this can become a type alias for consistency.

endgame: Wait, nevermind. That doesn't make sense.

lib/amazonka-core/src/Amazonka/Data/Path.hs
endgame: Probably better to put the "Used in SigV4" on the `newtype` declaration so when someone clicks `TwiceEscapedPath` in haddock, they see it.

lib/amazonka-core/test/Test/Amazonka/Sign/V4/Base.hs
mwu: Ah, good call.

lib/amazonka-core/src/Amazonka/Data/Path.hs
mwu: I went through the exact same sequence of thoughts, haha.

lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs
```diff
@@ -30,6 +30,7 @@ data V4 = V4
     metaPath :: Path,
```
endgame, line 30: Can we remove this?

lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs
```diff
@@ -125,6 +126,8 @@ type NormalisedHeaders = Tag "normalised-headers" [(ByteString, ByteString)]
 
 type Method = Tag "method" ByteString
 
+type CanonicalPath = Tag "canonical-path" ByteString
+
 type Path = Tag "path" ByteString
```
endgame, line 131: Can we remove this?

lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs
```diff
@@ -261,11 +265,15 @@ canonicalRequest meth path digest query chs shs =
         toBS digest
       ]
 
-escapedPath :: Request a -> Path
-escapedPath r = Tag . toBS . escapePath $
-  case _serviceAbbrev (_requestService r) of
-    "S3" -> _requestPath r
-    _ -> collapsePath (_requestPath r)
+escapedPaths :: Request a -> (Path, CanonicalPath)
```
endgame, line 268: Can we return `CanonicalPath` only, and not a tuple?

lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs
mwu: Not really. This one's still used for the actual http request.

mwu: We could remove metaCanonicalPath instead since I only really added it for making testing easier. Tests would have to grab the canonical path from the canonical request after an unlines, which seemed a bit clunky.

lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs
mwu: Thinking about this more, yes it's possible. Where the actual http request is constructed, we'd need to do an unescape: https://github.com/brendanhay/amazonka/blob/9043cb4dc894a1a0d0f24375d70df5cb5e3b1ac8/lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs#L160  But it'd have to be typed and not ByteString since it has to know how to go back from twice escaped to only once-escaped as the http request is always made against the once-escaped URL.

endgame: Okay, I didn't realise that the left hand side of the tuple was still used. Would you mind splitting into two functions `escapePath` and `escapeCanonicalPath`?

mwu: Yep.

lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs
endgame: I feel a bit unsure about it: Is the request assembled in this function the correct type that the test harness could call `canonicalPath` directly?

lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs
endgame: Pretty sure the answer is "no".

lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs
mwu: How does this look? https://github.com/mwu/amazonka/pull/1

endgame: I think it looks good.

endgame (approved):
All right, so once you get your secondary PR merged I think this is good to go. Would you mind firing off your manual smoke-test requests one last time, too, just to make sure the tweaks are all good?

mwu:
> All right, so once you get your secondary PR merged I think this is good to go. Would you mind firing off your manual smoke-test requests one last time, too, just to make sure the tweaks are all good?

Ok. These all returned success:
* Lambda InvokeFunction with full arn, partial arn, full and partial arn with alias suffix, full and partial arn with alias in qualifier, function name, name with alias suffix, name with alias qualifier.
* S3 GetObject with keys `test/myron/test`, `test//myron//test', `abc:def$1.foo`
* Some other requests with non-empty paths: APIGateway GetStages, Route53 GetHostedZoneLimit
* A few of the examples in the examples project

lib/amazonka-core/test/Test/Amazonka/Sign/V4/Base.hs
```diff
@@ -0,0 +1,115 @@
+-- |
+-- Module      : Test.AWS.Sign.V$
```
endgame, line 2:
```suggestion
-- Module      : Test.AWS.Sign.V4
```

endgame (changes requested):
Excellent, thank you for being so thorough. One more nit, and one last question: I think I have misunderstood the purpose of the various `metaCanonicalFoo` in the metadata record - it seems like they don't get used anywhere because they get built and immediately passed to `signRequest`. I think you pointed this out earlier? When I asked you to remove it, I didn't realise there was an established practice of keeping this data around. My question is - do you think we should reinstate `metaCanonicalPath`? I think your tests are fine as they are.

lib/amazonka-core/test/Test/Amazonka/Sign/V4/Base.hs
mwu: Hah, thanks. I might have fat fingered that.

mwu:
> Excellent, thank you for being so thorough. One more nit, and one last question: I think I have misunderstood the purpose of the various `metaCanonicalFoo` in the metadata record - it seems like they don't get used anywhere because they get built and immediately passed to `signRequest`. I think you pointed this out earlier? When I asked you to remove it, I didn't realise there was an established practice of keeping this data around. My question is - do you think we should reinstate `metaCanonicalPath`? I think your tests are fine as they are.

I think it's fine as it is, too. The canonical path gets logged already as part of the canonical request and I can't see any other use for returning it in the metadata. Carrying it around is only useful for testing but as you said, we should really approach that by expanding the test suite to test the whole signature.

endgame (approved):
All right then, sounds good. Thanks very much for this.

PR #817, opened 2022-09-21
Fix haddock escaping for `fromFileEnv` function

cdepillabout:
This PR adds some necessary escaping in the haddocks of the `fromFileEnv` function.

Without this PR, the haddocks look like:

image

With the proper escaping from this PR, the haddocks now look like:

image

It is a little confusing that things in `@...@` in haddocks still require escaping.


endgame:
Good catch, thank you.

PR #821, opened 2022-10-07
amazonka-dynamodb: Fix WriteRequest encoding

kfigiela:
`DeleteRequests` in `BatchWriteItem` have been encoded incorrectly. They need to use `Key` instead of `Item` which is used for `PutRequest`. This MR fixes this.

See an example request in the AWS docs for reference https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchWriteItem.html


endgame:
Sorry for the mistake and thank you for the fix.

PR #828, opened 2022-10-18
amazonka-emr-serverless: Remove duplicate (case insensitive) modules

kfigiela:
Looks like there were some leftover modules generated with incorrect names in terms of letter casing that haven't been deleted. This causes issues on case-insensitive filesystems like on macOS. In my use case, nix was complaining about incorrect source checksum when extracted on macOS, since duplicate files were extracted to the same path.

In this PR, I remove the culprit modules. I did remove amazonka-emr-serverless and regenerated it from scratch, and only these files are in the resulting diff, so the package should be left in a clean state now.

image


endgame:
Sorry, and thank you.

Issue #638, opened 2021-09-27
`Seconds` type is too coarse


endgame:
Some services (example: DynamoDB) generally respond really quickly (<10ms is normal, IME), and occasionally take much longer. A DynamoDB specialist advised us to give up on slow requests and retry them instead of waiting, but the `Seconds` type can only hold an integral number of seconds.

Replace it with either more granular integral type, or a type that can handle decimals?

brendanhay:
Probably should switch to using [DiffTime](https://hackage.haskell.org/package/time-1.12/docs/Data-Time-Clock.html#t:DiffTime) (or [Pico](https://hackage.haskell.org/package/base-4.14.1.0/docs/Data-Fixed.html) directly.)

endgame:
`DiffTime` sounds good, as it's a type about time, but will need to see what `http-client` is prepared to accept. But will think about that later.

PR #829, opened 2022-10-21
Implement Seconds with DiffTime

pbrisbin:
There are known use cases for more precision than whole seconds.

Some choices were made:

1. In conversion to/from text values, we format and parse as `%Es`,
   which is whole seconds and up to 12 decimal points, if necessary

   This may be a compatibility concern, if clients are trying to use a
   _formatted_ `Seconds` and don't expect the decimal point.

   I can't see this being an issue for parsing AWS responses. This
   represents a relaxing of the parser (we now accept decimals), so
   nothing that worked before should fail now.

2. In conversion to (whole) microseconds, we `round`.

   We could do that differently (e.g. `floor`, or `ceil`), or we could
   retain as much precision as we can by not returning `Int` from this
   function either.

   I don't feel that precision beyond microseconds is useful. Evidence
   for this would be the Haskell `threadDelay` and `timeout` functions,
   and our current (only) use-case, which is setting retries in the
   `HTTP` module. None would gain any value, and in fact would only be
   made more complex, if we returned more precision.

Closes https://github.com/brendanhay/amazonka/issues/638.

### DIFF
--- a/lib/amazonka-core/src/Amazonka/Types.hs
+++ b/lib/amazonka-core/src/Amazonka/Types.hs
@@ -150,6 +150,7 @@ import qualified Data.Text as Text
 import qualified Network.HTTP.Client as Client
 import Network.HTTP.Types.Method (StdMethod)
 import Network.HTTP.Types.Status (Status)
+import Data.Time (defaultTimeLocale, parseTimeM, formatTime)
 
 -- | A convenience alias to avoid type ambiguity.
 type ClientRequest = Client.Request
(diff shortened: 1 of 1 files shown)


brendanhay:
Thanks for the PR! Please go ahead and remove the `stack.yaml` from the PR as that's a separate discussion, namely we only have limited bandwidth/time to test all build paths. Adding stack + `n>=1` resolvers further compounds the amount of testing we need to do.

pbrisbin:
Seems like `Hashable DiffTime` only exists in newer versions. Should I remove the instance, or implement it by hand?

brendanhay:
At which version of what dependency does that instance exist? We can bump bounds if it's something reasonable, otherwise I'd suggest implementing by hand (over using CPP) and preserving the instance for compatibility reasons.

endgame:
The `Hashable` instance for `DiffTime` is not present in either `time` or `hashable`, it's in [`time-compat`](https://hackage.haskell.org/package/time-compat), which seems reasonably well-maintained and with good bounds. But I don't know how everyone feels about packages which provide instances.

pbrisbin:
Well this is really strange. Everything compiles fine here for me stack and lts-19.28. But I looked at the `DiffTime` and `Hashable` documentation and the instance is not listed on either side.

pbrisbin:
Ah I found it, it seems `time-compat` is already here. I didn't expect that as @endgame made it sound like bringing it in was an option.

```console
% stack repl amazonka-core
Configuring GHCi with the following packages: amazonka-core
...
λ> :i DiffTime
...
instance Hashable DiffTime
  -- Defined in ‘time-compat-1.9.6.1:Data.Time.Orphans’
...
```

It seems 1.9.6 is the first version where it shows up.

brendanhay:
>  But I don't know how everyone feels about packages which provide instances.

I'm not huge on depending on packages just for their orphan instances when we have a newtype specifically to avoid orphans in the first place.

In this case I'd suggest writing the instance by hand, or, if that goes beyond a small hand-rolled function and into the realms of needing more dependencies or something increasingly bespoke, just drop the instance and bump the amazonka version appropriately.

brendanhay:
> stack repl amazonka-core

I'm not familiar how stack repl brings (transitive?) dependencies into scope, but I don't believe we have a direct dependency on it currently?

pbrisbin:
I could not find a direct dependency, sorry for being unclear.

The package is there transitively, hence the orphan being in scope. So I don't know if that changes the calculus for adding it as a direct dependency. But I will write the instance by hand, or remove it if that becomes cumbersome.

brendanhay (approved):

brendanhay:
> scooped up into the big bump that will be 2.0

Yeah - that's totally fair.

PR #831, opened 2022-10-24
Use two-step token request for Instance MetaData

### LINKED ISSUE #745: IMDSv2 Support
`amazonka` bundles an EC2 Instance Meta-Data Service (IMDS) client in `Amazonka.EC2.Metadata`. It should support IMDSv2, where clients first request a token using `PUT`, and then present that token on subsequent requests.

AWS documentation: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html

pbrisbin:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html

A few caveats:

- This requests a new token on every separate use a `get`. It would be
  better to support requesting a token first and re-using it across many
  calls, but the current interface doesn't support that and would have
  to be re-designed. I'm happy to do that, if given guidance on
  preferred ergonomics (`ReaderT`? Explicit argument passing? Set up the
  `Manager` with a `modifyRequest`?)

- The timeout of 21600 comes from the docs. We could probably go
  shorter given point one.

- I don't have any easy way to verify that this works. Does there happen
  to be integration test coverage?

Closes #745.

### DIFF
--- a/lib/amazonka/src/Amazonka/EC2/Metadata.hs
+++ b/lib/amazonka/src/Amazonka/EC2/Metadata.hs
@@ -60,6 +60,7 @@ import qualified Data.ByteString.Char8 as BS8
 import qualified Data.ByteString.Lazy as LBS
 import qualified Data.Text as Text
 import qualified Network.HTTP.Client as Client
+import Network.HTTP.Simple (setRequestHeader, setRequestMethod)
 
 data Dynamic
   = -- | Value showing whether the customer has enabled detailed one-minute
@@ -324,7 +325,7 @@ userdata m =
     Exception.try (get m (latest <> "user-data")) >>= \case
       Left (Client.HttpExceptionRequest _ (Client.StatusCodeException rs _))
         | fromEnum (Client.responseStatus rs) == 404 ->
-          return Nothing
+            return Nothing
       --
       Left e -> Exception.throwIO e
       --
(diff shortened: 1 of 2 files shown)


lib/amazonka/src/Amazonka/EC2/Metadata.hs
```diff
@@ -324,7 +325,7 @@ userdata m =
     Exception.try (get m (latest <> "user-data")) >>= \case
       Left (Client.HttpExceptionRequest _ (Client.StatusCodeException rs _))
         | fromEnum (Client.responseStatus rs) == 404 ->
-          return Nothing
+            return Nothing
```
pbrisbin, line 328:
```suggestion
          return Nothing
```

I think ormolu did this. But I'll back it out to keep the PR more isolated.

endgame (changes requested):
Thank you for this. I've had a think about the points you raised:

* Tokens are really fast to generate, and I think most applications do a tiny amount of IMDS access to get their instance profile and don't really look it up otherwise. Rewriting everything to keep an `IORef Token` in `Env'` seems like a lot of work for minimal benefit. I'll make a follow-up issue and close it WONTFIX unless someone who cares a lot agitates for it.
* Agreed that the timeout should be shorter. I've never seen these responses not be instant, so I'm happy with tokens that have 60s timeout or so.
* I've launched an EC2 instance with IMDSv1 support disabled and played around with this branch. There are two main issues:
  1. It constructs malformed URLs, meaning that requests fail, and
  2. The default credential chain fails to detect that it's running on EC2, because the `isEC2` function isn't using tokens in its request.

Please also add an entry to `lib/amazonka/CHANGELOG.md`.

pbrisbin:
Thanks for the review, addressing these things now. Is there a reason `isEC2` uses `"http://instance-data/latest"`, while the rest of the calls use `"http://169.254.169.254/latest/"`? Does that need to be preserved? It makes it annoying to DRY the token-handling and is the reason I missed it on first coding.

endgame:
> Is there a reason `isEC2` uses `"http://instance-data/latest"`, while the rest of the calls use `"http://169.254.169.254/latest/"`?

I believe that in most cases, it's faster to attempt name resolution and have a local nameserver reply "nope", than it is to wait for an addressed TCP connection to the IMDS address to fail to open. Otherwise, you get a stall in any amazonka program not running on EC2 if the earlier auth methods fail to find keys.

lib/amazonka/src/Amazonka/EC2/Metadata.hs
```diff
@@ -295,7 +296,7 @@ isEC2 :: MonadIO m => Client.Manager -> m Bool
 isEC2 m = liftIO (Exception.catch req err)
   where
     req = do
-      !_ <- request m "http://instance-data/latest"
+      !_ <- get m "http://instance-data/latest"
```
endgame, line 299: This will actually do an unnecessary request: as soon as we get a token back we are able to say "yes, we're on EC2". But I don't think it's worth fixing; raise another PR if you disagree?

endgame (approved):
This is looking much better, and I've tested it on AWS EC2. I note that OpenStack claims to provide an [EC2-compatible IMDS](https://docs.openstack.org/nova/latest/user/metadata.html#metadata-ec2-format), but doesn't seem to support tokens as it's on a 2009 version of the API.

Last time I did IMDS work (trying to add V2 support to NixOS AMIs), I couldn't find anyone who cared about openstack, and couldn't get their test stack stuff to successfully launch on my local machine. I'm less concerned about it here as we're specifically an AWS SDK.

pbrisbin:
Thanks for the quick merge!

PR #797, opened 2022-07-09
A repro for issue #796

basvandijk:
This adds a reproducible test case for issue #796 "PutObject request failing on MinIO because of malformed XML" by adding an integration test checking `amazonka-s3` against MinIO. To execute the test just run the following:

```
$ nix-build tests/s3.nix
...
s3: must succeed:
    echo 'Hello World!' > ./some-file
    export AWS_ACCESS_KEY_ID="BKIKJAA5BMMU2RHO6IBB"
    export AWS_SECRET_ACCESS_KEY="V7f1CwQqAcwo80UEIJEjc5gVQUSSx5ohQ9GSrr12"
    amazonka-s3-test-app ./some-file http://s3:9000 my-bucket some-file

s3 # > > > > > [   11.524753] minio[818]: {
  "deploymentid": "0de4a2b3-eaf1-46a0-8ec7-2e706e68d3b4",
  "level": "ERROR",
  "errKind": "MINIO",
  "time": "2022-07-09T18:07:28.335404358Z",
  "api": {
    "name": "SYSTEM",
    "args": {}
  },
  "error": {
    "message": "EOF (*errors.errorString)",
    "source": [
      "cmd/handler-utils.go:55:cmd.parseLocationConstraint()",
      "cmd/auth-handler.go:344:cmd.checkRequestAuthTypeCredential()",
      "cmd/auth-handler.go:296:cmd.checkRequestAuthType()",
      "cmd/bucket-handlers.go:719:cmd.objectAPIHandlers.PutBucketHandler()",
      "net/http/server.go:2047:http.HandlerFunc.ServeHTTP()"
    ]
  }
}
s3 # amazonka-s3-test-app:   ServiceError
    ( ServiceError'
        { _serviceErrorAbbrev = Abbrev {fromAbbrev = "S3"},
          _serviceErrorStatus = Status {statusCode = 400, statusMessage = "Bad Request"},
          _serviceErrorHeaders =
            [ ("Accept-Ranges", "bytes"),
              ("Content-Length", "371"),
              ("Content-Security-Policy", "block-all-mixed-content"),
              ("Content-Type", "application/xml"),
              ("Server", "MinIO"),
              ("Strict-Transport-Security", "max-age=31536000; includeSubDomains"),
              ("Vary", "Origin"),
              ("Vary", "Accept-Encoding"),
              ("X-Amz-Bucket-Region", "us-east-1"),
              ("X-Amz-Request-Id", "17003B77BD220CF2"),
[...]

PR #832, opened 2022-10-27
S3 vhost customisation

### LINKED ISSUE #760: S3: Setting endpoint interferes with vhost style URLs
All s3 requests call [`s3vhost`](https://github.com/brendanhay/amazonka/blob/f73a957d05f64863e867cf39d0db260718f0fadd/lib/amazonka-core/src/Amazonka/Request.hs#L169-L194) after building the request. This moves the bucket name from the first segment in the path to the host name of the request URL. This only works as long as the S3 Service doesn't have the endpoint overridden. I tested this by using [`setEndpoint`](https://github.com/brendanhay/amazonka/blob/f73a957d05f64863e867cf39d0db260718f0fadd/lib/amazonka-core/src/Amazonka/Endpoint.hs#L21-L39), I couldn't find any other ways of overriding this endpoint. When the endpoint is overridden, neither the path nor the hostname have the bucket name.

My usecase is to support S3-compatible storage services like [minio](https://github.com/minio/minio).

Additionally, it would be great if there was a way to specify whether I want to use path style URLs or vhost style URLs.

endgame:
Implement S3 vhost rewriting customisation using the [design proposed by @basvandijk](https://github.com/brendanhay/amazonka/pull/797#issuecomment-1181531912). The basic design is the same:

* `request :: AWSRequest a => Service -> a -> Request a` in `class AWSRequest` is now accepts the service to send the request to, instead of generators hardcoding `defaultService`. This allows the `Request` to be computed from the overriden `Service` instead of `defaultService`.
* We need to be able to determine the default service for a request type when doing initial signing, so we need a typeclass method `service :: AWSRequest a => Proxy a -> Service` _somewhere_. We apply overrides to it before passing it to `request`.
* It might be possible to avoid the additional typeclass method by making `request :: AWSRequest a => (Service -> Service) -> a -> Request a`, but every option I could think of seemed worse than a new typeclass method.

There are some differences to the original design (which I cherry-picked from @ivb-supercede's [branch with generator fixes](https://github.com/SupercedeTech/amazonka/commits/temp-vhost-fix)):

* The `service` function is a method of `class AWSRequest a` instead of a separate `class AWSService a` and making `class AWSService a => AWSRequest a`; I couldn't see much benefit in a separate typeclass. Am I wrong?
* Instead of a `Bool`, I added a sum type to match the [S3 addressing style options](https://boto3.amazonaws.com/v1/documentation/api/1.9.42/guide/s3.html#changing-the-addressing-style)

Because we're going to have to regenerate everything, I updated botocore as well. A full regeneration is available at [`endgame:s3-vhost-customisation-generation`](https://github.com/endgame/amazonka/tree/s3-vhost-customisation-generation) and will be merged once this branch lands. I have rebuilt the entire repo locally and it at least compiles.

Ping @basvandijk @ivb-supercede @kfigiela @akshaymankar @periodic who might be interested — please test the branch above.

@basvandijk / @ivb-supercede: Is it OK that I've cherry-picked your commits here?

Fixes #760 
Closes #797

### DIFF
--- a/gen/src/Gen/AST/Data/Syntax.hs
+++ b/gen/src/Gen/AST/Data/Syntax.hs
@@ -216,6 +216,7 @@ serviceD m r = Exts.patBindWhere (pvar n) rhs bs
           field (unqual "Core._serviceEndpointPrefix") (m ^. endpointPrefix . Lens.to str),
           field (unqual "Core._serviceSigningName") (m ^. signingName . Lens.to str),
           field (unqual "Core._serviceVersion") (m ^. apiVersion . Lens.to str),
+          field (unqual "Core._serviceRewriteS3VHost") (var "Prelude.True"),
           field (unqual "Core._serviceEndpoint") (Exts.app (var "Core.defaultEndpoint") (var n)),
           field (unqual "Core._serviceTimeout") (Exts.app justE (Exts.intE 70)),
           field (unqual "Core._serviceCheck") (var "Core.statusSuccess"),
--- a/gen/src/Gen/AST/Data/Syntax.hs
+++ b/gen/src/Gen/AST/Data/Syntax.hs
@@ -360,6 +360,18 @@ notationE' withLensIso = \case
       TMaybe x -> var "Lens._Just" : lensIso x
       _other -> []
 
+serviceInstD ::
+  HasMetadata a Identity =>
+  a ->
+  Ref ->
+  Decl
+serviceInstD m a =
+  instD
+    "Core.AWSService"
+    (identifier a)
+    [ funArgsD "service" ["_proxy"] (var (m ^. serviceConfig))
+    ]
+
 requestD ::
   HasMetadata a Identity =>
   Config ->
@@ -374,7 +386,7 @@ requestD c m h (a, as) (b, bs) =
     (identifier a)
     $ Just
       [ assocD (identifier a) "AWSResponse" (typeId (identifier b)),
-        funD "request" (requestF c m h a as),
+        funArgsD "request" ["srv"] (requestF c m h a as),
         funD "response" (responseE (m ^. protocol) b bs)
       ]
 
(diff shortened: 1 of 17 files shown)


gen/src/Gen/AST/Data/Syntax.hs
```diff
@@ -374,7 +375,8 @@ requestD c m h (a, as) (b, bs) =
     (identifier a)
     $ Just
       [ assocD (identifier a) "AWSResponse" (typeId (identifier b)),
-        funD "request" (requestF c m h a as),
+        funArgsD "service" ["_"] (var (m ^. serviceConfig)),
+        funArgsD "request" ["srv"] (requestF c m h a as),
```
endgame, line 379: We could try to point-free this in the common case, but then we need to think harder when plugins are involved. Worth it?

ivb-supercede:
>The service function is a method of class AWSRequest a instead of a separate class AWSService a and making class AWSService a => AWSRequest a; I couldn't see much benefit in a separate typeclass. Am I wrong?

I don't know enough to comment on this in general, since I just cribbed this from Bas' code. But in the case where `AWSService` is a superclass of `AWSRequest`, then this is design doesn't really make a difference - without any typeclass laws, what meaningful code can be written with only a `AWSService` instance?

>Ping @basvandijk @ivb-supercede @kfigiela @akshaymankar @periodic who might be interested — please test the branch above.

Will do.

>@basvandijk / @ivb-supercede: Is it OK that I've cherry-picked your commits here?

All good with me :+1:.

kfigiela:
I can confirm that our unit tests work just fine with aws mocked with localstack and endpoint override configured not to use vhosts.

ivb-supercede:
Looks like these changes work fine for us, for both real AWS usage and for local mocking with Moto.

gen/src/Gen/AST/Data/Syntax.hs
endgame: Nobody seems to think it's worth it.

PR #856, opened 2022-12-08
Ghc updates

endgame:
Attempt to bump the nixpkgs pin so we can follow modern GHC. The `stackage_snapshot` in `WORKSPACE` only works with `ghc902` and I'm not sure how to fix it up so that we select a working snapshot for each GHC version.

Also, IOHK moved their cache's DNS name, so fix that.

We can't _use_ the GHC 9.2.5 available in nixpkgs because of a regression in cabal; once everything else is working we can fix that: https://github.com/tweag/gazelle_haskell_modules#using-ghc-version-92


endgame:
@brendanhay I'm really going to need your help on this one - I only have a very basic understanding of bazel.

joshburgess:
Any updates on this? It would really be nice to be able to get onto GHC 9.2.5 now that the rest of the ecosystem (HLS, etc.) is now reliably using it.

endgame:
We've been using GHC9.2.x and Amazonka `main` at work for most of the year. This PR is more about getting as close to 9.4 or 9.6 support as possible.

If GHC 9.2.x doesn't work for you on `main`, please file an issue and I'll take a look.

brendanhay:
Seems this is no longer necessary, closing!

endgame:
Yep, not required.

PR #859, opened 2022-12-08
Handwitten lenses in service bindings

### LINKED ISSUE #846: Rename handwritten optics to match new convention
e.g., in `Amazonka.S3.Internal`, there are a few handwritten lenses, isos and traversals.

endgame:
Fixes #846

### DIFF
--- a/lib/services/amazonka-s3/src/Amazonka/S3/Internal.hs
+++ b/lib/services/amazonka-s3/src/Amazonka/S3/Internal.hs
@@ -9,15 +9,13 @@
 
 -- |
 -- Module      : Amazonka.S3.Internal
--- Copyright   : (c) 2013-2021 Brendan Hay
+-- Copyright   : (c) 2013-2022 Brendan Hay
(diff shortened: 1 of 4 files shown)


lib/services/amazonka-s3/src/Amazonka/S3/Internal.hs
```diff
@@ -211,9 +220,9 @@ _ObjectKey = iso (\(ObjectKey k) -> k) ObjectKey
 --
 -- >>> "/" ^? keyPrefix '/'
```
Tristano8, line 221: Do these haddocks need to be updated to match the name changes?

Tristano8 (approved):
Just the one question

lib/services/amazonka-s3/src/Amazonka/S3/Internal.hs
endgame: Yes, they do.

PR #862, opened 2022-12-09
Gen sorted interfaces

### LINKED ISSUE #825: gen: Stop shape field names from jumping about
Sort field names or something to stop them jumping around unpredictably when we regenerate.

Sorting would move `createdAt` away from `updatedAt` - maybe the thing to do is make sure we strictly follow botocore's ordering?

endgame:
Switch the generator from `HashMap` to `Map`, which means that all the `toList` calls come out in sorted order. This will keep the ordering of fields predictable and should give us better diffs in future generator runs.

While it might be nice to follow botocore's ordering for struct fields, I can't see a good way to do that: https://github.com/haskell/aeson/issues/984

A full regeneration is on the `endgame:gen-sorted-interfaces-gen` branch (#863). Please try it and let me know if you think this is better than the arbitrary "however `HashMap.toList` sorted them" that we used to have.

Closes #825 .

### DIFF
--- a/gen/src/Gen/AST.hs
+++ b/gen/src/Gen/AST.hs
@@ -5,7 +5,7 @@ import qualified Control.Lens as Lens
 import qualified Control.Monad.Except as Except
 import Control.Monad.State.Strict (execState, modify)
 import qualified Control.Monad.State.Strict as State
-import qualified Data.HashMap.Strict as HashMap
+import qualified Data.Map.Strict as Map
 import qualified Data.HashSet as HashSet
 import qualified Data.Set as Set
 import qualified Data.Text as Text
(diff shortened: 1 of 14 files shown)


endgame:
It seems unlikely that aeson will gain the ability to do sorted `Object`s any time soon, so I think that makes sorted fields a better way to go.

mbj:
> Please try it and let me know if you think this is better than the arbitrary "however HashMap.toList sorted them" that we used to have.

I did not try the branch. But determinism that does not depend on an implementation artifact (that in case of `HashMap` should actually not be that stable, to avoid collision / DOS attacs!)  is better.

Given we cannot (easily) retain the source field order ATM, I think sorting by field name is the best compromise.

michaelwebb76 (approved):
Seems sensible enough to me, as a relative Haskell n00b

PR #868, opened 2023-01-03
Parse more timezones

### LINKED ISSUE #866: Serialize error: Failure parsing Time from value
```
cabal-cache: SerializeError (SerializeError' {_serializeAbbrev = Abbrev "S3", _serializeStatus = Status {statusCode = 200, statusMessage = ""}, _serializeBody = Nothing, _serializeMessage = "Failed reading: Failure parsing Time from value: \"Thu, 15 Dec 2022 08:43:15 PST\""})
```

https://cirrus-ci.com/task/6316388051058688?logs=main#L1389

It seems it can't parse `Thu, 15 Dec 2022 08:43:15 PST`.

Code seems to be here:

https://github.com/brendanhay/amazonka/blob/cfe2584aef0b03c86650372d362c74f237925d8c/lib/amazonka-core/src/Amazonka/Data/Time.hs#L105-L128

First discovered in cabal-cache: https://github.com/haskell-works/cabal-cache/issues/207

endgame:
Change the format string for RFC822 to use `%Z` for any timezone instead of the literal `GMT`. The use of `GMT` in that format string goes as far back as I can trace (at least 8 years), so I'm surprised to see errors beginning now. Regardless, it's valid RFC 822 so let's support it.

Fixes #866

### DIFF
--- a/lib/amazonka-core/src/Amazonka/Data/Time.hs
+++ b/lib/amazonka-core/src/Amazonka/Data/Time.hs
@@ -88,19 +88,7 @@ instance TimeFormat BasicTime where
 instance TimeFormat AWSTime where
   format _ = "%Y%m%dT%H%M%SZ"
 
-instance FromText BasicTime where
-  fromText = A.parseOnly ((parseUnixTimestamp <|> parseFormattedTime) <* A.endOfInput)
-
-instance FromText AWSTime where
-  fromText = A.parseOnly ((parseUnixTimestamp <|> parseFormattedTime) <* A.endOfInput)
-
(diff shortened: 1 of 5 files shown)


endgame:
@hasufell please test this PR and let me know if it fixes your issue.

hasufell:
> @hasufell please test this PR and let me know if it fixes your issue.

I'm afraid I can't test this easily, because the libraries that I use don't build with amazonka HEAD and I don't have the time to invest in updating all that.

Given that you have a test case using that exact date string, I guess it's probably fine.

PR #869, opened 2023-01-04
amazonka-core: Support custom base paths in endpoints

### LINKED ISSUE #786: How to change endpoint of a service? Especially the APIGatewayManagementAPI service
I'm reading the document:

> The Amazon API Gateway Management API allows you to directly manage runtime aspects of your deployed APIs. To use it, you must explicitly set the SDK's endpoint to point to the endpoint of your deployed API. The endpoint will be of the form https://{api-id}.execute-api.{region}.amazonaws.com/{stage}, or will be the endpoint corresponding to your API's custom domain and base path, if applicable.

But it doesn't say how we can change the endpoint.

The closet I can find is `serviceEndpoint`

```
let url = "https://{api-id}.execute-api.{region}.amazonaws.com/{stage}"
let env' = override (over serviceEndpoint (endpointHost .~ encodeUtf8 url)) env
```

But this only changes the domain. The `stage` part needs to be a part of the path. The `connectionId` is also on the path so stage needs to be prefixed. Is there a way to achieve this?

endgame:
@stevemao please test this and let me know whether it works for you.

Closes #865 
Fixes #786

### DIFF
--- a/lib/amazonka-core/src/Amazonka/Types.hs
+++ b/lib/amazonka-core/src/Amazonka/Types.hs
@@ -164,6 +164,7 @@ module Amazonka.Types
 
     -- ** Lenses
     endpoint_host,
+    endpoint_basePath,
     endpoint_secure,
     endpoint_port,
     endpoint_scope,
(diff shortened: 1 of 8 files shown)


lib/amazonka-core/src/Amazonka/Types.hs
```diff
@@ -417,8 +418,15 @@ instance AsError Error where
 
 data Endpoint = Endpoint
   { host :: ByteString,
```
endgame, line 420: Should say something about domain name here.

stevemao (approved):
I confirm this works

endgame:
Thanks for testing.

PR #873, opened 2023-01-05
amazonka: Stop using `Dual (Endo Service)` for overrides

endgame:
Instead unwrap the newtypes so that this is just a function.

This seems significantly simpler.

### DIFF
--- a/lib/amazonka/src/Amazonka/Env.hs
+++ b/lib/amazonka/src/Amazonka/Env.hs
@@ -24,7 +24,7 @@ module Amazonka.Env
     env_region,
     env_logger,
     env_retryCheck,
-    env_override,
+    env_overrides,
     env_manager,
     env_auth,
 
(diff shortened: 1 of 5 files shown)


lib/amazonka/src/Amazonka/Env.hs
```diff
@@ -189,7 +188,7 @@ retryConnectionFailure limit n = \case
 -- | Provide a function which will be added to the existing stack
 -- of overrides applied to all service configurations.
 overrideService :: (Service -> Service) -> Env' withAuth -> Env' withAuth
-overrideService f env = env {override = override env <> Dual (Endo f)}
+overrideService f env = env {overrides = f . overrides env}
```
endgame, line 191: This appears to be the only place where we used the monoid at all.

axman6 (approved):
Looks like a reasonable change to me, if the rest of the Env were also monoidal things it'd make sense but all this does at the moment is add confusing for newer developers.

PR #875, opened 2023-01-06
Hooks

### LINKED ISSUE #501: Should `Network.AWS.Internal.HTTP.retrier` log an error each time a request fails?
This is the current behavior, but it seems a bit severe. I usually try to reserve error-level logging for issues that someone needs to take a look at (generally with automated alerts). A request failing initially but succeeding later doesn't seem to rise to that level.

Perhaps it would make more sense to log only the final failure as an error, and the intermediate ones as info or warnings?

### LINKED ISSUE #584: Can't quietly handle exceptions
I have some simple code that checks if an image with a tag already exists in some ECR repository:

```haskell
        AWS.trying ECR._ImageNotFoundException (AWS.send describeImages) >>= \case
            -- We failed with ImageNotFound, clearly the tag is not present.
            Left{} -> pure False
            -- We got some result, must be for image ID we passed.
            Right{} -> pure True
```

I'm using `newLogger` with `Error :: LogLevel`. Running this code, in the console I see:

```
[ServiceError] {
  service    = ECR
  status     = 400 Bad Request
  code       = ImageNotFound
  message    = Just The image with imageId {imageDigest:'null', imageTag:'SOMETAG'} does not exist within the repository with name 'some_repo' in the registry with id 'xxxxx'
  request-id = Just xxxxx
}
# output from rest of program
```

I don't see how I can handle `ImageNotFoundException` such that the warning is not printed: it's expected here. Creating new env with silent logger isn't great either: I still want to see any other exceptions that may occur, like connection errors and such.

Am I missing something obvious?

### LINKED ISSUE #737: Discussion: Silent logging or better request/response lifecycle hooks?
I was writing some code the other day that talks to DynamoDB, where it's sometimes necessary to make conditional requests that are expected to fail. At higher logging levels, these generate noisy and irrelevant log messages (see also: #584), but we can't turn off logging for that one request without potentially missing failures for other reasons.

I went to add some more hooks to `Env` to selectively silence certain log messages, but then I started to wonder: what if we replaced the logging stuff with a bunch of hooks that are called before/after different points in the request/response cycle? Then logging becomes a single special case, albeit one where we can provide sensible defaults.

Thoughts @brendanhay ?

CC: @fuuzetsu, @lrworth @unisay who may also be affected.

endgame:
Add a system of request/response hooks, to allow fine-grained customisation of the request lifecycle. The main module that you should read is `Amazonka.Env.Hooks` (which is so big that GitHub "helpfully" collapses it), and I have reproduced its synopsis below:

```haskell
-- Hooks carried within an 'Env', allowing ad-hoc injection of
-- different behaviour during Amazonka's request/response cycle.
-- Hooks are currently experimental, but Amazonka uses the 'Hooks' API
-- to implement its default logging, and you can add your own
-- behaviour here as well. Some examples of things hooks can do:
--
-- * Log all requests Amazonka makes to a separate log, in order to
--   audit which IAM permissions your program actually needs
--   (see 'requestHook');
--
-- * Inject a [Trace ID for AWS X-Ray](https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-tracingheader),
--   into each request before it is signed and sent
--   (see 'configuredRequestHook'); and
--
-- * Selectively silence certain expected errors, such as DynamoDB's
--   @ConditionalCheckFailedException@ ('errorHook' and 'silenceError')
--
-- Most functions with names ending in @Hook@ ('requestHook', etc.)
-- are intended for use with lenses: partially apply them to get a
-- function @'Hook' a -> 'Hook' a@ that can go on the RHS of @(%~)@
-- (the lens modify function). You then use functions like
-- 'hookMatchingType' to selectively extend the hooks used at any
-- particular time.
--
-- Names ending in @_@ ('Hook_', 'hookMatchingType_', etc.) do not
-- feed an updated value back into Amazonka because it is either
-- unsafe to do so, or difficult to do anything useful with the
-- updated value.
```

Pretty much everything else is plumbing. Shout out to @axman6 for design help with this.
Ping @Fuuzetsu who raised the original issue about silencing exceptions.

Fixes #584 
Fixes #501
Closes #737


endgame:
`Logger` is still passed all the way down into `amazonka-core`, so that response deserialisation can log the raw response body (it cannot happen sooner for laziness reasons). If people like the hooks interface, I will replace the `Logger` parameter with a `ByteString -> IO ()` parameter so that we can have a hook for the raw response body too.

endgame:
#501 notes that retries are noisy. Should we have a `finalError` hook so that we spam the logs a little less?

lib/amazonka/src/Amazonka/Env/Hooks.hs
```diff
@@ -0,0 +1,420 @@
...
+    Nothing -> pure ()
+
+-- | When @a@ and @b@ are the same type, do not call any more hooks.
+--
+-- -- Example: Prevent any request hooks from running against a @PutObjectRequest@:
+-- @
+-- requestHook (blockMatchingType @PutObjectRequest) :: Hooks -> Hooks
+-- @
+blockMatchingType :: forall a b. (Typeable a, Typeable b) => Hook b -> Hook b
+blockMatchingType oldHook = \env -> case eqT @a @b of
```
axman6, line 324:
```suggestion
-- requestHook (removeHooksFor @PutObjectRequest) :: Hooks -> Hooks
-- @
removeHooksFor :: forall a b. (Typeable a, Typeable b) => Hook b -> Hook b
removeHooksFor oldHook = \env -> case eqT @a @b of
```

lib/amazonka/src/Amazonka/Env/Hooks.hs
```diff
@@ -0,0 +1,420 @@
...
+-- @
+-- errorHook (blockMatchingType @(Request PutObjectRequest, Error)) :: Hooks -> Hooks
+-- @
+blockMatchingType_ :: forall a b. (Typeable a, Typeable b) => Hook_ b -> Hook_ b
+blockMatchingType_ oldHook = \env a -> case eqT @a @b of
+  Just Refl -> pure ()
+  Nothing -> oldHook env a
+
+-- | Run the wrapped hook unless the given 'Fold' or 'Traversal'
+-- matches the error. With the error matches in each service binding,
```
axman6, line 340: Something wrong with this sentence, not sure what it was supposed to say.

lib/amazonka/src/Amazonka/Env/Hooks.hs
```diff
@@ -0,0 +1,420 @@
...
+          rq' <- clientRequest env rq
+          rq' <$ logDebug logger rq',
+        clientResponse = \env@Env {logger} t@(_, rs) -> do
+          clientResponse env t
+          logDebug logger rs,
+        requestRetry = \env@Env {logger} t@(_, name, retryStatus) -> do
+          requestRetry env t
+          logDebug logger
+            . mconcat
+            . intersperse " "
```
axman6, line 384: Since it's used twice here, is it worth defining a `unwords :: [Builder] -> Builder`?

lib/amazonka/src/Amazonka/Env/Hooks.hs
```diff
@@ -0,0 +1,420 @@
...
+--
+-- -- Example: Remove all response hooks:
+-- @
+-- responseHook noHook_ :: Hooks -> Hooks
+-- @
+noHook_ :: Hook_ a -> Hook_ a
+noHook_ _ _ _ = pure ()
+
+-- | When @a@ and @b@ are the same type, run the given 'Hook a' after
+-- all others, otherwise do nothing.
```
divarvel, line 289: i had a bit of trouble understanding what "do nothing" meant (since it's just after "run the given Hook a", at first i understood the "do nothing" to be when the hooks are run, not when this function is called).

Maybe "Append the provided 'Hook a' to the existing hooks when 'a' and 'b' match. Otherwise, leave existing hooks unaffected"?

divarvel:
It's a clever way to handle the different types of requests with a single function. It took me a bit of time to put all pieces together, so maybe an explanation or a complete example in the module header might be nice.

Also, as usual for hooks, a diagram might be a good addition for documenting stages in a visual way.

lib/amazonka/src/Amazonka/Env.hs
```diff
@@ -51,8 +53,14 @@ import qualified Network.HTTP.Client as Client
...
 import System.Environment as Environment
 
+-- | An environment with auth credentials. Most AWS requests need one
+-- of these, and you can create one with 'Amazonka.Env.newEnv'
 type Env = Env' Identity
 
+-- | An environment with no auth credentials. Used for certain
+-- requests which need to be unsigned, like
+-- @sts:AssumeRoleWithWebIdentity@, and you can create one with
+-- 'Amazonka.Env.newEnvNoAuth' if you need it.
```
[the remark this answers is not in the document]
endgame: I did it anyway.

lib/amazonka/src/Amazonka/Env.hs
```diff
@@ -84,6 +87,10 @@ env_region f e@Env {region} = f region <&> \region' -> e {region = region'}
 env_logger :: Lens' (Env' withAuth) Logger
```
brendanhay, line 87: The `env_` naming convention seems a little personal / out of place?

lib/amazonka/src/Amazonka/Env.hs-boot
```diff
@@ -0,0 +1,16 @@
+module Amazonka.Env where
+
+import Amazonka.Types (Region, Logger, Service, Auth)
+import Amazonka.Prelude
+import qualified Network.HTTP.Client as Client
+import {-# SOURCE #-} Amazonka.Env.Hooks (Hooks)
+
+data Env' (withAuth :: Type -> Type) = Env
```
brendanhay, line 8: Seems somewhat arbitrary to the naive reader (ie. me) parameterising over some `f :: Type -> Type` when you could just define:

```haskell
data EnvF auth = Env {
    ...,
    auth :: auth,
} deriving stock (Functor, ...)

type Env = EnvF (Identity Auth)
```

And pass parameters as: 

```haskell
retryRequest ::
  forall m a f.
  ( Foldable f
  ) =>
  EnvF (f Auth) ->
  a ->
  m (Either Error (ClientResponse (AWSResponse a)))
```

lib/amazonka/src/Amazonka/Env.hs
endgame: This is for consistency with the lenses generated for service bindings cf. https://github.com/brendanhay/amazonka/blob/main/lib/services/amazonka-s3/gen/Amazonka/S3/Lens.hs

If you think we should use a different convention, I'm happy to do that treewide in another PR?

lib/amazonka/src/Amazonka/Env.hs-boot
endgame: Let's discuss in a separate issue, since the `Type -> Type` version has been in since the [env rework](https://github.com/brendanhay/amazonka/pull/746/files#diff-af09bf5533244a54c9ff45b3d0a799424cf4e39c48af0726e0cf5c5b1161ca96R54-R61) in #746 . It shows up in this diff because this PR added a `.hs-boot` file.

endgame: Opened #877 - I dunno if/when github pings people.

lib/amazonka/src/Amazonka/Env/Hooks.hs
endgame: Good idea. I will also rename `hookMatchingType` -> `addHookFor`.

lib/amazonka/src/Amazonka/Env/Hooks.hs
endgame: It's a pointer to look up each service's error _matchers_. Will reword.

lib/amazonka/src/Amazonka/Env/Hooks.hs
endgame: Yeah I'll do it, since it's likely to pay off if we add any more logging hooks (e.g., for `finalError`.)

lib/amazonka/src/Amazonka/Env/Hooks.hs
endgame: Your suggestion sounds to me like it decides whether to append a hook before the time hooks are run. I'll try a different wording, let me know if you like it.

endgame: Updated and pushed. How do you like the docs now?

endgame:
I intend to merge this in the next couple of days - last call for comments. I'm interested to hear:

* @brendanhay are you happy with the lensy stuff?
* Do we want another logging hook for final errors, to fix #501?
  - Move the `error` logging hook to `debug` or `warning` level
  - Add a `finalError` logging hook which logs at `error` level

lib/amazonka/src/Amazonka/Env/Hooks.hs
divarvel: neat, thanks!

kokobd:
> * Do we want another logging hook for final errors

@endgame For `error` and `finalError`, we may merge them into one and provide some context information to the hook, including:

- how many times have we retried
- is this the final retry?
- how long time has passed since the first request?
- what's the current retry policy

Then the user may define some fine-grained logic to control errors in retry, such as increasing the alarm level when retry times increase or timing out at some point

endgame:
Yeah I agree that the `error`/`finalError` distinction is a bit rough. Probably better to merge the two and add more information to the `Hook_` itself. At the moment, the final error gets logged twice and it might be possible to avoid that with the right additional information.

lib/amazonka/src/Amazonka/Env.hs
brendanhay: The naming for the generated lenses exists because there aren't many obvious conventions that are guaranteed to be unambiguous, not because they're particuarly on brand with Haskell naming conventions.

Consider naming lenses as `camelCase {type-name} <>  camelCase {field-name}`:

```
MyDataType.foo -> myDataTypeFoo
MyData.typeFoo -> myDataTypeFoo
```

Versus `camelCase {type-name} <> "_" <> {verbatim-field-name}`:

```
MyDataType.foo -> myDataType_foo
MyData.typeFoo -> myData_typeFoo
```

I'm fine either way, doesn't need to be changed.

lib/amazonka/src/Amazonka/Env/Hooks.hs
```diff
@@ -0,0 +1,600 @@
...
+-- matches the error. You will probably want to use this with the
+-- error matchers defined by each service binding, allowing you to
+-- selectively silence specific errors:
+--
+-- @
+-- -- Assuming `env :: Amazonka.Env` and `putRequest :: DynamoDB.PutRequest`,
+-- -- this silences a single type of error for a single call:
+-- send (env & #hooks %~ errorHook (silenceError DynamoDB._ConditionalCheckFailedException))
+-- @
+silenceError ::
```
endgame, line 519: Need to think more about this — in its current form, will have to put it in twice: once for non-final errors and once for final errors. We should be able to do better than this.

lib/amazonka/src/Amazonka/Env.hs
endgame: I thought this over last weekend, and I think it's better for amazonka to be consistent within itself, even if that means the hand-written stuff is a bit different to the "standard". As with the service bindings, these lenses are for people who don't use `generic-lens` or `generic-optics`.

endgame:
> @endgame For `error` and `finalError`, we may merge them into one and provide some context information to the hook, including: -snip-

I agree that we should consolidate error hooks, and have done so. (This is also necessary to make a single use of `silenceError` actually silence an error.)

I think we avoid pushing too many additional parameters into the error/retry hooks at this stage. Instead, let's wait and see what people actually want, and work out the best way to give that to them. Otherwise, I fear we will add things for use-cases that nobody wants, and bind ourselves to maintaining them.

mbj:
BTW, once this one is merged I'll setup our XRay tracing to cover AWS service calls via that hooks. Quite happy about the new feature. Before I was doing a wierd "custom send" function based tracing of these that never was scalable on the developer overhead axis.

endgame:
@mbj Is there anything here that's missing for your use case? While I'm planning to ship the hooks as "officially experimental", it would be good to at least be able to support known use cases.

mbj:
@endgame on early inspection its what I need, but I cannot know for sure till I ported my ugly wrapper around `send`.

endgame:
@mbj Normally I'd ask people to test a PR branch so I have some confidence that merging it is going to be fine, but in this case I think merging now and testing its features is probably the way to go. Can you please let me know (via issue) whether or not the hooks are letting you do the things you want to do?

PR #878, opened 2023-01-09
Generalise Service Error Matcher Optics

endgame:
The service error matchers have a type annotation which restricts them to `AsError a => Getting (First ServiceError) a ServiceError`. Generalise the type signature to the `AsError a => Fold a ServiceError` that they are, and update the generator to annotate generated matchers correctly also.

The optics actually type-check as `AsError a => Traversal' a ServiceError` or even `AsError a => Prism' a ServiceError`, but because they use `filtered` internally, those optics are only lawful if data written back through them satisfies the predicate. Since their old type signature meant they were only usable for matching and nobody has complained, the ability to write back through them has probably not been missed.

This means we'll have to regenerate everything again, so update botocore and configure new services too. Note that this time existing code will continue to work, but I'll aim to get the regeneration up ASAP.


axman6 (approved):
LGTM, should stop people doing silly things.

Issue #880, opened 2023-01-09
`amazonka-emr-containers` and `amazonka-emr-serverless` do not compile


endgame:
This is caused by the switch from `HashMap` to `Map` in #825. I _think_ it's because the affected shape definitions are now full of `Ptr` instead of real structures, so the code which decides which instances to derive no longer excludes `Read` where it previously did.

PR #881, opened 2023-01-10
s3: make the timestamps use rfc822

### LINKED ISSUE #879: Proper formatting for `if-modified-since` timestamps in S3
Currently, amazonka sends the `if-modified-since` header in `GetObject` (and `HeadObject`) as a ISO8601 date (see <https://amazonka.brendanhay.nz/docs/libZSservicesZSamazonka-s3ZSamazonka-s3/Amazonka-S3-GetObject.html#t:GetObject>), but S3 seems to expect a RFC822-formatted date:
- that's the format the `aws` cli uses, where `if-modified-since` works as expected, whereas amazonka sends a iso8601-formatted date and the `if-modified-since` header is ignored
- that's the format present in the S3 api response (the parser for `Time` values is lenient and tries different formats, so the values are correctly parsed, even though they're declared as `ISO8601` in the response objects as well.

I'm not sure how the time format is selected during codegen, so i don't know how to fix it myself. If that's not too complicated, i can work on a PR

divarvel:
** depends on #882 **

** todo: regen S3 after #882 is merged and a tree-wide regen is done **

AWS does not explicitly specify that it is the required format, but RFC3339-formatted dates are ignored, all dates are formatted with RFC822 in responses, and examples also use RFC822 dates. Considering the age of S3, it makes sense.

Fixes #879

As for annotating S3 shapes, i've split the change in two commits. The first one is for fields where i'm reasonably sure it's okay (because i've tested it), the second one for fields where i'm less sure, even though it makes sense. All cache related headers tend to be RFC822-formatted since they are specified by HTTP RFCs, which require this format.

### DIFF
--- a/configs/annexes/s3.json
+++ b/configs/annexes/s3.json
@@ -4,14 +4,35 @@
     },
     "documentation": "Amazon Simple Storage Service is storage for the Internet. Amazon S3 has a simple web services interface that you can use to store and retrieve any amount of data, at any time, from anywhere on the web. It gives any developer access to the same highly scalable, reliable, fast, inexpensive data storage infrastructure that Amazon uses to run its own global network of web sites. The service aims to maximize benefits of scale and to pass those benefits on to developers.",
     "shapes": {
+        "AbortDate": {
+            "timestampFormat": "rfc822"
+        },
         "CompletedPartList": {
             "min": 1
+        },
+        "CopySourceIfModifiedSince": {
+            "timestampFormat": "rfc822"
+        },
+        "CopySourceIfUnmodifiedSince": {
+            "timestampFormat": "rfc822"
+        },
+        "Expires": {
+            "timestampFormat": "rfc822"
         },
         "GetBucketLocationOutput": {
             "required": [
                 "LocationConstraint"
             ],
             "payload": "LocationConstraint"
+        },
+        "IfModifiedSince": {
+            "timestampFormat": "rfc822"
+        },
+        "IfUnmodifiedSince": {
+            "timestampFormat": "rfc822"
+        },
+        "LastModified": {
+            "timestampFormat": "rfc822"
         },
         "ObjectStorageClass": {
             "enum": [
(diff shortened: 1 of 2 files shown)


divarvel:
Running `./scripts/generate --commit s3` outputs a bunch of stuff (most notably, it says that it wrote .hs files in a temp directory), but no commit is generated (and the working copy is not affected either).

So i'm not sure whether i'm holding `scripts/generate` wrong or editing the `s3.json` in a way that does not affect the codegen output.

endgame:
AFAICT you have written the annex correctly and that neither existing configs nor botocore set it in service metadata. That field should be moved from service metadata to the parser for individual types - try adding a `Timestamp` field to the `Time` constructor of `Lit` in `Ann.hs`, and fixing up the `instance FromJSON (ShapeF ())`.

divarvel:
Well, that worked. I had to move `Timestamp` to `Ann.hs` and had the `Time` constructor take a `Maybe Timestamp`.

Now that i have something that works, i'll dig a bit more to see if i can annotate other shapes.

endgame:
Yes, I think a tree-wide rebase is in order. Can you please:

1. Remove the s3 regeneration from this branch
2. Add an entry to `lib/amazonka/CHANGELOG.md` (which is the master changelog for the project)

I will merge #878 and regenerate everything once that's done. I'd hoped to get a fix in for #880 before I next ran the generator, but that seems a lot harder than I first thought, so we'll gen once the S3 fixes are in.

divarvel:
Alrighty, will do in the morning (CET)

endgame:
I think we should remove the `timestampFormat` field from the `Metadata` type, and hardcode the fallback in `Syntax.hs` - nothing in botocore declares anything there, and neither do any of our annexes.

endgame:
I think this will entail moving the check from `substitute` (`Subst.hs`) into `Syntax.directed`, which should be fine if you specialise the `HasMetadata a Identity` into `Service Identity a b c`. Then you can remove the `timestampFormat` stuff from the metadata record - I've never seen that field populated in botocore or in any of our annexes.

endgame (approved):
The gen changes will leave the diff after #882 gets merged, so I think the only other thing this needs is a changelog entry.

divarvel:
I have added a changelog entry for this one, i'll add a changelog entry for #882 as well.

endgame (approved):

PR #882, opened 2023-01-11
Per shape timestamp format

divarvel:
As discovered in #881

- `timestampFormat` was not supported in shapes definitions
- `timestampFormat` is supported in service metadata but not used anywhere (neither botocore nor amazonka annexes) (it was quickly checked by turning it into a `Maybe Void`; no parse error was triggered)

This PR adds support for reading it in shape definition and removes support from reading it from service metadata.

### DIFF
--- a/gen/src/Gen/AST/Subst.hs
+++ b/gen/src/Gen/AST/Subst.hs
@@ -43,13 +43,7 @@ substitute svc@Service {..} = do
   where
     meta :: Metadata Maybe -> Metadata Identity
     meta m@Metadata {..} =
-      m
-        { _timestampFormat = Identity ts,
-          _checksumFormat = _checksumFormat .! SHA256
-        }
-
-    ts :: Timestamp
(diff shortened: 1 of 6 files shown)


gen/src/Gen/AST/Subst.hs
```diff
@@ -44,13 +44,9 @@ substitute svc@Service {..} = do
     meta :: Metadata Maybe -> Metadata Identity
     meta m@Metadata {..} =
       m
-        { _timestampFormat = Identity ts,
-          _checksumFormat = _checksumFormat .! SHA256
+        { _checksumFormat = _checksumFormat .! SHA256
         }
```
endgame, line 48: Fold to one line?

endgame (approved):
If you can knock up a quick changelog entry in `lib/CHANGELOG.md`, I'll merge this.

gen/src/Gen/AST/Subst.hs
divarvel: done

PR #885, opened 2023-01-12
do not force MonadIO on presign

avieth:


lib/amazonka/src/Amazonka/Presign.hs
```diff
@@ -11,6 +11,9 @@
 --
```
endgame, line 11: Please update the copyright year in the haddock, and remove references to `'MonadIO'` (removed here) and to `AWS specific transformers` (which have been gone for a long time).

lib/amazonka/src/Amazonka/Presign.hs
```diff
@@ -11,6 +11,9 @@
...
 -- It is intended for use directly with "Amazonka.Auth" when only presigning
```
endgame, line 12:
```suggestion
-- It is intended for use directly with 'Amazonka.Auth.Auth' when only presigning
```

lib/amazonka/src/Amazonka/Presign.hs
```diff
@@ -11,6 +11,9 @@
...
 -- and no other AWS actions are required.
```
endgame, line 13:
```suggestion
-- and no other AWS actions are required. See 'Amazonka.Auth.withAuth' to
-- extract an 'AuthEnv' from an 'Auth'
```

endgame (changes requested):
Since `withAuth` still has to have a `MonadIO` constraint, I'm curious what this is buying you. I see it's coming from a workplace fork, which means it's proven necessary for something, so I'm happy to merge it with a couple of tweaks.

In addition to the suggestions inline, there's a couple more changes I'll need to see before I can merge it, but those are minor:

- [ ] Add an entry to `lib/amazonka/CHANGELOG.md`
- [x] Re-export `withAuth` from `Amazonka.Auth` (as a general rule, most normal users should not have to depend directly on `amazonka-core`)

endgame:
I agree that it's a better factoring, and I hope we can merge it before I tag RC2.

avieth:
Should be good to go now. LMK if I missed anything.

lib/amazonka/src/Amazonka/Presign.hs
```diff
@@ -6,11 +6,12 @@
 -- Stability   : provisional
 -- Portability : non-portable (GHC extensions)
 --
--- This module contains functions for presigning requests using 'MonadIO' and
--- not one of the AWS specific transformers.
---
--- It is intended for use directly with "Amazonka.Auth" when only presigning
--- and no other AWS actions are required.
+-- It is intended for use directly with "Amazonka.Auth.Auth" when only
```
endgame, line 9:
```suggestion
-- It is intended for use directly with 'Amazonka.Auth.Auth' when only
```

lib/amazonka/src/Amazonka/Presign.hs
```diff
@@ -6,11 +6,12 @@
 -- Stability   : provisional
```
endgame, line 6: Please update copyright year in the haddock.

endgame (changes requested):
Couple of final nits, but I look forward to merging this today if you have time to make the changes.

endgame (approved):
Thank you very much.

PR #892, opened 2023-01-19
amazonka-core: Remove another call to deprecated `iso8601DateFormat`

endgame:


lib/amazonka-core/src/Amazonka/Data/Time.hs
```diff
@@ -80,7 +80,7 @@ instance TimeFormat RFC822 where
   format _ = "%a, %d %b %Y %H:%M:%S %Z"
 
 instance TimeFormat ISO8601 where
-  format _ = iso8601DateFormat (Just "%XZ")
+  format _ = "FT%XZ"
```
ysangkok, line 83: Shouldn't it be `%F`? I suppose the `F` is meant to be `%Y-%m-%d`?

endgame: Argh. Yes.

Issue #739, opened 2021-12-16
Tracking issue: spurious `Maybe`s in request/response bodies


endgame:
At some point, the generator was changed to emit `Maybe` values for all optional fields, even for types that had a sensible "empty" value (e.g., maps and lists). Among other benefits, this made it possible to detect empty lists and maps in DynamoDB `AttributeValue`s (before #724 introduced an AST and solved things properly).

There's a big QoL downside to this, particularly for responses: it makes a lot of request/response fields look optional, and when handling responses you're forced to write partial code or handle errors that should never appear.

I've tried raising issues for these in botocore, but AWS service teams have a habit of not being very thorough when marking things as required in their service definitions (and changing this risks backwards-compat problems now).

Fixing these is not that difficult; it's just an additional `requiredFields` entry in `configs/$SERVICE.json` and would make a good first PR. If you make an issue or PR about one of these, please:

1. Mention this issue;
2. Provide proof that the field is always present (in the form of AWS documentation or a link to code in the SDK); and
3. Tag it [`service config`](https://github.com/brendanhay/amazonka/labels/service%20config)

ysangkok:
I am a bit unsure about the Value field on the Parameter type of AWS Systems Manager (formerly known as SSM).

It is required according to docs on
    https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ssm-parameter.html
    
These docs don't mention anything about optionality, but they do mention the field and supply it in their examples:
    
- Systems Manager API reference for get_parameter method
      https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_GetParameter.html
- Ruby Parameter docs
      https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/SSM/Types/Parameter.html
- Python get_parameter docs
      https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ssm.html#SSM.Client.get_parameter
    
In the Go bindings, it is a pointer, which suggests that it is optional. But it seems like everything is a pointer.
    https://docs.aws.amazon.com/sdk-for-go/api/service/ssm/#Parameter
    
This piece of Parameter documentation claims it is not required:
    https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_Parameter.html

So what to do? Don't know whether I should make a PR.

endgame:
[...]

PR #896, opened 2023-02-17
Change Item in GetItemResponse to optional

### LINKED ISSUE #895: Item in GetItemResponse is not optional
The item in `GetItemResponse` is declared as
```
item :: Prelude.HashMap Prelude.Text AttributeValue
```

But the [AWS Doc](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_GetItem.html) says:
> If there is no matching item, GetItem does not return any data and there will be no Item element in the response.

Currently when an item is not found, I need to use `resp ^. #item == mempty` to check for that. Will it be better to use `Maybe`, or document this behavior?

kokobd:
The description at the top of [this page](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_GetItem.html#API_GetItem_ResponseElements) says:
> If there is no matching item, GetItem does not return any data and there will be no Item element in the response.

`Item` was marked as "required" by [this commit](https://github.com/brendanhay/amazonka/commit/c8e3ac72d1d1dc17ebc7e831872638d194cbe2b4). @endgame Do you know if there are some reasons behind that change?

Fixes #895

### DIFF
--- a/configs/services/dynamodb.json
+++ b/configs/services/dynamodb.json
@@ -35,11 +35,6 @@
                 "underive": []
             }
         },
-        "GetItemOutput": {
-            "requiredFields": [
-                "Item"
-            ]
-        },
         "PutRequest": {
             "replacedBy": {
                 "name": "",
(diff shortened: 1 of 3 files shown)


endgame:
If AWS omits that field when there's no item found, then our marking it "required" is in error. Please run `scripts/generate --commit dynamodb` and push the commit that it generates.

kokobd:
Regenerated. Should we mention in #739 that we SHOULD generate?
> Please don't regenerate services - we do that separately.

endgame:
I should probably revise that. For generator changes or treewide updates, we do a separate PR to keep the diff clean.

PR #900, opened 2023-04-18
Adding ap-south-2 (Hyderabad) region

BharatKJain:
Added quick fixes for supporting ap-south-2 region, we might have to add support in other places also.

Questions:
- Should we add ap-south-2 support in other places too? Like comments in Chime and Sagemaker lambda ARNs

  Example:
  `    -- -   @arn:aws:lambda:ap-south-1:565803892007:function:ACS-NamedEntityRecognition@`

  `    -- | The Region in which you create the meeting. Available values:
    -- @af-south-1@, @ap-northeast-1@, @ap-northeast-2@, @ap-south-1@,
    -- @ap-southeast-1@, @ap-southeast-2@, @ca-central-1@, @eu-central-1@,
    -- @eu-north-1@, @eu-south-1@, @eu-west-1@, @eu-west-2@, @eu-west-3@,
    -- @sa-east-1@, @us-east-1@, @us-east-2@, @us-west-1@, @us-west-2@.`

Some notes:
- I have not added Hyderabad in the following files because AWS has deprecated the service principals for them:
  - lib/services/amazonka-redshift/src/Amazonka/Redshift/Internal.hs
  - lib/services/amazonka-elb/src/Amazonka/ELB/Internal.hs

### DIFF
--- a/lib/services/amazonka-redshift/src/Amazonka/Redshift/Internal.hs
+++ b/lib/services/amazonka-redshift/src/Amazonka/Redshift/Internal.hs
@@ -72,6 +72,7 @@ getCloudTrailAccountId = \case
   Jakarta -> Just "623197973179"
   Melbourne -> Just "945512339897"
   Mumbai -> Just "408097707231"
+  Hyderabad -> Just "297058826802"
   Osaka -> Just "398671365691"
   Seoul -> Just "713597048934"
   Singapore -> Just "960118270566"
--- a/lib/amazonka-core/src/Amazonka/Types.hs
+++ b/lib/amazonka-core/src/Amazonka/Types.hs
@@ -132,6 +132,7 @@ module Amazonka.Types
         Jakarta,
         Melbourne,
         Mumbai,
+        Hyderabad,
         Osaka,
         Seoul,
         Singapore,
@@ -880,6 +881,9 @@ pattern Melbourne = Region' "ap-southeast-4"
 pattern Mumbai :: Region
 pattern Mumbai = Region' "ap-south-1"
 
+pattern Hyderabad :: Region
+pattern Hyderabad = Region' "ap-south-2"
+
 pattern Osaka :: Region
 pattern Osaka = Region' "ap-northeast-3"
 
@@ -966,6 +970,7 @@ pattern Ningxia = Region' "cn-northwest-1"
   Jakarta,
   Melbourne,
   Mumbai,
+  Hyderabad,
   Osaka,
   Seoul,
   Singapore,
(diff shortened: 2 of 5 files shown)


lib/services/amazonka-redshift/src/Amazonka/Redshift/Internal.hs
```diff
@@ -72,6 +72,7 @@ getCloudTrailAccountId = \case
   Jakarta -> Just "623197973179"
   Melbourne -> Just "945512339897"
   Mumbai -> Just "408097707231"
+  Hyderabad -> Just "297058826802"
```
aravindgopall, line 75: can you add the same in getAccountId function `37` line too.

aravindgopall (approved):

brendanhay (approved):
LGTM, thanks!

endgame:
Thanks for this. Would you be able to add an entry to `lib/amazonka/CHANGELOG.md` next to the one for Melbourne?

brendanhay:
> Should we add ap-south-2 support in other places too? Like comments in Chime and Sagemaker lambda ARNs

Note: those specific comments are auto-generated - so the regions are omitted in AWS' own models/documentation, I'd just leave them out in this case.

BharatKJain:
> Thanks for this. Would you be able to add an entry to `lib/amazonka/CHANGELOG.md` next to the one for Melbourne?

Added.

BharatKJain:
Ah sorry, missed the link @endgame 😓

endgame:
No trouble at all. Figured it would be easier if I added it instead of waiting for another round of commits. Thanks for the PR!

BharatKJain:
@endgame @brendanhay Is it possible to publish a new version of sdk? (1.6.2?)

AWS just released Hyderabad region few weeks back, we wanted to use it in our org, maybe other folks might get blocked on this in future while testing this region.

endgame:
We hope to announce another release candidate of Amazonka 2.0 Soon™, because all the features and issues assigned to the milestone are essentially done. That needs to happen before 2.0 final goes to Hackage.

Until then, you should use `main`. Instructions are available at the release announcement for 2.0 RC 1: https://github.com/brendanhay/amazonka/discussions/716

brendanhay:
If you’ve been using a fork of `1.6.x` with the region update, I’d also recommended to do what @endgame suggested above regarding `main` as there a number of changes which you’ll have to contend with when `2.x` lands “soon ™️”.

BharatKJain:
Hmm, I made a mistake in this PR, `ap-south-2` was already present, somehow I overlooked, anyways I checked the current main and it's not having this issue, so it'a fine, I don't have to create a fix PR.

PR #917, opened 2023-06-27
Nixpkgs update gen

endgame:

Issue #888, opened 2023-01-17
Deriving clauses of `Ptr` are not calculated correctly


endgame:
In `amazonka-emr-containers`, there is the following set of mutually-recursive shapes:

* `Configuration` contains `ConfigurationList` and a `Sensitive` field.
* `ConfigurationList` contains `Configuration`

Under #862 , the map of shapes was populated in such a way that `ConfigurationList` held a `Ptr` to `Configuration`, which meant that it generated a `deriving` clause containing a static set of class names, including `Read`. This fails to compile, as `Configuration` contains a `Sensitive` field, and so cannot have a `Read` instance.

The correct fix is to determine the set of deriving clauses by constraint solving — https://github.com/endgame/amazonka/commit/bdcba74b64d536d303be99cf9ecdd781766c770d was an attempt at this, but I am abandoning it after a week and a half because it's still broken. Supporting #872 will require a serious re-architecting of the generator anyway, so let's do it properly:

* Build out a `botocore` library which is a direct representation of the data files provided by AWS - service, paginators, waiters, and the new endpoint-rule-set type. `botocore` will need some sort of "JSON deep-merge" functionality to support the splicing in of annexes.
* Rewrite the generator to use `botocore`, and differentiate each stage (classifying shapes; applying renames/overrides; generating code)
* Consider a more modern lib than `haskell-src-exts` for code generation (maybe `ghc-exactprint`?)
* Kill the cofree comonad of shapes once and for all. it was a defensible idea when the inclusion graph of shape definitions was a DAG, but that has not been true for a very long time. If every shape reference is handled as if it were a `Ptr` by looking it up in a `MonadReader` context, edge cases like this will stop happening.

endgame:
A possible workaround could be to allow the use of `"underive"` without `"replace"`?

PR #918, opened 2023-06-27
Nixpkgs update

endgame:
Update nixpkgs, update botocore, configure new services. Regenerations are in https://github.com/brendanhay/amazonka/pull/917

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,7 +29,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: cachix/install-nix-action@v20
+      - uses: cachix/install-nix-action@v22
         with:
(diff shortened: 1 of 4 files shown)


brendanhay:
LGTM

endgame:
Annoyingly, the regeneration of `amazonka-fsx` (and maybe others?) falls afoul of #888 . I am testing a build of everything locally with only the nixpkgs bump and it it works, I'll force-push it.

endgame:
Even sticking with the same botocore, `gen` makes dodgy service bindings for `amazonka-kendra`: the type prefixes mysteriously disappear for some lenses. I suspect more `Ptr` shenanigans; in the interests of keeping things moving (things are similarly broken on current `main`), I will merge this, not merge any regenerations, and defer fixing this until we can fix the `Ptr` problems properly.

PR #920, opened 2023-06-27
Switch from `cryptonite` to `crypton`

### LINKED ISSUE #912: Support `crypton`
`cryptonite` has been forked to `crypton` for ongoing maintenance. We should support it with a flag, defaulting to `crypton` and falling back to `cryptonite`.

endgame:
Fixes #912


endgame:
We're using this at work and it seems to be fine.

PR #921, opened 2023-06-28
gen: Use hashable ==1.3.* when generating bindings

### LINKED ISSUE #919: amazonka-kendra generates lenses without prefixes
`scripts/generate kendra` generates service bindings which do not compile, because some of the generated lenses lose the type prefix. I suspect something involving a `Ptr`, but I'm not sure what. Until this is fixed, we cannot generate new versions of service bindings.

endgame:
Workaround e.g. #919 when generating, so we can potentially bump botocore and/or continue nursing the current generator along until we get to making something better.

Closes #919

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -48,7 +48,7 @@
         ghc94 = pkgs.haskell.packages."ghc94";
         ghc96 = pkgs.haskell.packages."ghc96";
 
-        # The default ghc to use when entering `nix develop` and building amazonka-gen.
+        # The default ghc to use when entering `nix develop`.
         ghcDefault = ghc94;
 
         renameVersion = version: "ghc" + (pkgs.lib.replaceStrings [ "." ] [ "" ] version);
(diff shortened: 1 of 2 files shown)


brendanhay:
🤦🏽 Nice find. Given aeson's `ordered-keymap` flag is enabled by default the generator should probably switch over to use `containers/Map` directly, at some point.

endgame:
Long-term, I want to build a binding to `botocore` which uses `waargonaut` so I can track declaration order of fields in botocore shapes. Because AWS API Gateway supports IAM authentication (where you SigV4-sign requests with the `execute-api` service abbrev), I want a version of the generator which consumes a service description that isn't coupled to `botocore`, so that othe people can generate amazonka-style bindings to their APIs. Then, we'll need a bridge between botocore and this new input format. More details at #888.

All that has to happen after 2.0, and so I'll merge this now to unblock a) updating botocore, and b) some other changes to the generator which will be necessary pre-2.0.

PR #936, opened 2023-07-17
gen: Make it build against GHC9.6

endgame:


endgame:
I regenerated the entire repository with this change and it didn't alter the files at all, so I think it's good.

PR #938, opened 2023-07-18
Aeson 2.2

### LINKED ISSUE #923: aeson-2.2
This is going to be a bit annoying - we have to wait for everything to catch up, and `aeson` changed its module layout, breaking at least `http-conduit`, so I can't just `allow-newer` my way through it all.

It's also a bit annoying that because cabal wants a single set of packages for the whole project, `pandoc`'s bounds constrain what versions I can test for the rest of the project.

endgame:
And hlint `amazonka-core` and `amazonka` while I'm here.

Fixes #923

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -1,0 +1,10 @@
+# Ignore camel-case warnings for modules containing explicit lenses
+- ignore:
+    name: "Use camelCase"
+    within:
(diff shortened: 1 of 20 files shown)


endgame:
This passed `cabal build all && cabal test all` with an `aeson >=2.2` constraint in `cabal.project`, so I'm merging it.

PR #940, opened 2023-07-21
Migration guide

### LINKED ISSUE #939: 1.6.1 -> 2.0 Migration Guide
Write some text into the README about how to migrate across, for people who haven't been following `main` and who shouldn't be expected to piece it together from changelog entries.

endgame:
Fixes #939 

Question: Is it still worth exporting `Amazonka.{dynamic,metadata,userdata}` from the top-level module, since the metadata interface is mainly for when you don't have an `Env`? We also only export the top level of IMDS keys, which seems silly.

### DIFF
--- a/lib/amazonka/CHANGELOG.md
+++ b/lib/amazonka/CHANGELOG.md
@@ -3,7 +3,6 @@
 ## [2.0.0](https://github.com/brendanhay/amazonka/tree/2.0.0)
 Released: **?**, Compare: [2.0.0-rc2](https://github.com/brendanhay/amazonka/compare/2.0.0-rc2...2.0.0)
 
-### Major Changes
 ### Changed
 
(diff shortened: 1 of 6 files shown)


brendanhay:
LGTM. I don't feel particularly strongly about the metadata functions being exported or not. Given they are currently, I'd be inclined to leave them that way to not surprise any RC users that might be using them somewhere.

brendanhay (approved):

endgame:
> LGTM. I don't feel particularly strongly about the metadata functions being exported or not. Given they are currently, I'd be inclined to leave them that way to not surprise any RC users that might be using them somewhere.

I think it is a very niche use-case, and now that we support all metadata keys described by AWS, the stuff re-exported from `Amazonka` doesn't cover anywhere near all of them. Also, those functions take an `Env` instead of a `Manager`, which is a bit weird. I made sure to call this change out in the migration guide.

If you're happy with this and CI passes cleanly, feel free to hit merge.

endgame:
CI has passed.

Issue #948, opened 2023-07-28
No support for AWS_CONFIG_FILE, AWS_SHARED_CREDENTIALS_FILE


flokli:
I have `$AWS_CONFIG_FILE` pointing to another location than ~/.aws/config, but it seems amazonka isn't picking up that location.

flokli:
It also looks like it's not supporting AWS_SHARED_CREDENTIALS_FILE, and skipping config files alltogether if there's no `~/.aws/credentials` file around (even though it's not necessary for SSO-style logins)

flokli:
I was able to workaround this by copying my file from `$AWS_CONFIG_FILE` to `~/.aws/config`, and creating an empty `~/.aws/credentials` file.

arianvp:
Sounds like we should make the Credentials file optional. Small fix

endgame:
Does a description exist in an AWS SDK manual of what correct behaviour should look like? If not, maybe there's a clear link inside one of the SDK repos which we can reference?

And for a potential workaround: you don't have to use `discover`, you can use the other functions in `Amazonka.Auth` to find credentials in any way you please.

arianvp:
https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/config@v1.18.28#EnvConfig is how the official  Go SDK does it. which is in line with this issue.

arianvp:
I'm happy to send a PR to implement this behaviour

endgame:
That would be excellent. I recommend starting by reading under `lib/amazonka/src/Amazonka/Auth/*`, hopefully you shouldn't have to get too far into the weeds to understand that.

arianvp:
Also just random  drive-by. Thanks for reviving this project. Seems you put a lot of effort in the 2.0 release. Do you have a way I can send some love your way @endgame ? You do not seem to have Github Sponsors set up.

endgame:
Thank you, it was a lot hard work. While a good chunk of the 2.0 revival was done in my personal time, many of the meatier features in 2.0 RC 2 (fixing the generator's handling of recursive shape definitions, the hooks system, the authentication rewrite) were done on my employer's time, with their blessing. I don't think it would be appropriate for me to take money for that, so I would instead suggest paying it forward: support someone else working on a load-bearing project who isn't as lucky as I am.

PR #951, opened 2023-07-31
Allow config files without credential files and allow overriding paths

arianvp:
This fixes the issues marked in #948


lib/amazonka/src/Amazonka/Auth/ConfigFile.hs
```diff
@@ -60,7 +86,14 @@ fromFilePath ::
   Env' withAuth ->
   m Env
 fromFilePath profile credentialsFile configFile env = liftIO $ do
```
arianvp, line 88: It's  probably more "correct" to change the `credentialsFile` parameter to `Maybe FilePath` but I wanted  to avoid a breaking change

arianvp:
cc @flokli

lib/amazonka/src/Amazonka/Auth/ConfigFile.hs
flokli: I think it's actually more user-friendly this way.

Applications using amazonka might not be aware if credentials are needed (as this requires parsing the config file), and that way you only try to read it when needed.

flokli:
I gave this a try on a system:

 - using `AWS_CONFIG_FILE` pointing to elsewhere
 - no `~/.aws/credentials` file being present
 - and a (legacy) SSO config being present at the location in `AWS_CONFIG_FILE`.

It worked like a charm! Thanks :heart:

lib/amazonka/src/Amazonka/Auth/ConfigFile.hs
```diff
@@ -36,7 +36,36 @@ import System.Info (os)
...
+--   (Formerly AWS SSO) by setting @sso_start_url@, @sso_region@, 
+--   @sso_account_id@, and @sso_role_name@ in your profile section.
+--   Amazonka currently does not initiate the SSO login flow, so you will have
+--   to do that yourself using the awscli. Amazonka will then look in
+--   @~\/.aws\/sso\/cache@ for a cached token.
+--
+--  * We currently only support 'Legacy' SSO profiles and do not support
+--   setting common SSO settings in a @[sso-session <name>]@ section or
+--   support token refresh. So use the following guide to set up your awscli:
+--   https://docs.aws.amazon.com/cli/latest/userguide/sso-configure-profile-legacy.html
```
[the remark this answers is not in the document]
arianvp: Done https://github.com/brendanhay/amazonka/issues/953

arianvp:
Please have another look! I added a CHANGELOG

lib/amazonka/src/Amazonka/Auth/ConfigFile.hs
```diff
@@ -253,21 +302,39 @@ data CredentialSource = Environment | Ec2InstanceMetadata | EcsContainer
...
 -- This looks in in the @HOME@ directory as determined by the
 -- <http://hackage.haskell.org/package/directory directory> library.
 --
+
 -- * Not Windows: @$HOME\/.aws\/credentials@
+
 --
+
 -- * Windows: @%USERPROFILE%\\.aws\\credentials@
+
```
endgame, line 321: Please remove the blank lines in this section.

lib/amazonka/src/Amazonka/Auth/ConfigFile.hs
```diff
@@ -253,21 +302,39 @@ data CredentialSource = Environment | Ec2InstanceMetadata | EcsContainer
 --
 -- Throws 'MissingFileError' if 'credFile' is missing, or 'InvalidFileError'
 -- if an error occurs during parsing.
+
```
endgame, line 305:
```suggestion
--
```

endgame (commented):
Sorry for the delay. I think your logic is right and thank you for setting up a changelog which matches the existing format.

A couple of whitespace nits and then I think this is good to merge.

lib/amazonka/src/Amazonka/Auth/ConfigFile.hs
arianvp: huh idk where these came from. ormolu maybe went rogue

endgame:
Thank you for your contribution and for your patience. I'll let CI run and look at merging this tomorrow.

endgame:
CI failures are GHC panics which look like some sort of cache misbehaviour. @brendanhay can you please kick the caches again?

PR #952, opened 2023-07-31
[#947] Use `utfEncode'` to log bytestrings

### LINKED ISSUE #947: When the payload is a non-printable bytestring, printing the request terminates the app
Recently [we found out](https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_PutMetricData.html) that `PutMetricData` endpoint supports gzipped request bodies.

So, to be able to make use of that, I came up with this wrapper:

``` haskell
import qualified Amazonka
import qualified Amazonka.CloudWatch as CW
import qualified Amazonka.CloudWatch.Lens as CW
import qualified Amazonka.Data as AWS
import qualified Amazonka.Request as AWSRequest
import qualified Amazonka.Response as AWS
import qualified Amazonka.Types (AWSRequest, Request (..))
import qualified Codec.Compression.GZip as GZip

newtype GzippedPutMetricData = GzippedPutMetricData CW.PutMetricData
  deriving newtype (AWS.ToPath, AWS.ToQuery, AWS.ToHeaders)

instance Amazonka.Types.AWSRequest GzippedPutMetricData where
  type AWSResponse GzippedPutMetricData = CW.PutMetricDataResponse
  request overrides x =
    let req = (AWSRequest.postQuery (overrides CW.defaultService) x :: Amazonka.Types.Request GzippedPutMetricData)
        !bs = BS.fromStrict (AWS.toBS (AWS.toQuery x))
        !gzippedBody = GZip.compress bs
        setContentEncoding = AWS.hdr hContentEncoding "gzip"
     in req
              { Amazonka.Types.body = AWS.toBody gzippedBody,
                Amazonka.Types.headers = setContentEncoding (Amazonka.Types.headers req)
              }
  response = AWS.receiveNull CW.PutMetricDataResponse'
```
which works nicely. But, when running on debug logging, Amazonka [prints](https://github.com/brendanhay/amazonka/blob/6d8b87f622a31b0fce3f169633b08ac3b6bf0182/lib/amazonka-core/src/Amazonka/Data/Log.hs#L136-L150) payloads in size smaller than 4KBs.

Hence, the above wrapper fails if the compressed payload is smaller than 4KB, **and** if we're running it with DEBUG logging. The ad-hoc solution I found for this is to not compress the payload if the raw payload size is less than 100KBs:

``` haskell
newtype GzippedPutMetricData = GzippedPutMetricData CW.PutMetricData
  deriving newtype (AWS.ToPath, AWS.ToQuery, AWS.ToHeaders)

instance Amazonka.Types.AWSRequest GzippedPutMetricData where
  type AWSResponse GzippedPutMetricData = CW.PutMetricDataResponse
  request overrides x =
    let req = (AWSRequest.postQuery (overrides CW.defaultService) x :: Amazonka.Types.Request GzippedPutMetricData)
        !bs = BS.fromStrict (AWS.toBS (AWS.toQuery x))
        !gzippedBody = GZip.compress bs
        setContentEncoding = AWS.hdr hContentEncoding "gzip"
     in if LBS.length bs <= 100000
          then req {Amazonka.Types.body = AWS.toBody bs}
          else
            req
              { Amazonka.Types.body = AWS.toBody gzippedBody,
                Amazonka.Types.headers = setContentEncoding (Amazonka.Types.headers req)
              }
  response = AWS.receiveNull CW.PutMetricDataResponse'
```

Is there a way to make the `ToLog` instance of `ByteString` (or `ByteStringLazy`) not print contents if it's not printable?

korayal:
closes #947

### DIFF
--- a/lib/amazonka-core/src/Amazonka/Data/Log.hs
+++ b/lib/amazonka-core/src/Amazonka/Data/Log.hs
@@ -34,10 +34,14 @@ instance ToLog ByteStringBuilder where
   build = id
 
 instance ToLog ByteStringLazy where
-  build = Build.lazyByteString
+  build lbs = case LText.decodeUtf8' lbs of
+    Left _ -> "non-printable " <> build (LBS.length lbs) <> " lazy bytes"
+    Right _ -> Build.lazyByteString lbs
 
 instance ToLog ByteString where
-  build = Build.byteString
+  build bs = case Text.decodeUtf8' bs of
+    Left _ -> "non-printable " <> build (BS.length bs) <> " strict bytes"
+    Right _ -> Build.byteString bs
 
 instance ToLog Int where
   build = Build.intDec


lib/amazonka-core/src/Amazonka/Data/Log.hs
```diff
@@ -34,10 +34,14 @@ instance ToLog ByteStringBuilder where
   build = id
 
 instance ToLog ByteStringLazy where
-  build = Build.lazyByteString
+  build lbs = case LText.decodeUtf8' lbs of
+    Left _ -> "non-printable " <> build (LBS.length lbs) <> " lazy bytes"
```
endgame, line 38:
```suggestion
    Left _ -> mconcat ["non-printable lazy ByteString (" <> build (LBS.length lbs) <> " bytes)"]
```
And similarly below.

endgame (commented):
Can you please add an entry to `lib/amazonka/CHANGELOG.md`, which is the master changelog for the project?

Also, are we sure that this is the correct check to make? What if the user's terminal is set to some other encoding scheme? A TextEncoding (as returned by `System.IO.hGetEncoding`) isn't even in `Eq`, so I'm not sure what the best way is to fix this. I suspect what we have here might be a bit too naive; do you have any thoughts about other ways we could check this? It may also be the case that we successfully decode UTF-8 but it's got non-printable characters in it; do we want to do that check here too?

In the medium term, I probably want to extract logging from the main amazonka library so that people can use whatever logging framework suits them. Then handling non-printable stuff becomes the log library's concern.

korayal:
> Can you please add an entry to lib/amazonka/CHANGELOG.md, which is the master changelog for the project?

I've updated the changelog and added a section `Unreleased` at the top. would that work?

korayal:
> Also, are we sure that this is the correct check to make? What if the user's terminal is set to some other encoding scheme? A TextEncoding (as returned by System.IO.hGetEncoding) isn't even in Eq, so I'm not sure what the best way is to fix this. I suspect what we have here might be a bit too naive; do you have any thoughts about other ways we could check this? It may also be the case that we successfully decode UTF-8 but it's got non-printable characters in it; do we want to do that check here too?

I've added another check with `isPrint`, where it should ensure that whatever we're trying to print contains printable characters. At this point I think assuming the request body (if not compressed) is a valid utf-8 text since it's what we're sending to AWS. The part that prints it out is still left to `Build.lazyByteString` / `Build.byteString`, so at worst case we're keeping the existing behavior.

> In the medium term, I probably want to extract logging from the main amazonka library so that people can use whatever logging framework suits them. Then handling non-printable stuff becomes the log library's concern.

that'd be great.

lib/amazonka-core/src/Amazonka/Data/Log.hs
```diff
@@ -34,10 +35,14 @@ instance ToLog ByteStringBuilder where
   build = id
 
 instance ToLog ByteStringLazy where
-  build = Build.lazyByteString
+  build lbs = case LText.decodeUtf8' lbs of
+    Right t | all Char.isPrint (LText.unpack t) -> Build.lazyByteString lbs
```
endgame, line 39: `text` defines its own versions of `all`, so you don't need to unpack.

lib/amazonka/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Change Log
 
+## Unreleased
+
+- `amazonka`: During DEBUG logging, do not print bytestrings that have unprintable characters [\#952](https://github.com/brendanhay/amazonka/pull/952)
```
endgame, line 5: Please put the PR link on the next line to match existing entries. I'd also remove the word "DEBUG", since we're fixing this by changing typeclasses.

lib/amazonka-core/src/Amazonka/Data/Log.hs
korayal: replaced. thanks!

lib/amazonka/CHANGELOG.md
korayal: done.

korayal:
@endgame are we OK with the latest state of this PR?

lib/amazonka-core/src/Amazonka/Data/Log.hs
```diff
@@ -34,10 +35,14 @@ instance ToLog ByteStringBuilder where
...
-  build = Build.lazyByteString
+  build lbs = case LText.decodeUtf8' lbs of
+    Right lt | LText.all Char.isPrint lt -> Build.lazyByteString lbs
+    _ -> mconcat ["non-printable lazy ByteString (", build (LBS.length lbs), ")"]
 
 instance ToLog ByteString where
-  build = Build.byteString
+  build bs = case Text.decodeUtf8' bs of
+    Right t | Text.all Char.isPrint t -> Build.byteString bs
+    _ -> mconcat ["non-printable strict ByteString (", build (BS.length bs), ")"]
```
endgame, line 45:
```suggestion
    _ -> mconcat ["non-printable strict ByteString (", build (BS.length bs), " bytes)"]
```

lib/amazonka-core/src/Amazonka/Data/Log.hs
```diff
@@ -34,10 +35,14 @@ instance ToLog ByteStringBuilder where
   build = id
 
 instance ToLog ByteStringLazy where
-  build = Build.lazyByteString
+  build lbs = case LText.decodeUtf8' lbs of
+    Right lt | LText.all Char.isPrint lt -> Build.lazyByteString lbs
+    _ -> mconcat ["non-printable lazy ByteString (", build (LBS.length lbs), ")"]
```
endgame, line 40:
```suggestion
    _ -> mconcat ["non-printable lazy ByteString (", build (LBS.length lbs), " bytes)"]
```

endgame (commented):
Thank you for this, and sorry for the delay.

Issue #872, opened 2023-01-05
Track botocore order of shape fields in generator


endgame:
In #862 the generator was changed to emit request/response fields in sorted order. We could switch the generator from `aeson` to [`waargonaut`](https://hackage.haskell.org/package/waargonaut), which uses a JSON AST that tracks such things.

endgame:
The new parser in `aeson >= 2.1.2.0`, [`Data.Aeson.Decoding.Tokens`](https://hackage.haskell.org/package/aeson-2.1.2.0/docs/Data-Aeson-Decoding-Tokens.html) is probably easier to use than `waargonaut` will be, as its `hw-simd-json` dependency doesn't build right now.

PR #971, opened 2024-01-06
Botocore library

endgame:
As part of #872, this PR adds a library to parse botocore `service-2.json`, `waiters-2.json` and `paginators-1.json` files. They use a hand-written parser based on the new [`Data.Aeson.Decoding.Tokens`](https://hackage.haskell.org/package/aeson-2.1.2.0/docs/Data-Aeson-Decoding-Tokens.html) provided by recent versions of `aeson`.

Because the generator currently smooshes the aeson `Value`s of these three together before attempting parsing, this PR also includes separate parsers for `Service`, `Waiters`, and `Paginators`, so that we can load each type of file individually and then marshal them into the types currently used by `gen`.

Future work:

* Applying annexes
* Plumbing this into `gen`


endgame:
After discussions with Brendan and also internally, this code has been published at https://github.com/bellroy/hs-botocore .

This prevents the copyright situation of files in this repo from becoming too complicated, and ensures that Bellroy gets appropriate credit for the work it paid for in developing this library. It also means that should the JSON parser become interesting enough for further development, it's clearly available under BSD3.

Brendan has been invited as a maintainer to ensure that he still has control of his project.

Issue #969, opened 2023-12-27
Compilation failures with GHC 9.8


ysangkok:
Amazonka packages were thrown out of Stackage Nightly, there must have been some kinda of incompatibility with the compiler or newer libraries: [see build-constraints.yaml](https://github.com/commercialhaskell/stackage/commit/9d04eeeb3e408db255ec31a65beb7a1dc26b7f47#diff-9949098b9af80e5fcef9a718fdda416f3fa8272ca8add4d67ae65fe892990939R1974-R1998)

ysangkok:
Error message is available:

* https://github.com/commercialhaskell/stackage/issues/7261

endgame:
https://gitlab.haskell.org/ghc/ghc/-/issues/24317

ysangkok:
Fixed by https://github.com/brendanhay/amazonka/commit/c604b8f282785f41977f234d33e2bae10773ece2

ysangkok:
@endgame Would it be possible to fix this just by amending `default-extensions`? Then it wouldn't be necessary to reupload everything, a revision would do.

endgame:
No, I tried to preview a change like that, and Hackage only allows revisions which change dependency information.

PR #972, opened 2024-01-09
GHC 9.8

endgame:
Tweak the Nix and the ede templates so that service definitions compile with GHC 9.8.

Regenerated services to follow in a separate PR.

@brendanhay We will need to add upper bounds to the existing 2.0 service definitions because they don't build with GHC 9.8. Do you have any tooling to help with this? Given the stuff already on `main`, this deserves a major release, but since there's so many packages to upload, it's probably worth batching it with a couple of additional fixes. Also, can you please add me to the maintainer group for the uploaded packages?

See also: https://gitlab.haskell.org/ghc/ghc/-/issues/24317
Fixes: #969

### DIFF
--- a/configs/templates/types.ede
+++ b/configs/templates/types.ede
@@ -1,7 +1,7 @@
-{-# LANGUAGE DisambiguateRecordFields #-}
-{-# LANGUAGE NoImplicitPrelude        #-}
-{-# LANGUAGE OverloadedStrings        #-}
-{-# LANGUAGE StrictData               #-}
+{-# LANGUAGE DuplicateRecordFields #-}
+{-# LANGUAGE NoImplicitPrelude     #-}
+{-# LANGUAGE OverloadedStrings     #-}
+{-# LANGUAGE StrictData            #-}
 
 {-# OPTIONS_GHC -fno-warn-unused-matches #-}
 
--- a/flake.lock
+++ b/flake.lock
@@ -38,11 +38,11 @@
         "systems": "systems"
       },
       "locked": {
-        "lastModified": 1687709756,
-        "narHash": "sha256-Y5wKlQSkgEK2weWdOu4J3riRd+kV/VCgHsqLNTTWQ/0=",
+        "lastModified": 1701680307,
+        "narHash": "sha256-kAuep2h5ajznlPMD9rnQyffWG8EM/C73lejGofXvdM8=",
         "owner": "numtide",
         "repo": "flake-utils",
-        "rev": "dbabf0ca0c0c4bce6ea5eaf65af5cb694d2082c7",
+        "rev": "4022d587cbbfd70fe950c1e2083a02621806a725",
         "type": "github"
       },
       "original": {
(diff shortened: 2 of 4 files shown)


brendanhay:
> We will need to add upper bounds to the existing 2.0 service definitions

You mean the existing 2.0 libraries already uploaded to Hackage? No tooling for that unfortunately, everything died in the Bazel fire.

brendanhay (approved):

endgame:
> > We will need to add upper bounds to the existing 2.0 service definitions
> 
> You mean the existing 2.0 libraries already uploaded to Hackage? No tooling for that unfortunately, everything died in the Bazel fire.

Yes, and potentially pre-2.0 service definitions also, because they won't compile with new GHC for the same reason. I will see if I can chase down some information about Hackage's APIs.

> The extension itself could go also into each service library's cabal file under a if/ghc conditional to ensure forwards/backwards compat?

I'm not sure what you mean by this. If we declare `{-# LANGUAGE DuplicateRecordFields #-}` in the module for each product type as well as for `Types.hs`, we get code which works with both GHC >=9.8 as well as <9.8.

brendanhay:
> I'm not sure what you mean by this. If we declare `{-# LANGUAGE DuplicateRecordFields #-}` in the module for each product type as well as for `Types.hs`, we get code which works with both GHC >=9.8 as well as <9.8.

I misunderstood until I read the linked gitlab issue.

Bodigrim:
> We will need to add upper bounds to the existing 2.0 service definitions because they don't build with GHC 9.8. Do you have any tooling to help with this?

One can use [`hackage-cli`](https://github.com/hackage-trustees/hackage-cli) to make batch revisions.

ysangkok:
@endgame What is needed to progress on this issue?

endgame:
The mass-revision is now done. Sorry for the delay..

ysangkok:
@endgame I see that this is part of the 2.1 milestone, but there are many tickets left in that milestone. Stackage is planning to do [LTS-23 very soon](https://github.com/commercialhaskell/stackage/issues/7462#issuecomment-2188161232) and it would be a shame if Amazonka wasn't included only because a release hadn't been made...

endgame:
How much time do we have? It would be nice to do an RC round before we drop 300+ packages onto Stackage, and Amazonka itself has a bunch of checks we need to run before each release (see the issue template).

Another wrinkle: I don't actually have the Hackage rights to upload new versions. @brendanhay are you around and do you have enough time to do things only you can do, if we choose to cut a release?

ysangkok:
I don't think the exact date has been chosen yet, but since the last LTS is from around new year's, the next one should theoretically go out next month unless GHC 9.6.6 is somehow delayed. But I don't really know for sure.

ysangkok:
> How much time do we have?

I think Stackage was planning to move onto GHC 9.10 once a good GHC 9.8 is out. Since 9.8.4 was just released a few days ago, and it seems Julian (GHCup maintainer) is reasonably happy with it, he added it to GHCup. That's something that could speak in favour of releasing a new Stackage LTS.

That being said, GHC 9.10.1 seems to have many issues and I don't know the schedule for 9.10.2, which seems to [have slipped quite a bit from July 2024](https://discourse.haskell.org/t/priorities-for-upcoming-ghc-releases/9605/5?u=janus). That delay could affect the new Stackage LTS release, since, as far as I understand, the Stackage maintainers prefer to wait with releasing the new LTS (with GHC 9.8) until the next Nightly (with GHC 9.10) would also work reasonably well.

PR #977, opened 2024-02-22
Remove expect from normalised headers

### LINKED ISSUE #975: R2 Incompatibility
I am unable to send a PutObject to R2, Cloudflare's object storage. I get a `SignatureDoesNotMatch` error.

I eventually figured out a workaround: remove `expect` from the list of normalised headers.

While I consider this a bug in R2, it looks like the aws utility doesn't include `expect` in its signature, either. I don't know the security implications for removing it, so I'm not gonna open a PR yet. Any thoughts?

This is the workaround for 2.0:

```patch
commit 5f31f5b53f8daa41d1ea54fdd492e8ff2a2c4a65
Author: Bryan Richter <b@chreekat.net>
Date:   Mon Jan 15 10:48:35 2024 +0200

    Remove expect from normalised headers
    
    It appears this header is incompatible with R2, which is probably a bug
    in R2.

diff --git a/lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs b/lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs
index f1c22d81b8..faa8cc1fcb 100644
--- a/lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs
+++ b/lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs
@@ -308,4 +308,5 @@ normaliseHeaders =
     . Map.toAscList
     . Map.delete "authorization"
     . Map.delete "content-length"
+    . Map.delete "expect"
     . Map.fromListWith const
```

And similarly for 1.6.1:

```patch
commit 62295f17a728ca083ca4ba720559d9a5173bbb4f
Author: Bryan Richter <b@chreekat.net>
Date:   Mon Jan 15 10:48:35 2024 +0200

    Remove expect from normalised headers
    
    It appears this header is incompatible with R2, which is probably a bug
    in R2.

diff --git a/core/src/Network/AWS/Sign/V4/Base.hs b/core/src/Network/AWS/Sign/V4/Base.hs
index 8e0714444b..47d5758252 100644
--- a/core/src/Network/AWS/Sign/V4/Base.hs
+++ b/core/src/Network/AWS/Sign/V4/Base.hs
@@ -266,4 +266,5 @@ normaliseHeaders = Tag
     . map    (first CI.foldedCase)
     . nubBy  ((==)    `on` fst)
     . sortBy (compare `on` fst)
+    . filter ((/= "expect") . fst)
     . filter ((/= "authorization") . fst)
```

chreekat:
It appears this header is incompatible with R2, which is probably a bug in R2. But nothing stops us from just dropping it.

Fixes #975

### DIFF
--- a/lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs
+++ b/lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs
@@ -301,6 +301,14 @@ canonicalHeaders = Tag . BSL.toStrict . BSB.toLazyByteString . Foldable.foldMap 
 signedHeaders :: NormalisedHeaders -> SignedHeaders
 signedHeaders = Tag . BS8.intercalate ";" . map fst . untag
 
+-- | Besides normalising capitalisation, this function removes certain headers
+-- that shouldn't be signed. AWS states that only @host@ and @x-amz-*@ headers
+-- are required[1]. To support R2, @expect@ must be removed[2]. botocore[3]
+-- removes @user-agent@ and @x-amzn-trace-id@, so we do too.
+--
+-- [1]: https://docs.aws.amazon.com/IAM/latest/UserGuide/create-signed-request.html
+-- [2]: https://github.com/brendanhay/amazonka/issues/975
+-- [3]: https://github.com/boto/botocore/blob/d5b2e4ab4bc4ad84f8e0e568e70ddc8ab7f094a8/botocore/auth.py#L61-L65
 normaliseHeaders :: [Header] -> NormalisedHeaders
 normaliseHeaders =
   Tag
@@ -308,4 +316,7 @@ normaliseHeaders =
     . Map.toAscList
     . Map.delete "authorization"
     . Map.delete "content-length"
+    . Map.delete "expect"
+    . Map.delete "user-agent"
+    . Map.delete "x-amzn-trace-id"
     . Map.fromListWith const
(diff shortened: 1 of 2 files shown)


lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs
```diff
@@ -308,4 +308,5 @@ normaliseHeaders =
     . Map.toAscList
     . Map.delete "authorization"
     . Map.delete "content-length"
+    . Map.delete "expect"
```
endgame, line 311: Let's remove all of the headers blacklisted by botocore, so please remove the `user-agent` and `x-amzn-trace-id` headers too.

There are so many places where we have to do things just because an SDK does them that way, or someone once spoke to support, or a service doesn't quite behave to spec ([example](https://github.com/brendanhay/amazonka/blob/9c3594073a7787e27f13fa14c936f0fa5cb9f0e8/lib/amazonka-core/src/Amazonka/Sign/V4/Chunked.hs#L36-L42)). Can you please add a comment that we're doing this for R2's sake, and also link to https://github.com/boto/botocore/blob/d5b2e4ab4bc4ad84f8e0e568e70ddc8ab7f094a8/botocore/auth.py#L61-L65 ?

endgame (changes requested):
One small change. Can you please also add an entry to `lib/amazonka/CHANGELOG.md` as it's the changelog for the whole project?

lib/amazonka-core/src/Amazonka/Sign/V4/Base.hs
chreekat: Yeah I was lying in bed last night thinking I needed to have a much better comment in the code for this "simple" commit. :P Coming right up!

chreekat: Done

endgame (approved):

Issue #981, opened 2024-04-04
Throws CredentialChainExhausted unless `~/.aws/credentials` exists


pbrisbin:
It seems something about the credential chain loading fails without the presence of `~/.aws/credentials`. The file is not needed in all setups (e.g. SSO-only) and so may not exist. Other SDKs (tested with AWS CLIv2) seem to still work, but Amazonka does not.

Reproduction:

1. Have a fresh machine with no config _besides_ SSO auth
2. Run `aws --profile x sso login` and log in
3. Observe something like `AWS_PROFILE=x aws s3 ls` works
4. Run an Amazonka-based program `AWS_PROFILE=x ./something`
5. Observe `CredentialChainExhausted`
6. Run `touch ~/.aws/credentials`
7. Observe `AWS_PROFILE=x ./something` now works

chris-martin:
This issue was observed in Amazonka `2.0`.

pbrisbin:
I've done some investigation, but as far as I can tell the code really shouldn't be able to exhibit this bug, which is fun.

`discover` is the only place that throws `CredentialChainExhausted`. It ostensibly throws it if it gets to the end of the chain without anything working, but it also throws it if it gets to the last link in the chain, which is EC2 metadata, and is not on EC2. I find that surprising, but shouldn't be harmful.

The second element in the chain is `fromFileEnv`, this is the happy path we expect since it eventually gets to the SSO code. This function just reads some ENV and calls `fromFilePath`, which has this:

```hs
  -- If we fail to read the credentials file, assume it's empty and
  -- move on. It is valid to configure only a config file if you plan
  -- to assume a role using any of the assume role methods.
  credentialsIni <-
    Exception.catchJust
      (\(_ :: AuthError) -> Just mempty)
      (loadIniFile credentialsFile)
      pure
```

Which (based on `loadIniFile`) should absolutely, as far as I can tell, treat a missing `credentials` file identically to an empty one (`catchJust` -> `mempty`). So I can't see how a `touch` of this file could change any behavior in this function. And besides merging `credentialsIni` into `configIni` before proceed, nothing else happens with this value :thinking: 

I will attempt to write a smaller reproduction so I can test against latest `main` tomorrow.

endgame:
> which is EC2 metadata, and is not on EC2. I find that surprising, but shouldn't be harmful.

[...]

PR #983, opened 2024-04-09
Throw dedicated error for attempting EC2 auth not on EC2

pbrisbin:
As discussed in #981:

Each link in the discover chain should throw an `AuthError` to trigger
moving onto the next link in the chain. If all links fail, then a
`CredentialChainExhausted` error is thrown.

In the case of the final link, attempting an EC2 instance profile, the
code was throwing `CredentialChainExhausted` if run not on EC2. Throwing
when not on EC2 is important (to avoid an unnecessary and slow network
request) but throwing this particular exception is surprising and only
functions because this is the last link in said chain. If any other
authentication mechanisms were added after it, they would never be
reached.

To address this, a new constructor was added to `AuthError` to use
instead. The name is very much up for debate. This is also the first
case of a nullary constructor besides `CredentialChainExhausted`, so I
wasn't sure what to do in `AsAuthError` -- please advise.

The original discussion mentioned introduce two new errors:

> One to say "we don't think we're on a network with an IMDS" and one
> for "IMDS fetch failed"?

I did think about this and found that the second case is already
throwing a specific error (`RetrievalError`), so I only changed the
first case.

### DIFF
--- a/lib/amazonka/src/Amazonka/Auth.hs
+++ b/lib/amazonka/src/Amazonka/Auth.hs
@@ -115,7 +115,7 @@ discover =
       fromContainerEnv,
       \env -> do
         onEC2 <- isEC2 $ manager env
-        unless onEC2 $ throwM CredentialChainExhausted
+        unless onEC2 $ throwM NotOnEC2Instance
         fromDefaultInstanceProfile env
     ]
 
--- a/lib/amazonka/src/Amazonka/Auth/Exception.hs
+++ b/lib/amazonka/src/Amazonka/Auth/Exception.hs
@@ -21,6 +21,7 @@ data AuthError
   | MissingFileError FilePath
   | InvalidFileError Text
   | InvalidIAMError Text
+  | NotOnEC2Instance
   | CredentialChainExhausted
   deriving stock (Show, Generic)
 
@@ -33,6 +34,7 @@ instance ToLog AuthError where
     MissingFileError f -> "[MissingFileError] { path = " <> build f <> "}"
     InvalidFileError e -> "[InvalidFileError] { message = " <> build e <> "}"
     InvalidIAMError e -> "[InvalidIAMError]  { message = " <> build e <> "}"
+    NotOnEC2Instance -> "[NotOnEC2Instance]"
     CredentialChainExhausted -> "[CredentialChainExhausted]"
 
 class AsAuthError a where
@@ -57,11 +59,15 @@ class AsAuthError a where
   -- | The specified IAM profile could not be found or deserialised.
   _InvalidIAMError :: Prism' a Text
 
+  -- | Using an EC2 Instance profile was attempted not on an EC2 instance.
+  _NotOnEC2Instance :: Prism' a AuthError
+
   _RetrievalError = _AuthError . _RetrievalError
   _MissingEnvError = _AuthError . _MissingEnvError
   _MissingFileError = _AuthError . _MissingFileError
   _InvalidFileError = _AuthError . _InvalidFileError
   _InvalidIAMError = _AuthError . _InvalidIAMError
+  _NotOnEC2Instance = _AuthError . _NotOnEC2Instance
 
 instance AsAuthError SomeException where
   _AuthError = exception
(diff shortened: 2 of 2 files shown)


pbrisbin:
It seems some things have shifted under this PR. I don't know if it matters, but just FYI:

> Throwing when not on EC2 is important (to avoid an unnecessary and slow network request)

Whatever was meant to avoid a "slow network request" may no longer be happening. See #1018.

> If any other authentication mechanisms were added after it, they would never be reached

The PR fix for the above issue re-orders things. I would expect the above to mean that breaks things (without including this PR too), but it seems it doesn't somehow, so there may be further misunderstanding.

endgame:
Sorry for the delay, and thank you for this PR. The cause of the slowness reported in #1018 seems to be that it recurses inside the exception handler, so the credential chain runs back-to-front. #1029 should fix that.

Aside: I don't think your prism is correct. Prisms for nullary constructors should return `()` on success, so if we were going with this design, I think we'd want `_NotOnEC2Instance :: Prism' a ()`. But I don't think it's the right design; see below.

I don't think this PR is exactly what we want because it throws to trigger control flow in `fromCredentialChain`, not because the authentication method itself is failing. I don't even know why we do a DNS lookup to decide whether we are on EC2 or not. I assume it's to avoid us making random network requests and annoying non-IMDS hosts that might be listening on `http://169.254.169.254`, but `botocore` doesn't do it and so we should probably stop:

https://github.com/boto/botocore/blob/d5d90e9666c66080f006826a40bd7bf2cdb7c14f/botocore/utils.py#L93

Our DNS-based code is actually subtly wrong:

https://github.com/brendanhay/amazonka/blob/f3a7fca02fdbb832cc348e991983b1465225d50c/lib/amazonka/src/Amazonka/EC2/Metadata.hs#L602-L644

Notice that we only use the `http://instance-data` DNS name in `isEC2`, and in all of our actual metadata/userdata lookups, we use the IP address.

`botocore` also supports an IPv6 IMDS endpoint at `http://[fd00:ec2::254]/`, a configuration key `"ec2_metadata_service_endpoint_mode"` = `"ipv6"` to enable it, and a configuration key `"ec2_metadata_service_endpoint"` to provide a custom URL. Ideally, we should do the same.

IMHO, here's what I think we should do:

* Unconditionally request from the IMDS as the final step in the `discover` credential chain. `fromDefaultInstanceProfile` already throws an `AuthError` if it fails, so we should just let that happen.
* Clean up haddock for `discover` to stop talking about DNS-based checks for the IMDS.
* Remove (or at least deprecate) `isEC2`.

Let me know if you're willing to take this on.

pbrisbin:
Thanks for the response.

> * Unconditionally request from the IMDS as the final step in the discover credential chain

IIUC, now that you've fixed the chain order so this one is last, this really just amounts to removing `isEC2`, right?

> * Clean up haddock for discover to stop talking about DNS-based checks for the IMDS.

Seems easy to do as part of the above.


> * Remove (or at least deprecate) isEC2.

Once the above is done it would be unused. But if it's technically public API then removal should signal a major version bump, so I can see an argument for deprecating. But I also see an argument for not actually considering it public API, even if it technically is, so I leave that decision up to you.

If I have all that correct, I think I can put up a PR pretty easily. Either way, this one should be closed right?

endgame:
> IIUC, now that you've fixed the chain order so this one is last, this really just amounts to removing `isEC2`, right?

At least removing the use of it from `discover`, yes.

> doing the one that requires a network request last makes sense regardless.

Yes, and it also keeps us consistent with official AWS SDKs.

> major version bump

We will be doing a major version bump anyway, because `2.1` will be the first two elements of its PVP version. I can't imagine people wanting to use it, but let's deprecate it because it can return false negatives and false positives, warn that it will be removed in 2.2, and invite people to comment on #1033 if they think they need it.

I think you're right, we should eventually close this PR. Let's do it after your new one goes through so I don't forget for some reason.

pbrisbin:
I opened https://github.com/brendanhay/amazonka/pull/1034.

I did it slightly quickly, so do comment if there's any book-keeping stuff (e.g. commit messages) you'd like to see improved.

PR #993, opened 2024-06-09
Make (Typeable a, Typeable (AWSResponse a)) superclasses of AWSRequest a

dalpd:


endgame:
Thanks for this. I did a bit of cleanup, wrote the changelog and some more docs, and merged it in so we can start on new stuff tomorrow.

PR #996, opened 2024-06-14
Remove chonky Lens and replace with microlens for smaller footprint.

mankyKitty:
This PR removes the dependency on the mega `lens` package and creates optics and friends using the `microlens` set of packages. The use of Prisms does necessitate the inclusion of `profunctor` but that is still a smaller price than all of `lens`.

The work is somewhat incomplete as there are a few incompatibilities to iron out.


endgame:
Thank you for this, I've got it a lot closer. There's just a couple more things to fix and then hopefully I can hit merge - I should get to them in the next few days, fingers crossed.

PR #1014, opened 2025-01-13
updated constraints on addAWSRequestHook to include Typeable

Montmorency:
Compilation issue with ghc.9.8.3 not able to resolve Typeable constraint on addAWSRequestHook.


endgame:
This doesn't seem right to me: `Typeable` is a superclass of `AWSRequest`:

https://github.com/Montmorency/amazonka/blob/8c33c4ddef7430b4b052dec03e2bfc65c248a4e4/lib/amazonka-core/src/Amazonka/Types.hs#L656

Is there a GHC bug or something that's causing the `Typeable` constraint not to be solved?

endgame:
It's not clear what trouble you're having but see if #1015 solves your problem - there were some redundant `Typeable` constraints which it removes (though they should be redundant, as it is now a superclass of `AWSRequest`.

Montmorency:
Solved by [#1015 ]

PR #1022, opened 2025-01-15
Gen cabal template

### LINKED ISSUE #888: Deriving clauses of `Ptr` are not calculated correctly

endgame:
I was talking with a colleague about the generator today. I think one of the big reasons the generator is hard to untangle is that many of its types are pulling double or triple duty: first, they represent the JSON files are they are read in from `botocore` (and merged with our annexes); second, they represent the service as the generator understands them (as a service with endpoints and operations that send/receive shapes in formats); third, they represent (via their `ToJSON` instances) the input to the various `.ede` templates used to generate the output files.

I hope to eventually deal with #888 by starting at the output and refactoring backwards, until the types currently in `amazonka-gen` aren't participating in output any more. This PR is the first step in that plan. It introduces the `Gen.Output` namespace and a `Gen.Output.Template.Template` type, representing a template that tracks the type of the record it needs to consume. `Gen.Output.Template.CabalFile` then represents the fields needed by `cabal.ede`, and its only job is to get them into the `HashMap Text Aeson.Value` that EDE requires.

This then revealed some easy opportunities for simplification, so I did them too.

The hope is that once the current types are no longer overloaded, we can simplify them by removing type variables, instances, etc., and eventually replace them (e.g., by replacing the cofree comonadic shape functor with a type that only refers to shapes by name; when everything looks at it from that perspective, #888 becomes impossible).

The current goal is to have three phases — "input", "model", and "output":

* "Input" should read the source IDL as accurately as possible. Currently this is `botocore` and we have https://github/bellroy/hs-botocore to eventually swap in here;
* "Model" will be our playground to describe AWS services. I expect this to be aware of request and response shapes, but also Haskell-specific things like what types to use in a representation, which deriving clauses we need to generate, and which module names to import (and whether to import them `{-# SOURCE #-}` or not); and
* "Output" will be where we set up the output directory tree, feed data to EDE templates, etc. Because each template will eventually have its own record to track its inputs, we may want to include a thin layer of type-safety here. Future work might have an argument record contain a `[ModuleName]` naming the `exposed-modules` instead of a `[Text]`, but the main point is to keep this as simple as possible.

This PR does not change the output of the generator at all; I re-ran it over the repo and got a clean diff.

### DIFF
--- a/gen/src/Gen/Output/Template/CabalFile.hs
+++ b/gen/src/Gen/Output/Template/CabalFile.hs
@@ -9,7 +9,7 @@ import Data.Aeson (toJSON)
 import qualified Data.Aeson as Aeson
 import qualified Data.HashMap.Strict as HashMap
 import Gen.Prelude
-import Gen.Types.Config (Library)
+import Gen.Types.Config (Library, exposedModules, otherModules)
 import qualified Gen.Types.Config as Config
 import Gen.Types.NS (unNS)
 import Gen.Types.Service (metadata, service)
@@ -22,8 +22,7 @@ data CabalFile = CabalFile
     serviceAbbrev :: Text,
     serviceFullName :: Text,
     apiVersion :: Text,
-    exposedModules :: [Text],
-    otherModules :: [Text],
+    modules :: [Text],
     extraDependencies :: [Text]
   }
   deriving (Generic)
@@ -36,8 +35,7 @@ arguments CabalFile {..} =
       ("serviceAbbrev", toJSON serviceAbbrev),
       ("serviceFullName", toJSON serviceFullName),
       ("apiVersion", toJSON apiVersion),
-      ("exposedModules", toJSON exposedModules),
-      ("otherModules", toJSON otherModules),
+      ("modules", toJSON modules),
       ("extraDependencies", toJSON extraDependencies)
     ]
 
@@ -49,7 +47,6 @@ fromLibrary lib =
       serviceAbbrev = lib ^. service . metadata . Service.serviceAbbrev,
       serviceFullName = lib ^. service . metadata . Service.serviceFullName,
       apiVersion = lib ^. service . metadata . Service.apiVersion,
-      exposedModules = unNS <$> lib ^. Config.exposedModules,
-      otherModules = unNS <$> lib ^. Config.otherModules,
+      modules = unNS <$> lib ^. exposedModules <> lib ^. otherModules,
       extraDependencies = lib ^. Config.config . Config.extraDependencies
     }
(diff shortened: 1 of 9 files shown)


gen/src/Gen/Output/Template/CabalFile.hs
```diff
@@ -1,16 +1,28 @@
...
 import qualified Data.HashMap.Strict as HashMap
 import Gen.Prelude
+import Gen.Types.Config (Library)
+import qualified Gen.Types.Config as Config
+import Gen.Types.NS (unNS)
+import Gen.Types.Service (metadata, service)
+import qualified Gen.Types.Service as Service
 
+-- | Arguments for the @cabal.ede@ template.
 data CabalFile = CabalFile
```
Tristano8, line 19: Do you think the version of cabal that is described in the generated file should be a property of this datatype or of the template itself?

Tristano8 (approved):

gen/src/Gen/Output/Template/CabalFile.hs
endgame: Nah it's fine. It's just a file format specifier and if we start depending on new features we'll be changing the template structure anyway. No point putting `"2.2"` in our code forever.

PR #1029, opened 2025-03-10
amazonka: Catch auth exceptions in the correct order

### LINKED ISSUE #1018: newEnv discover adds excessive latency in latest code
Hi, thanks for this great project.
Relatively new to Haskell, but with working experience with AWS SDK for js (node).
I just started exploring amazonka to develop some command-line utilities with this sdk.
All good with the latest release, but noticed it was a bit outdated, so I tried switching to the latest commit in main.
At this point I started to notice some serious slowdown, 30 secs latency just to execute this silly snip:

```
main = do
  env <- newEnv discover
```

I tracked the origin of the "problem" down to the merge of the arianvp/fix-aws-sso branch, https://github.com/brendanhay/amazonka/commit/5461403cd5412cc0aedd64fe1d27ebf50cef66b3 .
From this point on 30 secs extra latency gets added to every program trying to set the env.
Went to profiling the whole thing to understand why, with no luck, all being assigned to SYSTEM, so RTS doing its business:

	   try +RTS -pa -hm -RTS

	total time  =       29.77 secs   (29765 ticks @ 1000 us, 1 processor)
	total alloc =  12,362,808 bytes  (excludes profiling overheads)

```
COST CENTRE                      MODULE                                       SRC                                                                    %time %alloc  ticks     bytes

SYSTEM                                SYSTEM                                        <built-in>                                                              99.5    0.3  29608     34760
readCertificates                    Data.X509.CertificateStore           Data/X509/CertificateStore.hs:(93,1)-(97,33)           0.2   58.2     60   7195424
GC                                         GC                                              <built-in>                                                                 0.1    0.0     39      4560
```
versus the release amazonka:

	   try +RTS -pa -hm -RTS

	total time  =        0.01 secs   (12 ticks @ 1000 us, 1 processor)
	total alloc =  11,720,176 bytes  (excludes profiling overheads)
```
COST CENTRE              MODULE                              SRC                                                                %time %alloc  ticks     bytes

GC                                GC                                       <built-in>                                                            33.3    0.0      4      4536
readCertificates            Data.X509.CertificateStore   Data/X509/CertificateStore.hs:(93,1)-(97,33)       25.0   61.4      3   7195424
envHelper                    Network.HTTP.Proxy             Network/HTTP/Proxy.hs:(338,1)-(380,87)             16.7   16.9      2   1985096
readCertificateStore     Data.X509.CertificateStore    Data/X509/CertificateStore.hs:(75,1)-(87,50)       16.7   17.9      2   2093536
```
Looking at the new code it seems the origin of the extra latency is the discover function lib/amazonka/src/Amazonka/Auth.hs:117, in which a DNS lookup is always performed to check if the process is running inside an EC2.

The problem is, this adds up 30 extra secs always, which I think it's not acceptable in latency sensitive apps, like CLIs and lambda.

My question is: are we ok with that ? Maybe someone already noticed that and there is some workaround I'm not aware of?

endgame:
Previously, we were recursing through the provider list inside the exception handler, which resulted in authentication methods being tried back-to-front.

@matteoeghirotta @pbrisbin please confirm whether this fixes your issue.

Closes #1023 
Fixes #1018

### DIFF
--- a/lib/amazonka/CHANGELOG.md
+++ b/lib/amazonka/CHANGELOG.md
@@ -32,6 +32,8 @@
 
 ### Fixed
 
+- `amazonka`: Attempt to load credentials in the correct order. In particular, this means that the IMDS is queried last and is once again consistent with other SDKs.
+[\#1029](https://github.com/brendanhay/amazonka/pull/1029)
 - `.cabal` files name the source repository with a `https://` URL, not a `git://` one (thanks @marinelli)
 [\#1026](https://github.com/brendanhay/amazonka/pull/1026)
 - `amazonka-*`: GHC 9.8 support
--- a/lib/amazonka/src/Amazonka/Auth.hs
+++ b/lib/amazonka/src/Amazonka/Auth.hs
@@ -66,8 +66,7 @@ import Amazonka.EC2.Metadata
 import Amazonka.Env (Env, Env' (..), EnvNoAuth)
 import Amazonka.Prelude
 import Amazonka.Types
-import Control.Exception (fromException)
-import Control.Monad.Catch (MonadCatch (..), catchJust, throwM)
+import Control.Monad.Catch (MonadCatch (..), throwM)
 
 -- | Attempt to fetch credentials in a way similar to the official AWS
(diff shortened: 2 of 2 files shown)


pbrisbin:
I can confirm the slow down is gone for me.

endgame:
Thank you.

Issue #1035, opened 2025-04-15
containers-0.8 compat


ysangkok:
it seems like amazonka works fine on GHC-9.12 with containers-0.8, however the bound on amazonka-core.cabal forbids this version. Would you accept a PR to bump this bound?

ysangkok:
They can, but it's not going to be very common since it is a boot package and I think Cabal prefers using the version that GHC likes. I will send a PR.

ysangkok:
I am sorry, I was confused, my issue description is incorrect. There are in fact *two* major versions of containers that amazonka-core doesn't allow. Only containers-0.7 works. Containers-0.8 isn't included by any GHC release yet, and other dependencies aren't compatible with it yet. So the PR I submitted only addresses the containers-0.7 compat.

endgame:
OK, thanks. I'll move this issue to 2.2 and we'll do 0.7 compat as part of 2.1.

ysangkok:
@endgame I think it would be better to also do containers-0.8 support in release 2.1, given that GHC 9.14.1 has containers-0.8, and has already been released.

endgame:
All right. I'll put it back in 2.1 for now, and we'll see how ready the library ecosystem is when we get closer to release.

PR #1036, opened 2025-04-15
Allow containers-0.7

ysangkok:
Tested using

```
cabal build lib:amazonka-core -c 'containers>=0.7' -w ghc-9.12.2
```

Even though containers-0.8 is out, it isn't supported by other dependencies yet. So let's start with just allowing containers-0.7. This means that this PR *doesn't* address #1035.


endgame:
Could you please add an entry to the main changelog under `lib/amazonka` and force-push this PR branch?

ysangkok:
@endgame Ok, done

endgame:
Thank you very much.

Issue #1027, opened 2025-02-13
New region: Mexico (`mx-central-1`)


endgame:
Add https://aws.amazon.com/blogs/aws/now-open-aws-mexico-central-region/

Also update any service bindings that have `Region -> Maybe Text` functions (usually returning an Account ID or Hosted Zone ID).

kahlil29:
@endgame Opened a draft PR #1028 
Not sure what the values for the Account ID & Hosted Zone ID would be. Could you help me find those for Mexico Central please?

endgame:
Each function that #1017 changed should have a link to the corresponding AWS documentation providing the relevant details in a table. If this isn't the case, please let me know and I'll help you fix it.

Some functions won't have new account IDs for the regions, because new regions use service-linked roles for e.g. log delivery.

Please also try to keep the order of regions consistent with the tables that drive them.

Good luck, and let me know how you go.

ysangkok:
@kahlil29 On Redshift/CloudTrail, it's "058264411980". Note that [the table](https://docs.amazonaws.cn/en_us/redshift/latest/mgmt/logging-with-cloudtrail.html#cloudtrail-rs-acct-ids) is scrollable, and on my device, it isn't shown in the initial rendering.

PR #1038, opened 2025-05-20
Adding Mexico Central Region

kushagarr:
Closes https://github.com/brendanhay/amazonka/issues/1027

### DIFF
--- a/lib/amazonka-core/src/Amazonka/Types.hs
+++ b/lib/amazonka-core/src/Amazonka/Types.hs
@@ -133,6 +133,7 @@ module Amazonka.Types
         Malaysia,
         Melbourne,
         Mumbai,
+        MexicoCentral,
         Osaka,
         Seoul,
         Singapore,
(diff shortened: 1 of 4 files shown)


endgame (changes requested):
Thank you for doing this, and sorry for the delay. We have previously had somewhat idiosyncratic naming, where we named the Canada (Central) `ca-central-1` region `Montreal` and named the Canada (West) `ca-west-1` region `Calgary`. I've been mulling over whether we want to do this for Mexico, but I think the answer is "no".

That said, I'd like you to please move the definitions so they remain in order with the tables in AWS documentation.

Please also add an entry to `lib/amazonka/CHANGELOG.md`, which is the primary changelog for the project and give yourself credit.

kushagarr:
Done, let me know if I have missed something.

lib/amazonka-core/src/Amazonka/Types.hs
```diff
@@ -133,6 +133,7 @@ module Amazonka.Types
         Malaysia,
         Melbourne,
         Mumbai,
+        MexicoCentral,
```
endgame, line 136: Move this under TelAviv too please.

endgame (commented):
Sorry, missed the export list in my previous review.

ysangkok:
thank you so much @kushagarr ! It's gonna be fun to try

Issue #989, opened 2024-05-17
Error handling in periodic credential update


ynishinaka:
The [`fromWebIdentity`](https://github.com/brendanhay/amazonka/blob/ea8679437dd93d2dea74521285d6260d60cea966/lib/amazonka/src/Amazonka/Auth/STS.hs#L62) passes to [`fetchAuthInBackground`](https://github.com/brendanhay/amazonka/blob/ea8679437dd93d2dea74521285d6260d60cea966/lib/amazonka/src/Amazonka/Auth/Background.hs#L28-L33) the action that performs the `AssumeRoleWithWebIdentity` [1]. `AssumeRoleWithWebIdentity` may throw `ServiceError` (as exception [`Amazonka.Types.Error`](https://github.com/brendanhay/amazonka/blob/ea8679437dd93d2dea74521285d6260d60cea966/lib/amazonka-core/src/Amazonka/Types.hs#L282-L289)), not just `HTTPException`. However, `fetchAuthInBackground` catches and throws only `HTTPException` to the original thread by `throwTo`, leaving other exceptions to the default handler [2]. Thus, if a `ServiceError` is thrown, *the temporary credential update will be interrupted without the original thread being aware of it*.

So, I believe that `fetchAuthInBackground` should either re-throw other exceptions raised by the action to the original thread or have some error recovery method.

-----

[1]
https://github.com/brendanhay/amazonka/blob/ea8679437dd93d2dea74521285d6260d60cea966/lib/amazonka/src/Amazonka/Auth/STS.hs#L80-L94

[2] Because the `RetrievalError` constructor accepts an argument of type `HTTPException`, `e` must be of the same type:
https://github.com/brendanhay/amazonka/blob/ea8679437dd93d2dea74521285d6260d60cea966/lib/amazonka/src/Amazonka/Auth/Background.hs#L63-L65

PR #1039, opened 2025-05-31
Error handling in periodic credential update

kushagarr:
Closes https://github.com/brendanhay/amazonka/issues/989

### DIFF
--- a/lib/amazonka/amazonka.cabal
+++ b/lib/amazonka/amazonka.cabal
@@ -120,3 +120,19 @@ library
     , transformers          >=0.2
     , unordered-containers  ^>=0.2.14.0
     , uuid                  >=1.2.6     && <1.4
+
+test-suite tests
+  import:            base
+  type:              exitcode-stdio-1.0
+  hs-source-dirs:    test
+  main-is:           Main.hs
+  other-modules:
+    Test.Amazonka.Auth.Background
+  build-depends:
+    , amazonka
+    , amazonka-core
+    , tasty              >=0.10
+    , tasty-hunit        >=0.9
+    , time
+    , http-client
+    , http-types
--- a/lib/amazonka/src/Amazonka/Auth/Background.hs
+++ b/lib/amazonka/src/Amazonka/Auth/Background.hs
@@ -60,9 +60,12 @@ fetchAuthInBackground menv =
           then untilExpiry - fiveMinutes
           else untilExpiry `div` 2
 
-      env <- Exception.try ma
+      env <- Exception.try @SomeException ma
       case env of
-        Left e -> Exception.throwTo p (RetrievalError e)
+        Left e -> Exception.throwTo p $ case Exception.fromException e of
+          Just (RetrievalError e') -> RetrievalError e'
+          Just (AuthServiceError e') -> AuthServiceError e'
+          _ -> OtherAuthError e
         Right a -> do
           mr <- Weak.deRefWeak w
           case mr of
--- a/lib/amazonka/test/Main.hs
+++ b/lib/amazonka/test/Main.hs
@@ -1,0 +1,8 @@
+module Main (main) where
+
+import qualified Test.Amazonka.Auth.Background as AuthBackground
+import Test.Tasty (defaultMain, testGroup)
+import Prelude
+
+main :: IO ()
+main = defaultMain (testGroup "amazonka" [AuthBackground.tests])
--- a/lib/amazonka/CHANGELOG.md
+++ b/lib/amazonka/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ### Changed
 
+- `amazonka`: Improve error handling during periodic credential refresh in `fetchAuthInBackground`.
+  Exceptions thrown by the refresh action are now categorized and rethrown to the parent thread as
+  `RetrievalError`, `AuthServiceError`, or `OtherAuthError` (instead of just `RetrievalError` which was a bug). (thanks @kushagarr)
+[\#1039](https://github.com/brendanhay/amazonka/pull/1039)
 - The hooks interface is now much harder to misuse. [\#1042](https://github.com/brendanhay/amazonka/pull/1042)
 
   It was previously extremely easy to write hook-using functions that typechecked but did not ever run. The main change is to provide specialised hook-changing functions named for each field in the `Hooks` record, so that it is much easier to get the types correct.
--- a/lib/amazonka/src/Amazonka/Auth/Exception.hs
+++ b/lib/amazonka/src/Amazonka/Auth/Exception.hs
@@ -23,6 +23,8 @@ data AuthError
   | InvalidFileError Text
   | InvalidIAMError Text
   | CredentialChainExhausted
+  | AuthServiceError ServiceError
+  | OtherAuthError SomeException
   deriving stock (Show, Generic)
 
 instance Exception AuthError
@@ -35,6 +37,8 @@ instance ToLog AuthError where
     InvalidFileError e -> "[InvalidFileError] { message = " <> build e <> "}"
     InvalidIAMError e -> "[InvalidIAMError]  { message = " <> build e <> "}"
     CredentialChainExhausted -> "[CredentialChainExhausted]"
+    AuthServiceError e -> "[AuthServiceError] { serviceError = " <> build e <> "}"
+    OtherAuthError e -> "[OtherAuthError] { message = " <> build (show e) <> "}"
 
 class AsAuthError a where
   -- | A general authentication error.
--- a/lib/amazonka/test/Test/Amazonka/Auth/Background.hs
+++ b/lib/amazonka/test/Test/Amazonka/Auth/Background.hs
@@ -1,0 +1,113 @@
+{-# LANGUAGE NumericUnderscores #-}
+
+-- |
+-- Module      : Test.Amazonka.Auth.Background
+-- Copyright   : (c) 2013-2023 Brendan Hay
+-- License     : Mozilla Public License, v. 2.0.
+-- Maintainer  : Brendan Hay <brendan.g.hay+amazonka@gmail.com>
+-- Stability   : provisional
+-- Portability : non-portable (GHC extensions)
+module Test.Amazonka.Auth.Background (tests) where
+
+import Amazonka.Auth.Background (fetchAuthInBackground)
+import Amazonka.Auth.Exception (AuthError (..))
+import Amazonka.Core (Sensitive (..), Time (..))
+import Amazonka.Types
+  ( AccessKey (..),
+    AuthEnv (..),
+    SecretKey (..),
+    ServiceError (..),
+  )
+import Control.Concurrent (threadDelay)
+import Control.Exception (Exception, SomeException (..), throwIO, try)
+import Data.IORef (atomicModifyIORef', newIORef)
+import Data.Time.Clock (addUTCTime, getCurrentTime)
+import Network.HTTP.Client (HttpException (..))
+import Network.HTTP.Types.Status (status500)
+import Test.Tasty
+import Test.Tasty.HUnit
+import Prelude
+
+-- | How long we wait for the background refresh to run (microseconds).
+refreshWaitMicros :: Int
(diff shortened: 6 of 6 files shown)


endgame (commented):
Sorry for the delay on this. This change looks right to me, and I like that you've caught `SomeException` too, but I think we should write a test for it.

We can do this by providing an `IO AuthEnv` action that uses an `IORef` to remember state. On its first invocation, we return an `AuthEnv` with `expiration = Just 0` or similar, and on subsequent invocations we throw an exception. The test can just be a test of `fetchAuthInBackground` and need not construct a full `Amazonka.Env`.

Please also add an entry to `lib/amazonka/CHANGELOG.md`.

kushagarr:
@endgame  I have added a test suite in amazonka itself, let me know if that is not the right place for it. and also do look at the tests. and i don't know why there is a confict error showing up in CHANGELOG.md.

please have a look

lib/amazonka/amazonka.cabal
```diff
@@ -120,3 +120,22 @@ library
...
+  build-depends:
+    , base                  >=4.12 && <5
+    , amazonka              ^>=2.0
+    , amazonka-core         ^>=2.0
+    , tasty                 >=0.10
+    , tasty-hunit           >=0.9
+    , time                  >=1.9
+    , exceptions            ^>=0.10.4
+    , http-client           >=0.5
+    , http-types            >=0.12
```
endgame, line 141: If a dependency is listed in the `library` target, we do not need to repeat its bounds here.

lib/amazonka/src/Amazonka/Auth/Background.hs
```diff
@@ -60,9 +60,12 @@ fetchAuthInBackground menv =
...
           else untilExpiry `div` 2
 
-      env <- Exception.try ma
+      env <- Exception.try @SomeException ma
       case env of
-        Left e -> Exception.throwTo p (RetrievalError e)
+        Left e -> case Exception.fromException e of
+          Just (RetrievalError e') -> Exception.throwTo p (RetrievalError e')
+          Just (AuthServiceError e') -> Exception.throwTo p (AuthServiceError e')
+          _ -> Exception.throwTo p (UnknownAuthError e)
```
endgame, line 68: Factor out the `throwTo`:

```haskell
Left e -> Exception.throwTo p $ case Exception.fromException e of
  ...
```

lib/amazonka/test/Main.hs
```diff
@@ -0,0 +1,97 @@
+{-# LANGUAGE OverloadedStrings #-}
```
endgame, line 1: Please add the standard module header. Also, explicit extensions are not required because they'll be pulled in by `amazonka.cabal` file once you import its `base` stanza in the `tests` target.

lib/amazonka/test/Main.hs
endgame, line 1: Please push the tests themselves into a `Test.Amazonka.Auth.Background` module, following the pattern in `amazonka-core`.

lib/amazonka/test/Main.hs
```diff
@@ -0,0 +1,97 @@
...
+        , secretAccessKey = Sensitive "test-sk"
+        , sessionToken = Nothing :: Maybe (Sensitive SessionToken)
+        , expiration = Just expiry
+        }
+  pure $ do
+    -- Flip the boolean once and return the old value:
+    --   • On the first run, old value is True (and we set it to False),
+    --     so `first` is True, return env0.
+    --   • On later runs, old value is False, so `first` is False, throw `ex`.
+    first <- atomicModifyIORef' firstRef (\b -> (False, b))
```
endgame, line 94: `TupleSections` is enabled, so:
```suggestion
    first <- atomicModifyIORef' firstRef (False,)
```

lib/amazonka/test/Main.hs
```diff
@@ -0,0 +1,97 @@
...
+        { accessKeyId = "test-ak"
+        , secretAccessKey = Sensitive "test-sk"
+        , sessionToken = Nothing :: Maybe (Sensitive SessionToken)
+        , expiration = Just expiry
+        }
+  pure $ do
+    -- Flip the boolean once and return the old value:
+    --   • On the first run, old value is True (and we set it to False),
+    --     so `first` is True, return env0.
+    --   • On later runs, old value is False, so `first` is False, throw `ex`.
```
endgame, line 93: I think this can be deleted

lib/amazonka/test/Main.hs
endgame, line 94: Suggest rename to `firstCall` (`first` is a method of `class Bifunctor`).

lib/amazonka/test/Main.hs
```diff
@@ -0,0 +1,97 @@
...
+mkAuthEnvAction :: Exception e => e -> IO (IO AuthEnv)
+mkAuthEnvAction ex = do
+  firstRef <- newIORef True
+  now <- getCurrentTime
+  let expiry :: ISO8601
+      expiry = Time (addUTCTime 1 now)
+      env0 = AuthEnv
+        { accessKeyId = "test-ak"
+        , secretAccessKey = Sensitive "test-sk"
+        , sessionToken = Nothing :: Maybe (Sensitive SessionToken)
```
endgame, line 86: Type signature may be redundant.

lib/amazonka/test/Main.hs
```diff
@@ -0,0 +1,97 @@
...
+      res <- try @AuthError $ do
+                _ <- fetchAuthInBackground menv
+                threadDelay 700000
+      case res of
+        Left (RetrievalError _) -> pure ()
+        Left e -> assertFailure ("unexpected error: " ++ show e)
+        Right _ -> assertFailure "expected RetrievalError"
+  ]
+
+-- Helpers
```
endgame, line 70: Please don't leave these marks in your code. It's distracting and they easily become out of date.

lib/amazonka/test/Main.hs
```diff
@@ -0,0 +1,97 @@
...
+-- that behaves like a typical credential source used by fetchAuthInBackground:
+--   • On the first execution, it returns a valid AuthEnv with a near-term expiry
+--     (so the background refresher kicks in promptly).
+--   • On subsequent executions, it throws the provided exception `ex`.
+mkAuthEnvAction :: Exception e => e -> IO (IO AuthEnv)
+mkAuthEnvAction ex = do
+  firstRef <- newIORef True
+  now <- getCurrentTime
+  let expiry :: ISO8601
+      expiry = Time (addUTCTime 1 now)
```
endgame, line 82: Inline this into the assignment of `expiration = Just . Time $ addUTCTime 1 now`. Then the type signature will be redundant.

lib/amazonka/test/Main.hs
```diff
@@ -0,0 +1,97 @@
...
+
+-- Helpers
+
+-- mkAuthEnvAction creates a refresh action
+-- that behaves like a typical credential source used by fetchAuthInBackground:
+--   • On the first execution, it returns a valid AuthEnv with a near-term expiry
+--     (so the background refresher kicks in promptly).
+--   • On subsequent executions, it throws the provided exception `ex`.
+mkAuthEnvAction :: Exception e => e -> IO (IO AuthEnv)
+mkAuthEnvAction ex = do
```
endgame, line 78:
```suggestion
mkFailOnRefreshAuthAction :: Exception e => e -> IO (IO AuthEnv)
mkFailOnRefreshAuthAction ex = do
```

lib/amazonka/test/Main.hs
```diff
@@ -0,0 +1,97 @@
...
+        Right _ -> assertFailure "expected RetrievalError"
+  ]
+
+-- Helpers
+
+-- mkAuthEnvAction creates a refresh action
+-- that behaves like a typical credential source used by fetchAuthInBackground:
+--   • On the first execution, it returns a valid AuthEnv with a near-term expiry
+--     (so the background refresher kicks in promptly).
+--   • On subsequent executions, it throws the provided exception `ex`.
```
endgame, line 76: I think if you do the rename suggested below, this will become redundant.

lib/amazonka/test/Main.hs
```diff
@@ -0,0 +1,97 @@
...
+        _ <- fetchAuthInBackground menv
+        threadDelay 700000
+      case res of
+        Left (AuthServiceError s) -> case s of
+          ServiceError' { code = c } -> c @?= "InternalFailure"
+        Left e -> assertFailure ("unexpected error: " ++ show e)
+        Right _ -> assertFailure "expected AuthServiceError"
+  , testCase "RetrievalError is propagated" $ do
+      let retr = InvalidUrlException "url" "bad"
+      menv <- mkAuthEnvAction (RetrievalError retr)
```
endgame, line 60: Inline the definition of `retr`

lib/amazonka/test/Main.hs
```diff
@@ -0,0 +1,97 @@
...
+import           Control.Concurrent             (threadDelay)
+import           Control.Exception              (Exception, SomeException(..), throwIO, try)
+import           Data.IORef                     (newIORef, atomicModifyIORef')
+import           Data.Time.Clock                (getCurrentTime, addUTCTime)
+import           Network.HTTP.Client            (HttpException(..))
+import           Network.HTTP.Types.Status      (status500)
+import           Test.Tasty
+import           Test.Tasty.HUnit
+
+-- Amazonka imports
```
endgame, line 15: Don't need this; it'll be clear from the way the imports are sorted.

lib/amazonka/test/Main.hs
```diff
@@ -0,0 +1,97 @@
...
+tests = testGroup "Auth background error categorization"
+  [ testCase "UnknownAuthError wraps non-auth exceptions" $ do
+      menv <- mkAuthEnvAction (userError "random failure")
+      -- Note: fetchAuthInBackground immediately calls the provided action once
+      -- to seed the initial AuthEnv (this is the "first" call and succeeds),
+      -- then a background thread calls it again on refresh (the "subsequent"
+      -- call), which our helper makes fail with the supplied exception.
+      res <- try @AuthError $ do
+                _ <- fetchAuthInBackground menv
+                threadDelay 700000 -- give the background thread time to run (0.7s)
```
endgame, line 35: Pull this out and make it a constant in `Test.Amazonka.Auth.Background`, once you create that. I think we can probably make it even shorter.

lib/amazonka/test/Main.hs
```diff
@@ -0,0 +1,97 @@
...
+      case res of
+        Left (UnknownAuthError (SomeException _)) -> pure ()
+        Left e -> assertFailure ("unexpected error: " ++ show e)
+        Right _ -> assertFailure "expected UnknownAuthError"
+  , testCase "AuthServiceError is propagated" $ do
+      let svc = ServiceError'
```
endgame, line 41: Just inline it into the `AuthServiceError` constructor, since you'll only need it once:

```haskell
menv <- mkFailOnRefreshAuthAction $ AuthServiceError ServiceError' {
  abbrev = "sts",
  status = status500,
  -- etc
  }

File: lib/amazonka/test/Main.hs [hunk 15]
```diff
@@ -0,0 +1,97 @@
...
+                  , message = Nothing
+                  , requestId = Nothing
+                  }
+      menv <- mkAuthEnvAction (AuthServiceError svc)
+      res <- try @AuthError $ do
+        _ <- fetchAuthInBackground menv
+        threadDelay 700000
+      case res of
+        Left (AuthServiceError s) -> case s of
+          ServiceError' { code = c } -> c @?= "InternalFailure"
```
endgame, line 55: `AuthServiceError` has only one constructor so just merge the patterns:

```haskell
Left (AuthServiceError ServiceError { code }) ->
  c @?= "InternalFailure"
```

endgame (commented):
Thank you for doing this. I have a lot of comments but many of them are small, so it should be relatively quick to crank through them.

Please also format all source files with `ormolu`.

kushagarr:
@endgame  i think i have addressed all of the changes you requested, if there is still something missing or incorrect, let me know

PR #1042, opened 2025-06-13
Hooks tidyup

### LINKED ISSUE #995: Hook system improvements
While writing an example that uses hooks, I noticed that they were very hard to use correctly, and reasonable-looking code would silently ignore the request type you think you're hooking into.

We should provide `{add,remove}Hook{,_}` variants which are specialised to each hook type and are harder to misuse:

1. `addHookFor` and `addHookFor_` should be moved to the bottom of the file and marked `{-# DEPRECATED #-}` with a message recommending the hook-specific functions, and promising removal in 2.2. The export list for `Amazonka.Env.Hooks` should have a `-- * section` for deprecated functions.
2. `addAWSRequestHook` should be renamed to `addRequestHookFor`, and `addAWSRequestHook` should remain as a `{-# DEPRECATED #-}` alias, also scheduled for removal in 2.2.
3. `addAWSRequestHook_` should be deprecated and scheduled for removal in 2.2.
4. `removeHooksFor` and `removeHooksFor_` should be moved to the bottom of the file and deprecated, much the same as №1.
5. For each field in the `Hooks` record (example `requestRetry`) that has a `forall a`, we should have:
   - An add-for function for this hook: `addRequestRetryHookFor :: forall a b. Hook_ (Request a, Text, Retry.RetryStatus) -> Hook_ (Request b, Text, Retry.RetryStatus) -> Hook_ (Request b, Text, Retry.RetryStatus) = addHookFor_ @(Request a, Text, Retry.RetryStatus)`
   - A remove-for function for this hook: `removeRequestRetryHookFor :: _ = removeHookFor_`
6. Haddocks for each new function from №5
7. Export these in a new `-- ** Functions for specific hooks` first, with an explanatory note:
    ```haskell
    -- ** Functions for specific hooks
    -- $functions-specific-hooks (say that the type application for these is the AWS Request type e.g. `@S3.GetObject`.)
    addRequestHookFor,
    removeRequestsHooksFor,
    ...
    -- ** Functions for arbitrary hooks
    -- $functions-arbitrary-hooks
    noHook,
    noHook_,
    addHook,
    addHook_,
    ...
    ```

endgame:
The polymorphic hooks in 2.0 are easy to misuse. In amazonka-2.0, to
apply a response hook requires getting the argument type exactly
right:

```
addHookFor_
  @(Request ListBuckets, ClientResponse ListBucketsResponse)
  (...)
```

If you got this type even slightly wrong, your code would compile and
your hook would silently never fire.

Therefore: deprecate the polymorphic hook-editing
functions (`addHook`, `addHookFor`, etc.) and provide functions to
modify the hook at a particular field.

Where the hook is polymorphic in the request type, provide two "add"
functions: one that runs the hook against values of every type, and an
"add for" function that runs the hook against a particular AWS request
type.

Also provide a "remove for" function to block hooks from firing for a
particular AWS request type. `noHook` and `noHook_` can suppress a
hook for every request type, so we don't need to provide versions of
that for each hook field.

Closes #995

### DIFF
--- a/examples/src/Hooks.hs
+++ b/examples/src/Hooks.hs
@@ -12,8 +12,14 @@
 -- 3. Log the owner name returned by any ListBuckets calls.
 module Hooks where
 
-import Amazonka (ClientResponse, Request, discover, newEnv, runResourceT, send)
-import Amazonka.Env.Hooks (addHook, addHookFor, addHookFor_, requestHook, responseHook)
+import Amazonka (discover, newEnv, runResourceT, send)
+import Amazonka.Env.Hooks
+  ( addRequestHook,
+    addRequestHookFor,
+    addResponseHookFor,
+    requestHook,
+    responseHook,
+  )
 import Amazonka.S3
 import Control.Lens
 import Control.Monad.IO.Class (liftIO)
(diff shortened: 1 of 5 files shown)


Tristano8 (approved):
LGTM

PR #1013, opened 2025-01-10
Support setting endpoints from environment variables

ynishinaka:
Partially solves https://github.com/brendanhay/amazonka/issues/980.

The added `customEndpoints` loads environment variables `AWS_ENDPOINT_URL` and `AWS_ENDPOINT_URL_<SERVICE>` if any to provide a function to override default endpoints.

Intended to be used as follows:

```haskell
main = do
  env <- AWS.overrideService <$>
    AWS.customEndpoints <*>
    AWS.newEnv AWS.discover
```

PR #1046, opened 2025-07-31
Respect service-specific endpoint env variables

chreekat:
Based on #1013. This version is enabled by default, and has fixed one bug in the original patch.

TODO:

* [ ] 1. What to do if the environment variable value is not a valid URL? The original patch silently skips it, and I'm still doing that here. But that is surprising behavior. Better would be... a warning? An error? I could use guidance on that front.
* [ ] 2. Add a mention of the default overrides somewhere??
* [ ] 3. Rename and expose `customEndpoints` from the module
* [ ] 4. Changelog should mention this PR number

Opening as a draft so I can get feedback on #​1.


chreekat:
Alas, this patch doesn't work the way it should.

I'm just using awscli for comparison. With awscli, if I just set AWS_ENDPOINT_URL, it isn't used as an absolute override for services. The service-specific endpoint prefix is still applied. This patch doesn't do that yet.

chreekat:
core's Amazonka.Endpoint.defaultEndpoint has a lot of special-casing. Somebody with more knowledge would need to figure out how AWS_ENDPOINT_URL interacts with all of that.

Issue #964, opened 2023-10-30
Support for AWS_STS_REGIONAL_ENDPOINTS


BharatKJain:
When the env AWS_STS_REGIONAL_ENDPOINTS is set to regional, it's calling [sts.amazonaws.com](https://sts.amazonaws.com/) whereas it should call [sts.ap-south-1.amazonaws.com](https://sts.ap-south-1.amazonaws.com/) or some other regional endpoint.

[AWS docs reference](https://docs.aws.amazon.com/sdkref/latest/guide/feature-sts-regionalized-endpoints.html)

This helps when we're haskell code running in a closed environment where there's no internet available to make the STS calls, and regional endpoints can be used using VPC endpoints service which allows direct connectivity to some services.

BharatKJain:
More information on this issue, python is handling each region's endpoint in a botocore file present in this relative path: `botocore/data/endpoints.json`

Should we handle it like this? Or some other way?

PR #1048, opened 2025-10-17
fix default to region sts url incase SHOULD_MAKE_URL_FROM_REGION_STS env is not present.

govindrathi-juspay:
AWS has deprecated the use of global endpoint sts.amazonaws.com and such changing the default to use regional endpoint.


endgame:
Is this a planned fix for #964?

Issue #1037, opened 2025-05-12
Support bedrock and bedrock-runtime


alexbiehl:
It would be nice for Amazonka to offer bedrock and bedrock-runtime bindings.

endgame:
Blocked by #888

PR #1057, opened 2026-05-03
Fix CI config issues (actually build with different GHC versions)

jhrcek:
Hello,
I'm potentially interested in helping contributing some work towards unblocking https://github.com/brendanhay/amazonka/issues/1037

But first things first, looking at latest CI run on main https://github.com/brendanhay/amazonka/actions/runs/25209453900
it seems that CI is not in a very good shape.

This PR aims to fix some of that and hopefully unblock bumping the flake.lock dependencies so we could test with more recent versions of GHC.

Together, these changes fix 3 of the 4 CI jobs currently broken on master (see this same PR in my fork: https://github.com/jhrcek/amazonka/pull/2).
What's left is a failure in macOS build, which seems to fail when building the GHC 9.0 bootstrap compiler - not sure how to fix that yet..

Heads up: I'm not very skilled with nix (just know some basics)..

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,6 +33,7 @@ jobs:
             trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ=
             extra-substituters = https://cache.iog.io
             experimental-features = nix-command flakes
+            accept-flake-config = true
 
       - uses: cachix/cachix-action@v17
         with:
@@ -72,6 +73,7 @@ jobs:
             trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ=
             extra-substituters = https://cache.iog.io
             experimental-features = nix-command flakes
+            accept-flake-config = true
 
       - uses: cachix/cachix-action@v17
         with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,13 +6,13 @@ concurrency:
 
 on:
   pull_request:
-    branches: main
+    branches: [ main ]
     types:
       - opened
       - synchronize
 
   push:
-    branches: main
+    branches: [ main ]
     paths-ignore:
       - "configs/**"
       - "scripts/**"
@@ -23,13 +23,10 @@ on:
 
 jobs:
   amazonka-gen:
-    strategy:
-      fail-fast: false
-
     runs-on: ubuntu-latest
 
     steps:
-      - uses: cachix/install-nix-action@v22
+      - uses: cachix/install-nix-action@v31
         with:
           nix_path: nixpkgs=channel:nixos-unstable
           extra_nix_config: |
@@ -37,12 +34,12 @@ jobs:
             extra-substituters = https://cache.iog.io
             experimental-features = nix-command flakes
 
-      - uses: cachix/cachix-action@v12
+      - uses: cachix/cachix-action@v17
         with:
           name: amazonka
           authToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
 
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v6
 
       - run: nix build
 
@@ -68,7 +65,7 @@ jobs:
         run: |-
           sudo rm -rf /usr/local/lib/android /usr/share/dotnet /opt &
 
-      - uses: cachix/install-nix-action@v22
+      - uses: cachix/install-nix-action@v31
         with:
           nix_path: nixpkgs=channel:nixos-unstable
           extra_nix_config: |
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -6,13 +6,13 @@ concurrency:
 
 on:
   pull_request:
-    branches: main
+    branches: [ main ]
```
jhrcek, line 9: The YAML extension in vscode caught couple of schema issues in this config (screenshot below)
- branches should be an array of values
- strategy: section has mandatory matrix parameter - but it seems to me since amazonka-gen job is a standalone job, it doesn't make sense to have the strategy key at all.


Image

.github/workflows/build.yml
```diff
@@ -23,26 +23,24 @@ on:
...
     steps:
-      - uses: cachix/install-nix-action@v22
+      - uses: cachix/install-nix-action@v31
         with:
           nix_path: nixpkgs=channel:nixos-unstable
           extra_nix_config: |
             trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ=
             extra-substituters = https://cache.iog.io
             experimental-features = nix-command flakes
+            accept-flake-config = true
```
jhrcek: Not sure if this is needed at all, but since there's this: https://github.com/brendanhay/amazonka/blob/main/flake.nix#L152

and I saw this in the CI output:
```
warning: ignoring untrusted flake configuration setting 'allow-import-from-derivation'.
Pass '--accept-flake-config' to trust it
```

I'm adding it here. Let me know if I should remove it.

.github/workflows/build.yml
```diff
@@ -68,27 +66,28 @@ jobs:
...
           authToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
 
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v6
 
       - name: cabal update && cabal freeze
         run: |
-          nix develop --print-build-logs --command \
-              bash -c 'cabal update && cabal freeze' '.#${{matrix.ghc}}'
+          nix develop '.#${{matrix.ghc}}' --print-build-logs --command \
```
jhrcek, line 85: This matrix argument was not taken into account at all - all the jobs were using the ghc 9.4.8 provided by the default shell

Image

It seems to me this is because the last param was somehow consumed by the `bash -c ...` part and not passed to the nix develop command.

Moving the arg before the --command seems to have fixed it (in that I see `Build profile: -w ghc-9.6.6 ...` in the `cabal build core` step).

.github/workflows/build.yml
endgame: That flag looks benign but is actually rather dangerous. See https://github.com/NixOS/nix/issues/9649 where it can be used to gain `root` access rather easily.

Can you please instead enable `allow-import-from-derivation` in the `extra-nix-config`?

flake.nix
endgame, line 1: Can you please explain what is forcing you to add the extra packages? I haven't needed them even when running NixOS (which is where I expect the most shenanigans around paths). If there's something wrong with the GHC you get when you use Nix as a package manager on a standard distribution, I think setting up the right `ghcWithPackages` calls to wire in the native libraries is probably going to work better, but I want to know what the thinking is before I insist on that change.

endgame (commented):
Thank you for taking this on. I'm not the best at GHA so I'd been avoiding it.

I'm in the middle of a house move and I'm not sure if this PR is completely ready to go but in the interest of not losing momentum I wanted to give it at least a once-over. I'll try to be responsive but if I go silent, please ping.

As for your goal of fixing #1037. That's going to be a substantial amount of work, because we are stuck on an old pin of `botocore`. We can't advance `botocore` because if we do, we break the generation of existing libraries. The generation of existing libraries is brittle because we use a `Ptr` type to break cyclic references in the generator (because the generator usually resolves references by inlining), and a `Ptr` doesn't contain enough information for the generator to correctly calculate the set of instances a type needs to have (see #888 for the gory details). I've written up my vision for carving off an "output" layer to make the gnarlier bits of the generator to work on. If you're interested, those notes are at #1058.

endgame:
Please also add an entry in `lib/amazonka/CHANGELOG.md`, section `Fixes`, as this is the primary changelog for the project and it's good to get credit for a thankless task.

flake.nix
jhrcek: I'm still iterating on this (and learning ins and outs of nix along the way), but TBH I initially just let claude-code thrash around to make the build pass. It seems to have found a fix which however has bunch of unnecessary stuff. Based on your remark I pushed it harder into making it demonstrate to me that all the changes are actually necessary and managed to trim it down to just this PKG_CONFIG_PATH declaration below. I still want to look around to see how people using nix + zlib do it in their projects, but this seems to be necessary to avoid linker picking the `zlib1g-dev` package that comes pre-installed on ubuntu github action runners.

jhrcek:
Thanks for the initial feedback and staking out the path that would need to be followed towards making bedrock lib(s) available. I'm learning nix as I go but I'd like to also try to fix the macos job as part of this PR. Consider this still WIP.

.github/workflows/build.yml
jhrcek: I removed it as it's not actually needed for anything in CI.

jhrcek:
@endgame I consider this ready for review.
You can see the 3 linux configurations passing in my [fork](https://github.com/jhrcek/amazonka/pull/2).

I tried to fix the macos-latest build too, but that will probably require bumping to more recent version of ghc in the macos job. Still I think this PR is worth merging as it's better to have at least some end-to-end build in CI than none.

I fell down the rabbit hole of trying to bump the dependency versions in flake.nix, but got tons of failures, because package sets we're getting generally require newer version of hashable, which I saw is [problematic](https://github.com/brendanhay/amazonka/blob/8690f1a4efe70f507d41c1d018d207334f176a2c/flake.nix#L88-L96).

It looks like the gen improvements will have to be implemented first before we're able to untie this Gordian knot and bump to newer nixpkgs / ghc versions in CI.

I'll start looking into how the gen stuff works to see if I can contribute in meaningful way.

PR #1060, opened 2026-05-19
fix: throw AuthError in fromSSO upon getRoleCredentials fail

### LINKED ISSUE #1059: Amazonka.Auth.SSO needs to throw AuthError
In https://github.com/brendanhay/amazonka/blob/8690f1a4efe70f507d41c1d018d207334f176a2c/lib/amazonka/src/Amazonka/Auth/SSO.hs#L95 the call to getRoleCredentials will throw an exception if it fails. it should raise an `AuthError` so that `runCredentialChain` correctly exhausts.

This should be a simple change: use `sendUnsignedEither` instead of `sendUnsigned` and throw the correct exception on `Left`.

mikeplus64:
Fixes #1059 

Happy to adjust to preferred style

### DIFF
--- a/lib/amazonka/CHANGELOG.md
+++ b/lib/amazonka/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ### Changed
 
+- `amazonka`: `Amazonka.Auth.SSO.fromSSO` throws an `AuthError` now.  Previously
+  it could throw a `Amazonka.Types.Error`.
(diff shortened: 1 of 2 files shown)


endgame (commented):
Please add an entry to `lib/amazonka/CHANGELOG.md`, which is the high-level changelog for the whole project.

mikeplus64:
Sorry - done

endgame (approved):

PR #1061, opened 2026-05-28
Allow aeson 2.3

fgaz-scrive:
https://hackage.haskell.org/package/aeson-2.3.0.0/changelog https://haskell.github.io/security-advisories/advisory/HSEC-2026-0007.html

Built with

    cabal build all \
      --enable-tests --enable-benchmarks \
      --constraint=aeson==2.3.0.0 \
      --allow-newer=pandoc:aeson \
      --allow-newer=pandoc-types:aeson \
      --allow-newer=aeson-pretty:aeson

And tested with `cabal test` with the same flags

### DIFF
--- a/lib/amazonka-core/amazonka-core.cabal
+++ b/lib/amazonka-core/amazonka-core.cabal
@@ -102,7 +102,7 @@ library
     Amazonka.Waiter
 
   build-depends:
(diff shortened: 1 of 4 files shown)


endgame (approved):

endgame:
Thanks

PR #1064, opened 2026-06-10
Update crypton deps to support v1.1.x and migrate from memory to ram

### LINKED ISSUE #1062: Replace `memory` dependency with `ram`

nikita-volkov:
- Relax crypton upper bound from ^>=1.0.0 (i.e. <1.1) to <1.2 to support crypton 1.1.x
- Replace 'memory' package dependency with 'ram' (>=0.22), an API-compatible fork required by latest crypton
- Update both amazonka-core and amazonka-s3-encryption cabal files

### DIFF
--- a/lib/amazonka-core/amazonka-core.cabal
+++ b/lib/amazonka-core/amazonka-core.cabal
@@ -109,16 +109,16 @@ library
     , conduit               >=1.3
     , conduit-extra         >=1.3
     , containers            >=0.5       && <0.8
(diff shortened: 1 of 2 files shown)


endgame:
Thank you.

ysangkok:
It doesn't make sense to claim compatibility with older crypton, as they use `memory` exclusively. If we want to support old crypton, the dependency on `ram` should be controlled by a flag. But I think it would be better to just require crypton-1.1 (which requires `ram`).

